Issue #1, opened 2017-08-15
Job succeeds, but results tab not presents


chrisgorgo:
![image](https://user-images.githubusercontent.com/238759/29306737-817f3ba0-81d1-11e7-8994-0f435ded5b77.png)

https://openneuro.dev.sqm.io/datasets/ds000118/versions/00001?app=TRACULA&version=7&job=4de25fb8-0d60-4323-bfd5-59e66b3afb05

Previously I noticed that for some (other) jobs the results were loaded asynchronously and the tab would show up after some delay. This time it does not show up at all.

chrisgorgo:
For some context - this job took much longer than expected to complete and the number of subjobs seems to low considering the number of input participants. Furthermore the "Logs" tab is also missing.

chrisgorgo:
Another case https://openneuro.dev.sqm.io/datasets/ds000118/versions/00001?app=TRACULA&version=10&job=6876a8ee-66d9-45fc-b35c-a705a618867f

chrisgorgo:
Could be related to https://gitlab.sqm.io/stanford_crn/openneuro/issues/215

PR #15, opened 2017-08-28
Status check updates

### LINKED ISSUE #4: Analysis Status Issue for long running analysis with multiple jobs
The current behavior of the application is to update the overall status of the analysis once all the jobs in a multi-job analysis have been completed.   This can cause problems if finished jobs sit in the SUCCEEDED or FAILED state on batch for longer than 24 hours waiting for other jobs to finish because the jobs are only persisted for 24 hours.  Those jobs are no longer returned during status polling causing a potentially inaccurate status to be reported back to the client.  
For example, this job below shows displays a status of SUCCEEDED even though the job status in mongo is RUNNING.  
image

We should track the completion of individual jobs "real-time" so as to not lose track of the status of each individual job due to this issue.

jstiehl:
* resolves https://github.com/OpenNeuroOrg/openneuro/issues/4 and should resolve some other issues with long running jobs.
* adding a batchStatus property to job.analysis in mongo.
* checking for consistency between the number of jobs returned during status check from batch and the total number expected. If there is a mismatch, the the dropped job statuses from the previous check are used for the missing ones.

### DIFF
--- a/server/handlers/awsJobs.js
+++ b/server/handlers/awsJobs.js
@@ -430,6 +430,10 @@ let handlers = {
                 batchStatus = job.analysis.batchStatus.map((statusObj) => {
                     return batchStatus[statusObj.job] ? batchStatus[statusObj.job] : statusObj;
                 });
+                // update status array to include ALL job statuses
+                statusArray = batchStatus.map((statusObj) => {
+                    return statusObj.status;
+                });
             }
 
             analysis.batchStatus = batchStatus;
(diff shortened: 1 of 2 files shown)


jstiehl:
I think this change will also address https://github.com/OpenNeuroOrg/openneuro/issues/1

nellh:
These changes look good and are working for me. Merging and we can do some further testing on dev.

PR #16, opened 2017-08-29
Adding checkbox input in job run modal

### LINKED ISSUE #11: Boolean app parameters do not translate to checkboxes
image

results in

image

instead of a checkbox.

If the checkbox is checked this should result in inclusion of `--ica` flag if it is not checked it should not be included.

jstiehl:
* resolves https://github.com/OpenNeuroOrg/openneuro/issues/11
* adding a checkbox input for Boolean parameters in job run modal.


nellh:
No issues with functionality but checkbox styles are inconsistent between platforms. Here's Chrome on Linux:

screenshot from 2017-08-29 16-08-41

I think we can address that as part of #18 or a new issue if we want to separate the job run improvements.

PR #17, opened 2017-08-29
Handle all versions of the log stream names after Batch API changes

nellh:
This solves issue #12. There's three situations to handle here.

1. Jobs run prior to 2017-08-21T14:00-07:00. These use the old log stream names in CloudWatch and are correct. These need to be served without changes.
1. Jobs run prior to pulling in this fix but after the August 21st date. These use the old log stream name internally but CloudWatch uses the new format.
1. Jobs run after this branch is merged. These store the new CloudWatch stream name format.

All of these should have accessible logs (for individual tasks or download all) with these changes.

### DIFF
--- a/server/handlers/awsJobs.js
+++ b/server/handlers/awsJobs.js
@@ -330,7 +330,7 @@ let handlers = {
         let appName = req.params.app;
         let jobId = req.params.jobId;
         let taskArn = req.params.taskArn;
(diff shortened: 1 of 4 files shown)


jstiehl (approved):
This works great. My broken logs were fixed on refresh on my local machine. Ran a new job and the logs were available. Ran tests locally and all looked good.

PR #21, opened 2017-08-30
moving bucket definitions to buildBatchParams

### LINKED ISSUE #20: S3 configuration is only updated by new app versions
This forces updates to app versions when any changes are made to the S3 buckets, rather than applying that configuration to a submitted job. The configuration needs to be added to job submission and removed from app definition registration.

jstiehl:
* resolves #20 
* moving out bucket env settings out of job definition and into buildBatchParams so the buckets are defined during the job run instead during the app definition.

### DIFF
--- a/server/libs/aws/batch.js
+++ b/server/libs/aws/batch.js
@@ -73,11 +73,6 @@ export default (aws) => {
                 return callback(error);
             }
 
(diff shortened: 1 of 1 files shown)


nellh (approved):
Tested manually fixed and existing app definitions with this. Looks good.

PR #22, opened 2017-08-30
Add check submit status method to componentWillReceiveProps

### LINKED ISSUE #14: intermittent button failure on analysis dropdown
I have had an intermittent problem where the start button on an analysis fails to do anything when clicked.  It seems to occur only when I run an analysis after just having run another within the same session on that window.  refreshing the window seems to resolve the problem. low priority but I can imagine that it would flummox some people.

jstiehl:
* resolves #14 
* moving submission button status check to different lifecycle method as modal is mounted on page load, whether shown or not, and does not get unmounted unless we navigate away from dataset page.

### DIFF
--- a/app/src/scripts/dataset/tools/jobs/index.js
+++ b/app/src/scripts/dataset/tools/jobs/index.js
@@ -74,6 +74,8 @@ export default class JobMenu extends React.Component {
                 }
             });
         }
+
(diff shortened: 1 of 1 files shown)


nellh (approved):
Looks good. I wondered if this was redundant with the call in componentDidMount but looks like it is needed in both places.

PR #28, opened 2017-08-31
Adding canceled to ignored statuses for auth

jstiehl:
CANCELED status was counting against job run limit for non-admin users.  This should not happen.  Adding that status to ignored status list for jobs

### DIFF
--- a/server/libs/auth.js
+++ b/server/libs/auth.js
@@ -115,7 +115,7 @@ let auth = {
             return next();
         }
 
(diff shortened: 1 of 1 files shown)


nellh (approved):

chrisgorgo:
Can you deploy this in prod as a quick fix?

jstiehl:
Hi Chris,

We will get the change out today after we get it tested on Dev.

James

Issue #25, opened 2017-08-31
Upgrade to React 15


nellh:
Some front end bugs (especially some UI performance issues) depend on upstream fixes in React or one of the other libraries requiring newer React to upgrade. The major things that need to be done here are:

- [x] Resolve dependency issues (switch to yarn and fix the dependency list to accurately reflect installed deps).
- [x] Write some tests for the UI components to help validate changes for the update.
- [x] Refactor routing to upgrade to react-router v4.
- [x] Handle all the now invalid uses of the React API and missing packages that were refactored out of React.
- [x] Extra padding below navbar in some views. 
- [x] Switching snapshots does not correctly trigger the loading state and animations.
- [x] Active states for many tabs are not working.

nellh:
This is nearly ready but a few regressions to fix that I've added to the description.

PR #30, opened 2017-08-31
Switch from npm to yarn

nellh:
This switches to yarn over npm. The main benefits are faster yarn installs (speeds up clean builds by a factor of 4) and more correct dependency resolution.

This branch also updates Node to the latest stable LTS version, 6.11.2. This is preparation for #25 in later pull requests.


jstiehl (approved):
Was able to get my dev system running with these changes.

PR #33, opened 2017-09-01
Switch to Jest for app tests

nellh:
This drops Mocha and switches to Jest + Enzyme for client side unit tests. The goal here is to support snapshot tests and enable some handy wrappers for testing React components. This includes a few example tests just to verify that all the pieces are working together and converts the Footer component to a functional one (to test swapping out a component for another implementation with the tests.


nellh:
Making one more commit before this is reviewable.

jstiehl (approved):

PR #34, opened 2017-09-01
"Invalid" special case Validation component fix.

nellh:
This replaces the list or string behavior in the Validation component and adds some tests to validate that going forward. Instead, an 'invalid' boolean prop returns the invalid message early.

This depends on #33.


nellh:
Working on an updated fix for this.

PR #46, opened 2017-09-07
Webpack support

nellh:
This drops Gulp and Uglify in favor of Webpack and babel-minify.

* Automated dead code/asset pruning.
* Better source maps.
* Simpler build scripts.
* Working automatic reload/partial builds for development within docker containers.

There's some tradeoffs, the dev build is faster but the production build is slower. The resulting bundle is bigger for dev but smaller for prod.

There's a new branch of [crn_deploy here](https://github.com/poldracklab/crn_deploy/tree/webpack-dev-support) needed to use this locally.

A known issue is the docker container from the gulp build will conflict with this build, so you need to build a new local container to test. This shouldn't be a problem when the container is rebuilt on master.

Another known issue is Papaya is not copied into the assets, so viewing nii files is broken. I think this is reviewable without this but I should have this figured out in the next day or so before merging this.


jstiehl (commented):
Was able to get app up and running with this branch. Will wait until papaya issue is fixed before final approval.

nellh:
The Papaya issue is worked around now. Really it could be shimmed with webpack and that could shrink the include size some but this restores the dependency as it was pre-webpack.

jstiehl:
Papaya error is gone now from browser console and I am able to view nii files still.

jstiehl (approved):

PR #48, opened 2017-09-08
UI/UX

### LINKED ISSUE #18: App Definition Parameters UI Improvements
The interface for adding parameters to an app definition is currently a little confusing and needs improvement.  The current UI is displayed below. 
image
A common issue with this UI is the adding of a parameter but forgetting to press ADD button before submitting the app definition.  This interface should be redesigned to make the user experience better and less confusing.

As part of this work, we should also address hidden options, as this will add another checkbox to the interface.  See below for details from Chris.

"Some BIDS App command line arguments (for example those regarding cpu and memory restrictions) should be always the same for all jobs and users submitting jobs should not be able to change them. The easiest way to implement this is to add a checkbox next to each parameter to determine if this parameter will be visible to users submitting jobs."

Xesme:
fixes #18

### DIFF
--- a/app/src/sass/crn_app/_admin.admin.scss
+++ b/app/src/sass/crn_app/_admin.admin.scss
@@ -194,7 +194,7 @@
     line-height: 30px;
 }
 
-.admin-job-parameters .required-button {
+.admin-job-parameters .admin-button {
     border: none;
     background-color: transparent;
     padding-top: 0;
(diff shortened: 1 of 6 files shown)


nellh (changes requested):
These changes look good and the hidden functionality works. One issue with a related commit (efe5fb1165bcf5de67d7c240dca55c590e066765) causes some dropdown issues.

nellh:
Here's a screenshot of the dropdown issue.

commit efe5fb11 issue

nellh (approved):

Issue #19, opened 2017-08-29
Analyses Sections Redesign


jstiehl:
Currently the analyses sections in the dashboards are just list of analyses that link to the dataset that the analyses were run again.  The dataset can be filtered by app and app version and sorted by date.  See below.
![image](https://user-images.githubusercontent.com/5668826/29847786-44a72bbe-8cd2-11e7-9709-c206ca3c1985.png)

There is probably a better way to display this data.  Perhaps by grouping the analyses by dataset? Filtering by user who ran the analyses (for Public Dashboard)? Other ideas?

The expected outcome of this issue is to provide preliminary recommendations for UI updates that might improve the user experience with the Analyses sections of the application.

chrisgorgo:
I am pretty happy with how this view looks like right now. The one addition I would think of is color coding status.

I see three use cases:
1) User wants to see their most recent jobs (to check their status) - current default view serves this issue well.
2) User wants to see all jobs run for a given dataset - this is relatively easy by going to the dataset page, one improvement I would suggest is to flatten the hierarchy of folded boxes so getting to results would require fewer clicks.
3) User wants to see all public jobs of a particular app - this is possible with filtering of the public dashboard, but in the future I see this being also a part of dedicated "app library" section

PR #59, opened 2017-09-12
Analyses Sections Redesign

Xesme:
fixes issue #19 

Adds color coding for analyses, fixes bug with sorting by name, adds sorting by user and dataset.

### DIFF
--- a/app/src/scripts/dashboard/dashboard.jobs.jsx
+++ b/app/src/scripts/dashboard/dashboard.jobs.jsx
@@ -84,7 +84,7 @@ let Jobs = React.createClass({
             return (
                 <div className="fade-in  panel panel-default" key={job._id}>
                     <div className="panel-heading">
-                        <div className="header clearfix">
+                        <div className="header clearfix" id={job.analysis.status}>
                             <Link to={'snapshot'} params={{datasetId: bids.decodeId(job.datasetId), snapshotId: bids.decodeId(job.snapshotId)}} query={{app: job.appLabel, version: job.appVersion, job: job.jobId}}>
                                 <h4 className="dataset-name">{job.appLabel} - v{job.appVersion}</h4>
                             </Link>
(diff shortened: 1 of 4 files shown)


app/src/scripts/dashboard/dashboard.jobs.jsx
```diff
@@ -84,7 +84,7 @@ let Jobs = React.createClass({
             return (
                 <div className="fade-in  panel panel-default" key={job._id}>
                     <div className="panel-heading">
-                        <div className="header clearfix">
+                        <div className="header clearfix" id={job.analysis.status}>
```
nellh: This is going to lead to conflicting IDs with multiple jobs with the same status on the page. I'd suggest changing it to use a CSS class and maybe making the selector more specific.

app/src/scripts/dashboard/dashboard.jobs.store.js
```diff
@@ -56,8 +56,9 @@ let DashboardJobStore = Reflux.createStore({
                 isTimestamp: true
             },
             sortOptions: [
-                {label: 'Name', property: 'datasetLabel'},
-                {label: 'Date', property: 'analysis.created', isTimestamp: true}
+                {label: 'Name', property: 'jobName'},
+                {label: 'Date', property: 'analysis.created', isTimestamp: true},
+                {label: 'Dataset', property: 'datasetLabel'},
```
nellh: The linter doesn't like this because of the extra comma at the end of the list. You can see the linting issues before you commit with 'yarn lint'.

> /srv/crn-app/src/scripts/dashboard/dashboard.jobs.store.js
  61:61  error  Unexpected trailing comma  comma-dangle
> ✖ 1 problem (1 error, 0 warnings)

nellh (changes requested):
The conflicting id attributes needs to be fixed to merge this.

app/src/scripts/dashboard/dashboard.jobs.jsx
Xesme: I have changed this to be a wrapping div with the class(Name) instead.  As for the selector, it is already embedded to be fairly specific in css, I can make it more specific if need be. But, as of right now it only color codes analyses.

app/src/scripts/dashboard/dashboard.jobs.store.js
Xesme: I have removed the rogue comma

nellh (approved):

PR #63, opened 2017-09-14
Fix boolean shell flags

### LINKED ISSUE #11: Boolean app parameters do not translate to checkboxes

nellh:
This adds tests for and handles the case where boolean flags should be converted to '--flag' instead of '--flag true'. Fixes #11.


jstiehl (approved):
I added tests for the false boolean case to test the PR so I thought I would just commit those as well. Works great.

nellh:
Awesome, thank you for the extra tests!

Issue #60, opened 2017-09-12
Linux / Chrome issue with submit jobs dropdowns


nellh:
Noticed this in #48 and it's Linux specific, so making it a new issue to fix.

![commit efe5fb11 issue](https://user-images.githubusercontent.com/11369795/30301783-9aa28e20-9711-11e7-8294-1d3286494094.png)

PR #64, opened 2017-09-14
Chrome + Linux boolean job parameter css fixes

nellh:
@Xesme I restructured your change to the job run parameters to only change the checkbox layout and support a clickable label. Let me know what you think of this.

> boolean option example
> Parameter key is "boolean" and description is "some kind of boolean" in this example.

This reverts the flexbox change for the other components since it was breaking rendering for Chrome on Linux with the React-Select based parameters.


nellh:
This fixes #60

Issue #45, opened 2017-09-07
Display "Support" field in the failed job results


chrisgorgo:
When job fails we want to instruct users how to get in touch with the app developers.

nellh:
Merged #70

Issue #44, opened 2017-09-07
Display "Acknowledgements" field in the successful job results


chrisgorgo:
We need to show users how to acknowledge the authors of the app. Best place to do it in the job result box.

This is relatively high priority if we want to make app developers happy.

nellh:
Merged #70

PR #70, opened 2017-09-17
Job results window updates

Xesme:
This updates the "jobs" window in reference to issue #45  to provide users information to contact support for app and #44 to acknowledge authors of the app.

### DIFF
--- a/app/jestsetup.js
+++ b/app/jestsetup.js
@@ -1,6 +1,10 @@
 // Make Enzyme functions available in all test files without importing
 import React from 'react';
 import { shallow, render, mount } from 'enzyme';
+import moment from 'moment-timezone';
+
+// Run all tests in virtual Katmandu (UTC +05:45)
+moment.tz.setDefault('Asia/Katmandu');
 
 global.React = React;
 global.shallow = shallow;
--- a/app/package.json
+++ b/app/package.json
@@ -58,6 +58,7 @@
     "file-loader": "^0.11.2",
     "html-webpack-plugin": "^2.30.1",
     "jest": "^20.0.4",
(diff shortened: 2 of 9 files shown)


nellh:
@Xesme This can be simplified quite a bit. You can get the app definition values during render and avoid needing to compare against the activeJob state like this:

```javascript
// Get the most recent app definition
let newestVersion = Math.max(...Object.keys(this.state.apps[app.label]));
let appDef = this.state.apps[app.label][newestVersion];
// Destructure the needed values.
let {acknowledgements, support} = appDef.descriptions;

return (
    <Panel className="jobs" header={app.label}  key={app.label} eventKey={app.label}>
        <div>{acknowledgements}</div>
        <div>{support}</div>
        <Accordion accordion className="jobs-wrap" activeKey={this.state.activeJob.version} onSelect={actions.selectJob.bind(null, 'version')}>
            {version}
        </Accordion>
    </Panel>
);
```

You should do something similar to grab the version specific description and pass it into the Runs component as a prop. That way these can be displayed as part of the results (for acknowledgements) or near the failed status (for support).

These fields are also meant to be displayed as Markdown so they need some preprocessing before they are added to the DOM:

```javascript
<div dangerouslySetInnerHTML={markdown.format(support)}/>
```

Because they are Markdown documents, they should probably be full width blocks to preserve as much formatting as possible.

nellh (changes requested):
Should be simplified and needs to add the descriptions conditionally.

nellh (approved):

nellh:
@Xesme I rebased this against master and made some tweaks to support longer markdown descriptions. Going to merge it once tests finish :+1:

PR #71, opened 2017-09-17
Event logs size/consistency improvements

### LINKED ISSUE #8: Event logs are slow to load
Load time: 90 sec.
Data transferred: 45 Mb (40 Mb from https://openneuro.org/crn/eventlogs?root=true)

This situation will only get worse. One solution would be to provide server side paging.

### LINKED ISSUE #40: Add links to jobs in the event log
For debugging and monitoring purposes it would be useful to be able to go from an event log entry to the dataset/job page.

nellh:
This reduces the size of the JOB_COMPLETED logs by filtering out stuff like the results and logs file references, and adds a link to the related object for jobs events in the admin dashboard.

The current logs collection has to be truncated to test/deploy this. This fixes #8, fixes #40, and part of #3.

### DIFF
--- a/app/src/scripts/admin/__tests__/__snapshots__/admin.logs.spec.jsx.snap
+++ b/app/src/scripts/admin/__tests__/__snapshots__/admin.logs.spec.jsx.snap
@@ -1,0 +1,86 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`admin/Logs renders successfully 1`] = `
+<div
+  className="dashboard-dataset-teasers fade-in admin-users clearfix"
+>
+  <div
+    className="header-wrap clearfix"
+  >
+    <div
+      className="col-sm-9"
+    >
+      <h2>
+        Events Log
+      </h2>
+    </div>
+    <div
+      className="col-sm-3"
+    >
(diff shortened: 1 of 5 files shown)


nellh:
This also fixes #40.

jstiehl:
There seems to be a test failing that is unrelated to this change when I run locally.  
```
  ● dataset/tools/jobs/parameters › renders successfully

    Warning: Failed propType: Calling PropTypes validators directly is not supported by the `prop-types` package. Use `PropTypes.checkPropTypes()` to call them. Read more at http://fb.me/use-check-prop-types Check the render method of `JobParameters`.
```

Are you seeing this as well, Nell?

nellh:
@jstiehl You need to run `yarn install`, that's caused by React 0.14.8 vs React 0.14.9

jstiehl:
Server side pagination is in place but client side is not making "paginated" requests.  Client side should be updated to make event log requests for paginated results.

nellh:
The client side patch to enable pagination was starting to look like a rewrite, so I've disabled it here and I'll make a new PR with that.

jstiehl (approved):
This looks good.

PR #72, opened 2017-09-17
Add admin datasets/jobs dashboards

nellh:
This adds two new tabs to the admin view that let admins explore all datasets and jobs. @chrisfilo I think this is a decent solution to allowing you to quickly see new datasets being added over time? Let me know what you think of adding these two tabs.

screenshot from 2017-09-17 16-33-34

This works the same as the user or public dashboards, but without the user / public filters.


chrisgorgo:
Yes! This is perfect!

jstiehl:
This change is resulting in jobs run on unpublished datasets showing up in the Public Analyses section of the app for Admin users.  I realize that Admins will be able to see these jobs in the All jobs section anyways but it seems like we probably still want the Public Analyses section to reflect only jobs run on published datasets.
image

nellh:
@jstiehl Thanks for catching that. Fixed in the latest commit.

jstiehl (approved):
This looks good.  Analyses on unpublished datasets are no longer showing up in Public Analyses.

Issue #13, opened 2017-08-27
problem uploading large files


poldrack:
I am trying to upload a dataset that includes a large fMRI data file (~487 MB) and the upload has failed twice on that file.  This is for https://openneuro.org/datasets/ds001040

nellh:
@sadevp Since you are seeing this too, can you record a HAR file with Chrome and send that to me to help debug the root cause of this issue?

Here's some [instructions for recording the HAR file](https://community.box.com/t5/Managing-Content-Troubleshooting/How-to-Generate-a-HAR-File-in-Chrome-IE-or-Firefox/ta-p/366). Since this will contain some private information, please share it without attaching it to the issue.

sadevp:
Attachments available until Oct 12, 2017
Hi Nell,

The file is too large to share over email.

Thanks,
Sadev

Click to Download<link removed>
chrome_console.log
35 KB
Click to Download<link removed>
openneuro.org.har
95.6 MB




On Sep 12, 2017, at 11:21 AM, Nell Hardcastle <notifications@github.com<mailto:notifications@github.com>> wrote:


@sadevp<https://github.com/sadevp> Since you are seeing this too, can you record a HAR file with Chrome and send that to me to help debug the root cause of this issue?

Here's some instructions for recording the HAR file<https://community.box.com/t5/Managing-Content-Troubleshooting/How-to-Generate-a-HAR-File-in-Chrome-IE-or-Firefox/ta-p/366>. Since this will contain some private information, please share it without attaching it to the issue.

nellh:
@sadevp Thanks, I've downloaded those files and removed the links from the issue.

suyashdb:
i tried uploading ds208 three times. Following is the description what happened each time. 

In each of the case below my computer didn't sleep and i was on wired internet connection.

Try 1: Failed  
Upload it via `normal chrome` browser with lot of plugins.  All Errors were mostly for plugins expect this one: `Failed to load resource: net::ERR_NETWORK_CHANGED
/api/projects/202020206473303031303638/files Failed to load resource:`

Try 2: Failed 
Uploaded via `incognito` window(without plugins). Got same network error
`Failed to load resource: net::ERR_NETWORK_CHANGED
/api/projects/202020206473303031303638/files Failed to load resource:`

Try3: Success 
Incognito window. Success No errors

nellh:
[...]

PR #73, opened 2017-09-18
Support retries for uploads

nellh:
This will retry network exceptions during uploads up to 3 times before canceling the upload. The main change is upgrading superagent from 1.0.0 to 3.6.0 (to get retry support).


nellh:
This should fix #13 in many cases but I'd like to put this on dev and have the affected users test it some if we're not seeing any regressions around uploads or other client side requests.

jstiehl (approved):
I am still able to upload a dataset.

PR #75, opened 2017-09-18
Automatic prettier formatting on commit

nellh:
I want to discuss this a bit and add it to the server package.json as well before merging but this enables a git hook to always run prettier before committing.

I changed some of the code styling rules and the idea is to use prettier as an automatic style guide. Here's the settings used:

```yaml
singleQuote: true
jsxBracketSameLine: true
semi: false
trailingComma: all
```

Changes from what we had before and justifications.

1. 2 space indents. This matches a lot of other Javascript style guides and is the prettier default. Some nested HTML is a lot more compact.
1. Trailing commas. [Makes rearranging lists easier and diffs better.](http://2ality.com/2015/11/trailing-comma-parameters.html)
1. JSX closing brackets on the same line. We had this mixed all over the place, so just picking one to be consistent.
1. No semicolons. I think it saves some overhead of thinking about expression vs declaration contexts and we commit a decent amount of linter errors that are fixed by just adding one, so removing them seems like a net win.


nellh:
@jstiehl I'd like to merge this but not include it in the 1.0.2 update that's been tagged.

jstiehl (approved):
App starts up and prettier cleans up code as expected on commit.

PR #107, opened 2017-09-27
Fix >20 dependencies rejection with more than two analysis levels

### LINKED ISSUE #100: More than 20 participants can schedule more than 20 dependencies
Hello, 

I am attempting to run freesurfer on several datasets, and I keep receiving notification that my analysis has been 'rejected'. Please see screenshot for an example. 

I am using v21, and typing the random string '12345' into the license key field. 

Please advise. I would like be able to run freesurfer on all uploaded datasets. 

Thanks!

screen shot 2017-09-23 at 3 17 22 pm

nellh:
The "passes dependencies up exactly one level" test covers this case.

This fixes #100

### DIFF
--- a/server/tests/libs/aws.spec.js
+++ b/server/tests/libs/aws.spec.js
@@ -139,6 +139,72 @@ describe('libs/aws/batch.js', () => {
       assert.equal(Math.max(...lengths) - Math.min(...lengths), 1)
     })
   })
+  describe('_submitLevels()', () => {
+    const batchJob = {
+      containerOverrides: {
+        environment: [],
+      },
+    }
+    const analysisLevels = [
+      { value: 'participant' },
+      { value: 'group1' },
+      { value: 'group2' },
+    ]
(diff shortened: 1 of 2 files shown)


server/tests/libs/aws.spec.js
```diff
@@ -139,6 +139,72 @@ describe('libs/aws/batch.js', () => {
...
+      const done = jest.fn()
+      aws.batch._submitLevels(
+        batchJob,
+        analysisLevels,
+        parallelSubmit,
+        singleSubmit,
+        done,
+      )
+      expect(parallelSubmit).toHaveBeenCalledTimes(1) // One participant level
+      expect(singleSubmit).toHaveBeenCalledTimes(2) // One group level
```
jstiehl: Two group levels.

jstiehl (approved):
I am able to submit jobs and tests pass as expected. There is just the minor issue with the comment in the test file but other than that this looks good to me.

PR #121, opened 2017-10-04
Upgrade to React Router v4 and React 16

### LINKED ISSUE #25: Upgrade to React 15

nellh:
* Paving the way for React 16 and better bundle sizes.
* All components are now ES6 classes or stateless functional components.
* Routing in v4 is now standard React components and top level has been simplified. All existing routes should work.
* Google Analytics has its own higher order component wrapper now to support react-router v4.
* Some top level styling is now in the related component, so they are easier to reuse.
* To get compatible versions I ended up upgrading to latest react-bootstrap and reflux as well.

This fixes #25.

Three known issues that I am working on in their own branches:
* OpBeat is disabled because it was conflicting with a dependency and causing out of memory crashes.
* There's some remaining 'Unknown prop' warnings that do not cause issues but will need to get fixed.


jstiehl:
It looks like those props warnings area actually an issue.  The ones I have seen seem to all be related to bootstrap.  This is causing issues with dataset upload.  I can't upload a dataset and I think it is probably because of the activeKey prop that we use in the bootstrap tabs (I have not dug into this too much yet however).

app/src/scripts/dataset/dataset.store.js
```diff
@@ -1363,14 +1347,10 @@ let datasetStore = Reflux.createStore({
...
-          {
-            datasetId: bids.decodeId(datasetId),
-            snapshotId: bids.decodeId(snapshotId),
-          },
-          { app: appLabel, version: appVersion, job: jobId },
-        )
+        const base = '/datasets/' + datasetId + '/versions/' + snapshotId
+        const query = { app: appLabel, version: appVersion, job: jobId }
+        const url = base + querystring(query)
+        history.push(url)
```
jstiehl, line 1353: I am getting an Uncaught TypeError here for history.push not being a function.  Looks like other methods have a history arg passed but dismissJobModal does not.

nellh: I've pushed a fix for this.

app/src/scripts/client.jsx
```diff
@@ -2,25 +2,17 @@
...
-
-let router = Router.create({ routes: routes, location: Router.HistoryLocation })
-
-RouterContainer.set(router)
-
-Router.run(routes, Router.HistoryLocation, function(Handler) {
-  ReactDOM.render(<Handler />, document.getElementById('main'))
-})
+ReactDOM.render(
+  <Router history={browserHistory}>
```
jstiehl: is specifying history still needed with react-router v4 (when using BrowserRouter specifically)?

nellh: Nope, I've taken this out. Thanks!

nellh:
React 16 is working now. Needs a shim to get modals to work with react-bootstrap, but that can get removed once a [version of react-bootstrap supporting it is out](https://github.com/react-bootstrap/react-bootstrap/issues/2812). Looks like that will be fixed soon with the new portals feature in 16.

nellh:
The submit job redirect is the only part still broken here. Holding off on finishing it to fix a few higher priority issues.

nellh:
The submit job stuff is buggy because the actions no longer make a callback. This is due to the Reflux upgrade needed to fix some of the other incompatibilities with React 16. Need to track down and replace the callbacks with equivalent behavior or child actions, mostly in the modal tools for datasets.

jstiehl:
Pushed a fix for the buggy behavior during submit job. Also tested uploading a dataset, creating a snapshot and running a job just to validate that nothing else might have been impacted. I searched the entire repo to make sure that no history argument is ever sent to toggleModal method that I updated (removed that unused argument) and nothing is passing that arg so I think we are good to remove it.

jstiehl:
Participants are no longer space separated
image

jstiehl:
Parameter inputs are broken now in App Definition form
image

jstiehl:
I updated the Array input component to fix issue shown in last comment.  @nellh could I just get you to double check that the fix won't cause any other issues.  I just updated the change handlers to bind context when passing to child components.

PR #129, opened 2017-10-06
Workarounds for issue #6 for jobs endpoint

nellh:
This hides results and logstreams to reduce the size of the jobs data returned. This improves the bandwidth requirements for issue #6.


jstiehl:
This looks good.  Verified that results are no longer being sent back from jobs endpoint (dashboard) but results are present when drilling down into a specific job/dataset
👍

jstiehl (approved):

PR #137, opened 2017-10-16
Cleanup jobs after 48 hours of run time

### LINKED ISSUE #37: Jobs should automatically be stopped after 48 hours
Jobs running for more than 48 hours should be failed automatically to prevent stuck or very resource intense jobs from running indefinitely.

nellh:
This runs once per hour and checks if the actual run time of any tasks has exceeded 48 hours, then kills up to 100 tasks that meet the criteria. This will fail a job if any part of it takes over 48 hours to run. If lots of jobs are stuck at once, killing 100 at a time prevents overwhelming the Batch API with extra calls.

Fixes #37

### DIFF
--- a/server/libs/aws/batch.js
+++ b/server/libs/aws/batch.js
@@ -125,7 +125,7 @@ export default aws => {
       // Tasks are filtered to only long run times in _terminateOldJobs
       const query = {
         'analysis.status': 'RUNNING',
-        'analysis.started': { $lte: twoDaysAgo },
+        'analysis.created': { $lte: twoDaysAgo },
       }
       const projection = { 'analysis.batchStatus': true }
       c.crn.jobs.find(query, projection).toArray((err, jobs) => {
--- a/server/libs/aws/batch.js
+++ b/server/libs/aws/batch.js
@@ -40,12 +40,29 @@ const extractJobLog = job => {
   return jobLog
 }
 
+/*
+ * Filter for stale jobs
+ */
+const staleJobFilter = (now) => job => {
+  // Kill jobs that are still running and where the actual container start time was 48 hours ago
+  if (
+    'startedAt' in job &&
+    now.getTime() - job.startedAt >= 60 * 60 * 48 &&
+    job.status === 'RUNNING'
+  ) {
+    return true
+  }
+  return false
+}
+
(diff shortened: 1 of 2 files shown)


chrisgorgo:
hey I just want to make sure the following aspects of this feature are there:
- only time spend in "running" should count to 48h
- jobs terminated after 48h should be labelled as such in the user interface (to distinguish them from other crashes)
- all intermediate outputs and logs should be preserved so we would be able to debug why those jobs are freezing

Some (or all) of this might be already in this PR - I just wanted to double check.

server/libs/aws/batch.js
```diff
@@ -87,6 +113,61 @@ export default aws => {
...
+    _cleanupJobs() {
+      // Subtract two days from the current date to search
+      const twoDaysAgo = new Date()
+      twoDaysAgo.setDate(twoDaysAgo.getDate() - 2)
+      // This query will get some jobs we don't want to stop since
+      // those jobs may have spent time in pending or runnable
+      // Tasks are filtered to only long run times in _terminateOldJobs
+      const query = {
+        'analysis.status': 'RUNNING',
+        'analysis.started': { $lte: twoDaysAgo },
```
jstiehl: I am not seeing a 'started' property in my local db.  Is this is supposed to be 'created'?

jstiehl:
I tested this by changing the stale job filter to return true and fmriprep job i was running terminated. also, validated that tests pass for me locally (and, accidentally, that they failed when I bypassed the stale job filter).

jstiehl:
👍

nellh:
@chrisfilo FYI, all that's taken care of except the log message is not displayed. I am working on another pull request with that, the same field has some other useful error cases to report to users as well (parent job failed and container failed to start). The error message that will be displayed is "This analysis task did not complete within 48 hours and has failed due to timeout".

chrisgorgo:
Great! Is it deployed in production or just dev?

nellh:
It's just on dev at the moment.

chrisgorgo:
I run a short job to see if nothing regressed and it worked. Let's deploy it to prod. There are plenty of old "ghost" jobs on prod to hopefully this will deal with.

Issue #133, opened 2017-10-13
Add ability to log in using ORCID


chrisgorgo:
ORCID is and identity management service for academics - see https://orcid.org/. It's an attractive alternative as an authentication service to Google because it's widely used by academics and is not blocked in China.

Integrating ORCID into OpenNeuro will provide an alternative to Google. When a user will want to log in they will be asked to chose between ORCID and Google. Matching users by emails will have to be implemented (so the same internal account will be used independently which login method a user chooses). 

More information about ORCID OAuth: https://members.orcid.org/api/oauth2

anibalsolon:
Hi @chrisfilo I believe I can put some work on this.

I've already requested a sandbox API access for the client identification, but to execute the "usual" OAuth2 two-legged flow, instead of the implicit flow (as it is already used by Google on OpenNeuro).

I looked into ORCID user group, and they are rolling out the implicit flow on beta now. Tomorrow I will try to get more info about this feature with [Eric Olsen](https://orcid.org/0000-0002-5989-8244). This will be a better auth flow for OpenNeuro.

chrisgorgo:
Hi Anibal! Thank you so much for offering to help. We appreciate this very much! 

Let me know if you need anything.

anibalsolon:
Hi @chrisfilo 

I already implemented most of the OAuth2 flow from ORCID on my repo, but I end up in a dependency from Scitran, that uses custom auth providers (https://github.com/scitran/core/blob/master/api/auth/authproviders.py)

Now I will implement the auth provider there, but the deployment of this feature has to be coordinated with Scitran. Do you have any thoughts about it?

chrisgorgo:
Please mind that OpenNeuro is not using the latest versions of scitran, but it's own fork (which is older but includes features such as snapshots and bids support) that you can find in: https://github.com/poldracklab/bids-core.

The full deployment (for testing) can be done via https://github.com/poldracklab/crn_deploy

I hope this helps and thank you again for your hard work!

anibalsolon:
Oh great! It will make it a lot easier.

Thank you for the directions!

nellh:
I can help with any issues you run into getting the site working locally.

[...]

PR #150, opened 2017-10-25
[Feature] ORCID oAuth2 flow

anibalsolon:
In order to support ORCID oAuth2 on OpenNeuro, this PR breaks the current auth flow into two providers (i.e. Google & ORCID) and implement the process to authenticate via ORCID. It also generalizes some implementations in order to support different auth providers.

Linked to https://github.com/OpenNeuroOrg/openneuro/issues/133
Depends on https://github.com/poldracklab/bids-core/pull/17


nellh:
@anibalsolon Thank you for the contributions!

These changes look pretty good but I am running into a few issues:

* I initially got '{error: "Missing e-mail."}' back from the orcid endpoint because my email wasn't public but the error message displayed just says to try again later. I think this case is important to give some user feedback for it.
* Once I'd fixed my email privacy, I ran into ```DOMException: Blocked a frame with origin "http://localhost:9876" from accessing a cross-origin frame.
    at http://localhost:9876/app.bundle.js:108452:38``` on line 96 in orcid.js. Is the localhost path the issue here? FYI, we also have a dev server setup at https://openneuro.dev.sqm.io that I'd like to test on before production deployment and that might need to be configured.
* If a user was previously logged in with a Google account prior to these changes, this is not correctly handled after switching to this branch (navbar show non-functional login buttons and the upload dataset button when it shouldn't). It is working for me with a fresh browser session.

We've also merged another branch with some significant changes (React upgrade) so this will need to be rebased to merge.

anibalsolon:
Hi @nellh !

About your points:
- ORCID does not provide any info if the e-mail is private. So I think that the message could comprehend both cases: user without email and user with private email; such as "Your account does not have an e-mail or your e-mail is not public. Please fix your account before continuing."
- It is an expected error since the script tries to read the current URL on the opened window used for ORCID sign-in. I will add a handler to this error. It is also a limitation for this auth flow, enforcing client and server to be exposed in the same domain:port (such as localhost:9876 or openneuro.org:80).
- I will try to reproduce this scenario, to understand the behavior.

anibalsolon:
@nellh I have fixed the issues you mentioned and merged the upstream to my branch! I also pushed changes to bids-core repo.

Please let me know if it works properly now.

nellh:
@anibalsolon Great, retested everything and all my test cases worked this time. Thanks again for the contributions!

chrisgorgo:
Excellent! Thanks for your help!

anibalsolon:
Awesome! Thank you for your reviews @nellh and @chrisfilo

PR #161, opened 2017-10-30
Display status reason below job status

### LINKED ISSUE #142: Batch status reason is not displayed for failed jobs
Right now the reason message for failures outside of the host container lifecycle is not displayed to users. It should be shown for failed jobs to help with debugging.

nellh:
This will only work for new jobs since we haven't recorded this part of the status. Fixes #142 

screenshot from 2017-10-30 13-28-24


jstiehl:
This works fine for me. My only question/concern is that we are displaying this information in the job status drop down during job execution and this dropdown actually goes away once the analysis has finished running. Would it make more sense to add this to the logs dropdown instead?
image

nellh:
@jstiehl I've simplified the JobAccordion to have just one version of the markup at the top level to always show job status (unless all sub-jobs are successful).

jstiehl (approved):
This looks good to me.

PR #168, opened 2017-10-31
Remove client initiated batch polling

### LINKED ISSUE #68: Polling never updates status when the pending / running time difference is larger than 24 hours
This mostly happens when jobs are waiting in the queue between analysis levels. A participant level job can complete and 24 hours passes before the group level starts. This will get stuck and the status from batch will never be updated.

jstiehl:
* Removing client initiated batch polling since we have server side polling now.
* Client will still poll the server jobs but the handler will only return the current state of the job in mongo as opposed to triggering a poll of AWS Batch for job status.
* this is intended to resolve #68 but will need testing on some longer running jobs.

### DIFF
--- a/server/handlers/awsJobs.js
+++ b/server/handlers/awsJobs.js
@@ -250,7 +250,7 @@ let handlers = {
         return
       }
       //Send back job object to client
-      // server side polling handles all interactions with Batch now therefore we are not initiating batch polling from server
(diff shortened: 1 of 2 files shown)


nellh (approved):
Working for me, going to deploy to dev so we can test some longer running stuff with this.

PR #174, opened 2017-11-01
Resume upload changes

jstiehl:
* fixing the resume upload for the case where the dataset name is that same as the dataset being uloaded


nellh (approved):
This fixes the dead end and works in the all other cases I've been testing.

PR #175, opened 2017-11-01
Login dropdown

Xesme:
Currently the sign in drop down is triggered through an onClick, later I would like to go in and change this. Currently the layout/css is set up in a way that I cannot easily adjust heights without breaking the nav. Absolute positioning and z-index are being used in an effort for efficiency.

### DIFF
--- a/app/src/scripts/nav/navbar.jsx
+++ b/app/src/scripts/nav/navbar.jsx
@@ -20,6 +20,10 @@ class BSNavbar extends Reflux.Component {
   constructor() {
     super()
     refluxConnect(this, userStore, 'users')
+
+    this.state = {
+      login: true,
+    }
   }
 
   // life cycle methods ------------------------------------------------------------
@@ -73,9 +77,7 @@ class BSNavbar extends Reflux.Component {
 
     return (
       <ul className="nav navbar-nav main-nav">
-        <li className="link-dashboard">
-          {isLoggedIn ? dashboardLink : null}
-        </li>
+        <li className="link-dashboard">{isLoggedIn ? dashboardLink : null}</li>
         <li className="link-public">
           <NavLink className="nav-link" to="/public/datasets">
             <span className="link-name">Public Dashboard</span>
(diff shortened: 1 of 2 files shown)


app/src/scripts/nav/navbar.jsx
```diff
@@ -162,22 +162,54 @@ class BSNavbar extends Reflux.Component {
...
-            <i className="fa fa-google" />
-            <span> Sign in</span>
-          </button>
-          <button
-            className="btn-blue half"
-            onClick={userStore.orcidSignIn.bind(null)}>
-            <span className="icon"><img alt="ORCID" width="16" height="16" src="https://orcid.org/sites/default/files/images/orcid_24x24.png" /></span>
-            <span> Sign in</span>
-          </button>
+          <div
```
nellh: I think this can be simpler if it reuses the existing DropdownButton component and styles (like the user menu), especially since it's in the same spot as that menu.

app/src/scripts/nav/navbar.jsx
```diff
@@ -162,22 +162,54 @@ class BSNavbar extends Reflux.Component {
...
+  }
+
+  _loginNav() {
+    return (
+      <div className="dropdown-login">
+        <button
+          className="btn-blue half"
+          onClick={userStore.googleSignIn.bind(null)}>
+          <i className="fa fa-google" />
+          <span> Sign in</span>
```
nellh: One thing I'm wondering about is if the "<icon> sign in" is clear enough what each is for. I took a look at Google's sign in button guidelines and it seems like even if we don't use their color scheme, we might want to adopt the "<icon> Sign in with Google" text and "<icon> Sign in with ORCID" versions for this menu to make each button's provider more obvious.

nellh (commented):
I didn't find any bugs but I added a few comments about how this could be improved.

nellh:
@Xesme I pushed some proposed tweaks to this. I tried making the loading state match the new button and styling it to match the user menu. Maybe we can do this for now and create a backlog issue for something like a sign in modal? It seems like that would be the best thing to do, especially if there are any other providers or sign in methods added in the future.

Xesme:
Agreed. I would love to make a backlog for a modal. Thanks @nellh

nellh:
I'm going to merge this then and we can plan any other improvements here in #176.

Xesme:
Sounds great!

PR #178, opened 2017-11-02
Fix a crash where logged out users could not view public datasets

nellh:
Users who have logged out (but not users who have never logged in) are seeing a crash on any public datasets because localStorage.scitran is set to the string "null" in this check. This fixes the check to work if it is undefined or if a null value is set.


jstiehl (approved):
Test this as follows:
1) Logged out of application
2) Navigated to a local dataset url while on the current master branch and got an error in the console about not being able to read property root of null and no page was loaded.
3) switched to public-dataset-crash-fix branch and page reloaded without any exception.

PR #184, opened 2017-11-03
Restore admin versions of the jobs/dataset components.

### LINKED ISSUE #181: "All jobs" admin tab does not show jobs submitted by ORCID users
Could be related to #180

nellh:
As part of the React 16 upgrade, I removed the route detection from these components and replaced it with a prop but left the prop out of the component instantiation in the admin parent.

Fixes #181


jstiehl:
This looks good to me. Ran a job as an ORCID user and then logged back in with my "admin" account and I can see the job there.
image

jstiehl (approved):

Issue #176, opened 2017-11-02
Improve login flow and UI


nellh:
With multiple login options, we should revisit the UI for this. The current solution removed the single button for either multiple navbar buttons or a dropdown and there's some limitations with either. A modal or login page could support more than one provider better, explain the available features for logged in users, and do a better job of showing the available options.

See #170 and #175 for related issues.

JohnKael:
Additional UI updates to clarify the ORCID capabilities/features for the end user

PR #189, opened 2017-11-03
Login modal

Xesme:
This incorporates fixes for issue #176 and in reference to issues #170 and #175.


app/src/scripts/nav/navbar.jsx
```diff
@@ -11,9 +11,10 @@ import actions from '../user/user.actions.js'
 import { Navbar } from 'react-bootstrap'
 import { Modal } from '../utils/modal.jsx'
 import { refluxConnect } from '../utils/reflux'
-
 import brand_mark from './assets/brand_mark.png'
 
+import logo_app from './assets/logo_app.png'
```
nellh: Looks like this an the related new asset are unused and can be removed.

nellh (changes requested):
The modal looks good and I tested login/logout with both providers.

We should revert the styling for the button in the navbar now that it is a button again (blue call to action).

When you logout, the modal remains open. This state should get reset so the modal doesn't block the page after logout.

app/src/scripts/nav/navbar.jsx
Xesme: Updates have been pushed to fix these issues.

nellh (approved):

nellh:
I think we should merge this after the 1.2 release tomorrow and include it with the other login improvements in the next release, but this is good. Thanks @Xesme

Xesme:
@nellh sounds great. I would like to add the helper-text to this branch as well, we can talk about it when you get in today. 👍

Issue #190, opened 2017-11-05
Editing app parameters does not work


chrisgorgo:
(on prod)
![image](https://user-images.githubusercontent.com/238759/32412199-6cc4c920-c1ad-11e7-973d-1c07ced743cc.png)

```
array-input.jsx:353 Uncaught TypeError: Cannot read property 'setState' of undefined
    at value (array-input.jsx:353)
    at t.value (warn-button.jsx:139)
    at Object.g (react-dom.production.min.js:36)
    at Object.invokeGuardedCallback (react-dom.production.min.js:35)
    at Object.invokeGuardedCallbackAndCatchFirstError (react-dom.production.min.js:35)
    at _ (react-dom.production.min.js:36)
    at Object.executeDispatchesInOrder (react-dom.production.min.js:38)
    at I (react-dom.production.min.js:46)
    at H (react-dom.production.min.js:46)
    at Array.forEach (<anonymous>)
```

nellh:
This fix was deployed with 1.1.4.

PR #193, opened 2017-11-05
Fix event binding regression with editing job parameters

nellh:
Fixes the React 16 regression in #190 and a related bug when canceling out of an edit that led to the edits not being reset.


jstiehl (approved):
Was able to edit and remove a parameter in an app definition

Issue #81, opened 2017-09-20
Enable Firefox support


nellh:
Firefox 49+ supports the needed multifile upload APIs. To enable it, we need to do a once over to fix some minor CSS issues and update the browser check.

- [x] Firefox CSS fixes.
- [ ] Remove check preventing uploads from Firefox

Xesme:
I have gone through mozilla FF css support on newer version of FF. Currently working through process of obtaining older version of FF. Browser stack offers options, but we would need to upgrade our account to access prev version of FF.

- [x] update links for visibility 
- [x] update padding-end for buttons w/in admin
- [x] update selects that are not react-select
- [x] update filter widths to be browser compatible
- [x] comb over css w/in FFv49
- [x] run through upload tests w/in FFv49+

PR #199, opened 2017-11-07
Mozilla support and css updates

Xesme:
This branch adds updates to styling and enables Mozilla Firefox support.  Issue #81

### DIFF
--- a/app/src/sass/crn_app/_elements-wrappers.scss
+++ b/app/src/sass/crn_app/_elements-wrappers.scss
@@ -78,7 +78,6 @@ body {
 
 a {
     color: $primary;
-    font-weight: 600;
     cursor: pointer;
 }
 
--- a/app/src/sass/crn_app/_elements-wrappers.scss
+++ b/app/src/sass/crn_app/_elements-wrappers.scss
@@ -78,6 +78,7 @@ body {
 
 a {
     color: $primary;
+    font-weight: 600;
     cursor: pointer;
 }
 
@@ -740,5 +741,3 @@ select {
 .markdown {
     overflow-wrap: break-word;
 }
-
-
--- a/app/src/scripts/common/partials/happybrowser.jsx
+++ b/app/src/scripts/common/partials/happybrowser.jsx
@@ -31,8 +31,8 @@ export default class Happybrowser extends React.Component {
               site may not work as expected.{' '}
               <strong>
                 <a href="http://www.google.com/chrome/">
-                  Please consider using Chrome or Firefox, V49.0 or higher, as
-                  your browser
+                  Please consider using a minimum of Chrome V49.0 or Firefox
+                  V50.0 as your browser
                 </a>.
               </strong>
             </p>
@@ -62,11 +62,7 @@ export default class Happybrowser extends React.Component {
       ua || window.navigator.userAgent,
     )
     if (
-      bowser.check(
-        { chrome: '41', firefox: '50' },
-        true,
-        ua || window.navigator.userAgent,
-      )
+      bowser.check({ chrome: '41' }, true, ua || window.navigator.userAgent)
     ) {
       // Also allow Chrome 41 even though uploading does not work
       // This is for Googlebot to index correctly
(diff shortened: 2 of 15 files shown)


app/src/scripts/common/partials/happybrowser.jsx
```diff
@@ -56,12 +57,16 @@ export default class Happybrowser extends React.Component {
...
   _incompatibleBrowser(ua) {
     const check = bowser.check(
-      { chrome: '49', chromium: '49', googlebot: '0' },
+      { chrome: '49', chromium: '49', googlebot: '0', firefox: '50' },
       true,
       ua || window.navigator.userAgent,
     )
     if (
-      bowser.check({ chrome: '41' }, true, ua || window.navigator.userAgent)
+      bowser.check(
```
nellh: This check should not include Firefox, it is a fallback for Googlebot (Chrome 41 only).

app/src/scripts/common/partials/happybrowser.jsx
```diff
@@ -31,7 +31,8 @@ export default class Happybrowser extends React.Component {
               site may not work as expected.{' '}
               <strong>
                 <a href="http://www.google.com/chrome/">
-                  Please consider using Chrome, V49.0 or higher, as your browser
+                  Please consider using Chrome or Firefox, V49.0 or higher, as
```
nellh: This should specify Chrome 49 and Firefox 50.

app/src/sass/crn_app/_elements-wrappers.scss
```diff
@@ -78,6 +78,7 @@ body {
 
 a {
     color: $primary;
+    font-weight: 600;
```
nellh: This seems to change the font-weight for all of the analysis views. Is that intentional?

nellh (changes requested):
Have a few screenshots of other issues I'll add to the PR comments.

nellh:
Datasets dashboard filters wrap in Firefox.

datasets-wrap

This "draft dataset has been deleted" message is visible in Firefox but not Chrome.

draft-dataset

Panels in the analysis components have some extra space.

job-status-spacing

In both Firefox and Chrome, the upload dataset tabs are no longer aligned.

tabs

nellh:
I've also noticed two upload bugs with Firefox.

* The check for a dataset with the same name does not work, Firefox will let you upload a dataset with the same name as one uploaded by Chrome and silently fail out of the upload process.
* Somehow "%3A" got prepended to dataset_descripton.json's filename on a successful upload that had been resumed. "%3Adataset_description.json".

Xesme:
@nellh I think we may be having some os/linux styling issues. 

I can see 'This "draft dataset has been deleted" message is visible in Firefox but not Chrome.' within both FF and Chrome.
screen shot 2017-11-07 at 2 17 33 pm

Also, my datasets dashboard filters do not wrap in FF.

screen shot 2017-11-07 at 2 17 49 pm
screen shot 2017-11-07 at 2 18 05 pm

nellh:
@Xesme Is there a good reason to not remove the width on `.filters-sort-wrap .filters` - it seems this fixes the Linux wrapping issue with the dataset filters.

Xesme:
@nellh unfortunately, it was added as a fix to wrapping filters on OS. I will try to find another work around.

Xesme:
@nell originally I had added it to prevent wrapping on OS Mozilla, I just tested it by taking it it out, seems like it's not causing an issue anymore. I think it may be related to the same issue within the upload modal, which was an inherited style that has been changed. I will push a fix right now.

Xesme:
@nell originally I had added it to prevent wrapping on OS Mozilla, I just tested it by taking it it out, seems like it's not causing an issue anymore. I think it may be related to the same issue within the upload modal, which was an inherited style that has been changed. I will push a fix right now.

Xesme:
@nell originally I had added it to prevent wrapping on OS Mozilla, I just tested it by taking it it out, seems like it's not causing an issue anymore. I think it may be related to the same issue within the upload modal, which was an inherited style that has been changed. I will push a fix right now.

nellh (approved):

PR #209, opened 2017-11-10
add defaultChecked to boolean parameters

### LINKED ISSUE #208: No way to set default value for boolean parameters
I've tries `true` and `1`, but neither resulted in a checked box.

Xesme:
MR bug fix #208


nellh:
I've just tested this and did not find any issues :+1:

PR #215, opened 2017-11-15
Add auth for rerun of failed jobs

### LINKED ISSUE #202: Limit ability to rerun jobs only to users who originally scheduled a job and admins
If this is relatively quick to implement lets prioritize it.

jstiehl:
* resolves #202 
* hiding the rerun link unless user is admin or is the user who ran the job
* adding in an auth middleware on server side just to make sure no one without proper access can trigger rerun by hitting endpoint.


jstiehl:
I went ahead and added the fix to the circle.yml file to keep latest tag on docker hub from getting updated on branch and tag builds.

nellh (approved):
Tested this as a couple kinds of users and did not run into any issues.

PR #217, opened 2017-11-16
Fix React 16 regressions with front page pipelines.

### LINKED ISSUE #106: Pipeline views are empty when latest app version has no results
This is mostly a problem right after adding a new version, but since we just did that for all apps... the front page has no pipeline data.

nellh:
This fixes the behavior for the front page pipeline previews to include more results and not crash with React 16.

Fixes #106.


jstiehl:
I don't think this is specifically related to this PR but I noticed that acknowledgments are not showing up for these pipelines either.  Not sure if we want to go ahead and fix this or just make another issue?
image

jstiehl:
Looks like it was a spelling mismatch "acknowledgments" versus "acknowledgments". I'll push up a fix.

jstiehl (approved):
This looks good to me.

PR #219, opened 2017-11-17
Prevent the canceled status from being overwritten by events.

nellh:
@jstiehl  Let me know if this fixes the canceled jobs issue you were seeing.


jstiehl (approved):
This looks good.

PR #220, opened 2017-11-17
Fix bad state after a job is retried

nellh:
This fixes an issue where old jobs are terminated unexpectedly by the job timeout after a retry and cleans up any partial results and batchStatus.


jstiehl (approved):
this updates mongo as expected on retry
image

Issue #201, opened 2017-11-08
Log job retries


nellh:
When a job is retried, that should be logged with the original job metadata and which user initiated the retry.

jstiehl:
Looks like we are not capturing the right data needed on retry to build a link to the dataset page. We should probably use the same batch method "extractJobLog" like we use for initial JOB_STARTED event.
![image](https://user-images.githubusercontent.com/5668826/33400813-0dd40fe4-d50c-11e7-830b-16606961978f.png)

another thing, I know we use that same icon for rerun job on dataset page that we are using here, but I am not sure that it is totally obvious to me on first glance that that icon signifies a rerun job. If that is just me that is fine we can just leave it. Was just curious about other folks' feed back. @nellh ?

PR #221, opened 2017-11-18
Log job retries

Xesme:
Fix for issue #201


jstiehl:
Looks like we are not capturing the right data needed on retry to build a link to the dataset page. We should probably use the same batch.js method "extractJobLog" like we use for initial JOB_STARTED event.
image

another thing, I know we use that same icon for rerun job on dataset page that we are using here, but I am not sure that it is totally obvious to me on first glance that that icon signifies a rerun job. If that is just me that is fine we can just leave it. Was just curious about other folks' feed back. @nellh ?

nellh:
Maybe this is better? I suggested the icon but I agree it's not that clear.



This is just `<span className="label label-info">Retry</span>`.

Issue #226, opened 2017-11-21
Unable to see all files


chrisgorgo:
This seems to be a UI issue. If description of the dataset is short and there are many files you cannot scroll to the bottom of the file list and thus see all of the files:
![image](https://user-images.githubusercontent.com/238759/33049478-2dd2939a-ce15-11e7-8894-8a358ada13e1.png)
Only after zooming out (to the extent scrolling is not necessary) I can see all the files:
![image](https://user-images.githubusercontent.com/238759/33049508-5aea26cc-ce15-11e7-91c4-5709e4828cd0.png)

Observed on Firefox and Chrome.

https://openneuro.org/datasets/ds001105/versions/00001

nellh:
This fix is merged and on dev.

PR #227, opened 2017-11-21
Issue 226

Xesme:
This is a fix for issue #226, where users were unable to see all file directories. In addition, this PR updates select arrows within the jobs modal.

### DIFF
--- a/app/src/sass/crn_app/_common.modals.scss
+++ b/app/src/sass/crn_app/_common.modals.scss
@@ -293,14 +293,38 @@
 
 .analysis-modal {
 
-    .task-select {
-        border: 1px solid $ccc;
-        padding: 0;
-
-        select {
-            border: 0;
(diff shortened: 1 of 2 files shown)


chrisgorgo:
Not sure if this information is relevant, but the original issue was with displaying dataset files not files that are part of job outputs.

Xesme:
Indeed. The dataset file tree is a child of div.dataset-inner by adjusting the bottom margin on the parent you force the overflow children to be visible. :)

Xesme:
Hmm. Yes that is helpful. Thank you!

chrisgorgo:
Awesome. Thanks for clarifying!

Best,
Chris

Xesme:
Double checked this fix. Please let me know if the file tree remains hidden. Checked fix on chrome and firefox.

Issue #191, opened 2017-11-05
Uneven checkbox sizes for boolean parameters


chrisgorgo:
![image](https://user-images.githubusercontent.com/238759/32412212-ad99b7f8-c1ad-11e7-9556-2af067f97095.png)

Issue #157, opened 2017-10-27
Support for single selection inputs


chrisgorgo:
Similar to #156 but with radio button.

chrisgorgo:
@Xesme I tried it out on dev and found couple of issues:
1) I could not find a way to chose which option should be selected by default
2) If I submit a job without touching the radio buttons a list of all options is being reported as input parameters https://openneuro.dev.sqm.io/datasets/ds000123/versions/00001?app=fmriprep&version=23&job=5a3abe241b9c040001dee629

chrisgorgo:
```
fmriprep: error: argument --skull-strip-template: invalid choice: 'OASIS NKI' (choose from 'OASIS', 'NKI')
```

Issue #156, opened 2017-10-27
Support for multiple selection inputs


chrisgorgo:
Some app inputs require multiple selection from a prescribed list. This could be implemented as a series of checkboxes (but not a radio button since it's multi choice).

Here's a failed job that could've benefited from such feature: https://openneuro.org/datasets/ds001123/versions/00002?app=FMRIPREP&version=21&job=b19a7530-ed91-4da2-8b16-dc0adf6ab09d

Xesme:
@nellh I would like to get started on this. Will you add me as an admin, so I can see the example? Thank you :)

PR #229, opened 2017-11-21
Parameter input updates  (App Defs Update)

Xesme:
This branch adds two new input types for app definitions, and fixes input type 'list' to render a select list for users.  

Issues Refernce: 
- [x] #156 Support for multiple selection inputs (multi choice checkboxes)
- [x] #157 Support for single selection inputs (radio buttons)
- [x] #191 Uneven checkbox sizes for boolean parameters 

QA: Did cross browser testing. All updates viable on Chrome and Firefox.


nellh (changes requested):
@Xesme This looks great but it is not passing unit tests. It should at least pass the current tests (looks like you just need to update snapshots) and it would be even better if there was some coverage for the new code.

nellh (approved):

PR #231, opened 2017-11-22
Avoid reading the entire uploaded file into memory for S3 uploads.

nellh:
Needs testing but this should dramatically reduce the memory requirements for uploading datasets to S3.


jstiehl (approved):
This made a massive difference in memory used during upload.

PR #173, opened 2017-11-01
[WIP] Enable opbeat-react

nellh:
Blocked by https://github.com/opbeat/opbeat-react/issues/19 but this is the needed changes.

PR #233, opened 2017-11-28
Integrate Sentry to replace OpBeat

nellh:
OpBeat still does not work in the same page as React 16, this replaces the client side crash reporting functionality and sorts the tracebacks by their source environment in Sentry.

Obsoletes #173


jstiehl (approved):
Got this up and running locally and it is catching client side errors
image

PR #234, opened 2017-11-28
Quick fix for no dataset title in search results

nellh:
This just sets the page title to 'OpenNeuro - <dataset label>' if the label is loaded. When the component unmounts it resets the title to 'OpenNeuro' (currently on every page).


jstiehl (approved):
this looks good 
image

PR #235, opened 2017-11-28
Streaming job logs

### LINKED ISSUE #222: Downloaded logs are truncated
Log downloads should provide full logs, but instead are truncated (same as logs shown in the web UI). Example:
https://openneuro.org/datasets/ds001033/versions/00001?app=FMRIPREP&version=32&job=f0e26bac-f9ec-4c3c-a176-358d9ef4bbef

nellh:
This refactors all the logs endpoints to return streaming results, so we don't build up large job log blobs and run out of memory when download all logs is used.

Logs begin streaming as soon as CloudWatch has returned them, so there's less of a delay opening logs in bandwidth constrained scenarios since the client can download them earlier.

Fixes #222.

### DIFF
--- a/server/libs/aws/cloudwatch.js
+++ b/server/libs/aws/cloudwatch.js
@@ -115,41 +115,65 @@ export default aws => {
     },
 
     /**
-     * Get all logs given a logStreamName or continue from nextToken
-     * callback(err, logs) returns logs as an array
+     * Fetch logs based on a CloudWatch stream name
+     * 
+     * @param {string} logStreamName - CloudWatch stream to read
+     * @param {boolean} truncate - limit results to 1000 lines
+     * @param {function} callback - Callback each time logs are received 
+     * 
+     * @returns {Promise} - Resolves when all logs have been retrieved
      */
-    getLogs(
+    getLogs(logStreamName, truncate, callback) {
+      return new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


server/libs/aws/cloudwatch.js
```diff
@@ -114,54 +113,82 @@ export default aws => {
...
+     * Fetch logs based on a CloudWatch stream name
+     * 
+     * @param {string} logStreamName - CloudWatch stream to read
+     * @param {boolean} truncate - limit results to 1000 lines
+     * @param {function} callback - Callback each time logs are received 
+     * 
+     * @returns {Promise} - Resolves when all logs have been retrieved
+     */
+    getLogs(logStreamName, truncate, callback) {
+      return new Promise((resolve, reject) => {
```
jstiehl: Since we are going to do an implementation it might be worth making the getCloudwatchLogs non-recursive. This might help a little with confusion about about passing a callback, resolve and reject functions to the method.

jstiehl:
Refactor is working. Even though it is a do/while loop now it does seem less confusing that the resolve/reject/callback approach :-)

chrisgorgo:
Could the deployment of this fix in prod be prioritized? Our ability to debug apps is a little bit impaired now (not all testers have access to AWS console).

PR #237, opened 2017-11-30
Refactor docker images to use top level context and include docker-compose support

nellh:
* Moves context to repo root so .git is accessible during builds within containers (reproducible self versioning builds).
* Adds the crn_deploy repo configuration so it's easier to keep in sync with the Dockerfiles for each service.
* Builds containers locally when running the dev version of docker-compose. `docker-compose build` will now let you test local changes to the containers.


jstiehl (approved):
was able to get the app running locally

Issue #238, opened 2017-12-01
Downloading large datasets can exhaust server resources


nellh:
Active and canceled downloads of all analysis results stack up and waste server resources, sometimes leading to partial downloads or downtime.

nellh:
Have not seen any download memory issues since deploying #240.

PR #240, opened 2017-12-01
Improve memory usage for downloading results by using a stream object

nellh:
This is part of the fix for #238 - this prevents entire S3 objects from getting copied into memory.


jstiehl (approved):
Tested this on a small data set and bids-example and no issues downloading all results.

PR #241, opened 2017-12-02
Update node base image and fix path issues with dev configuration

nellh:
This should fix the code not reloading with webpack dev server locally. I also updated to the current release of Node 6.x in the containers, since the 6.11.3 version has been failing to pull.


Xesme (approved):
Looks great. Working on my local.

Issue #245, opened 2017-12-05
Downloading results does not work


chrisgorgo:
The proposed download file name is `fileName` and the download fails immediately with "Server problem"

chrisgorgo:
Tested on https://openneuro.org/datasets/ds001147/versions/00004?app=FMRIPREP&version=32&job=22de4ecf-e568-461d-8110-b132dbae97e1

nellh:
```
server_1  | /srv/server/node_modules/aws-sdk/lib/request.js:31
server_1  |             throw err;
server_1  |             ^
server_1  | 
server_1  | RangeError: Maximum call stack size exceeded
server_1  |     at Object.stringify (native)
server_1  |     at Function.Module._findPath (module.js:158:25)
server_1  |     at Function.Module._resolveFilename (module.js:467:25)
server_1  |     at Function.Module._load (module.js:417:25)
server_1  |     at Module.require (module.js:497:17)
server_1  |     at require (internal/module.js:20:19)
server_1  |     at Object.userAgent (/srv/server/node_modules/aws-sdk/lib/util.js:34:43)
server_1  |     at HttpRequest.setUserAgent (/srv/server/node_modules/aws-sdk/lib/http.js:111:78)
server_1  |     at new HttpRequest (/srv/server/node_modules/aws-sdk/lib/http.js:104:10)
server_1  |     at new Request (/srv/server/node_modules/aws-sdk/lib/request.js:319:24)
server_1  |     at makeRequest (/srv/server/node_modules/aws-sdk/lib/service.js:190:19)
server_1  |     at svc.(anonymous function) [as getObject] (/srv/server/node_modules/aws-sdk/lib/service.js:466:23)
server_1  |     at _async2.default.eachSeries (/srv/server/handlers/awsJobs.js:352:41)
server_1  |     at /srv/server/node_modules/async/dist/async.js:3083:16
server_1  |     at replenish (/srv/server/node_modules/async/dist/async.js:946:17)
server_1  |     at iterateeCallback (/srv/server/node_modules/async/dist/async.js:931:17)
```

I'm looking into the cause of this.

nellh:
@chrisfilo 1.3.1 fixed this issue but revealed another download problem where you could only download one thing at a time. I pushed 1.3.2 to resolve that issue.

chrisgorgo:
Is it deployed on prod? I'm still getting 'fileName' and "Server problem".

Best,
Chris

[...]

PR #247, opened 2017-12-05
Replace archiver with yazl

nellh:
The goal here is to fix the blocking case where aws streams and archiver zips deadlock.

This also removes one of the agave endpoints that depends on archiver. I think we should start cleaning these up anyways.


jstiehl (approved):

nellh:
Does not fix everything for #245 but still an improvement - so merging this.

PR #249, opened 2017-12-05
Switch to using s3-stream-download as input stream for results

nellh:


jstiehl (approved):

nellh:
When a zero byte file was in the results, the archive would hang there. I've updated this to catch the error and continue in this case.

Issue #256, opened 2017-12-07
Unset input parameters of any kind look like boolean 'false'


chrisgorgo:
![image](https://user-images.githubusercontent.com/238759/33725181-409201a0-db26-11e7-81d4-c260c2b0d6a9.png)

Maybe we could have an explicit label for unset parameters?

Xesme:
Currently unchecked booleans are treated as false by the server and changing this would be confusing to the user. In the instance where a user checks and then unchecks a checkbox a false value would be visible to the end user instead of unset.

chrisgorgo:
In this example `type_scanner` is a string and `delay_in_tr` is a number. Both were not set.

Xesme:
@chrisfilo could you link me to the example?

chrisgorgo:
https://openneuro.dev.sqm.io/datasets/ds000123/versions/00001?app=NIAK&version=5&job=5a29689928581e00017d3496

Best,
Chris

Xesme:
Thanks @chrisfilo ! I have updated the logic to render booleans and other input types differently.

PR #258, opened 2017-12-07
update conditional to render bools differently than other input types

Xesme:
fix to resolve issue #256 where all unset input types = false.


nellh (approved):
This is working except for the boolean/Boolean issues that will be fixed in another PR.

PR #269, opened 2017-12-13
Fix ORCID logins for Chrome 63+

### LINKED ISSUE #267: Users reporting broken ORCID login
After login on orcid servers they get redirected to:

https://openneuro.org/crn/users/signin/orcid?code=XXXXX

Which gives 400 (I redacted the code)

nellh:
1d4a455 is the fix, Chrome 63 returns a different URL for the error 400 request because it's a failed request. This request should succeed since it's the end of the oauth redirect flow, returning an empty 200 makes it work again in Chrome 63+

a63800a replaces the polling loop that checks for the redirect URL in the login window with an event handler for when the window closes and a document to trigger it.

Fixes #267

### DIFF
--- a/app/src/scripts/utils/orcid.js
+++ b/app/src/scripts/utils/orcid.js
@@ -76,7 +76,7 @@ let orcid = {
       'ORCID',
     )
 
-    const pooling = callback => {
+    this.oauthWindow.addEventListener('beforeunload', () => {
       try {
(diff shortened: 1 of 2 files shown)


jstiehl:
👍 
Tested on Firefox and Chrome and logged in with ORCID with no issues.

PR #270, opened 2017-12-13
185 fix bids core docker config variables

olgn:
Updated the environment variable names for google auth in config.env.example. 

Changed all references to these environment variables throughout node server + app, as well as in the readme.

Created a .prettierignore file to ignore the app/webpack.* files so lint doesn't break them.


nellh (approved):
I ran into some issues with orcid variables not matching up with the new names in the server but after fixing that this works well. :+1:

olgn:
doh, doesnt help to update the variable names if its to the wrong thing, eh? thanks for catching this!

PR #272, opened 2017-12-15
197 human readable orcid identifiers in jobs

olgn:
updated getDatasetJobs function to tie metadata of the user that submitted the job. this is stored in job.userMetadata.  -- server/handlers/jobs.js

utilize run.userMetadata.email in jobs side panel to display the email of the user that ran the job, if it exists. (note the header in commit c6bb86e should say userMetadata, not user.metaData) --
 app/src/scripts/dataset/dataset.jobs.run.jsx


nellh (approved):
I think this looks good. I'm going to merge this and the related #275 after I rebase them.

nellh:
Merged in a94694d03c4800a950180a53611b95bb6cad8ac2

Issue #273, opened 2017-12-15
Human readable ORCID identifiers in analysis tabs


olgn:
replace orcid with user email for 'by'field : 

public dashboard / public analyses,

my dashboard / my analyses

PR #275, opened 2017-12-16
273 human readable orcid in analysis tabs

olgn:
issue #273 

made the job footer 'by' field user email address by default.


nellh:
This is merged as well.

PR #276, opened 2017-12-18
NavMenu component refactor and test improvements

nellh:
This refactors the _navMenu template into its own component to make testing it easier, along with upgrading Enzyme and Jest and catching warnings in tests.

Also disables Raven wrapping warnings/errors which means they do not appear as breadcrumbs. This might be something we want to turn on in production builds only - but for now I've just turned it off.


jstiehl:
Tests pass locally and navigation seems to work as expected still. 👍

jstiehl (approved):

PR #277, opened 2017-12-18
Fix NavMenu eventKey prop warning

nellh:
The eventKey prop is not accomplishing anything but producing this warning. The test here is mostly useful as an example of a test that required mounting the component over shallow rendering but it does catch this issue with the changes in #276.


jstiehl (approved):
Nav still works for me and tests pass locally.

PR #278, opened 2017-12-18
Adding ability to add and delete a subject (directory)

### LINKED ISSUE #159: Support adding new subjects
Possibles routes:

1. Upload a single new 'sub-' directory (one by one or multi select). Related to https://github.com/OpenNeuroOrg/openneuro/issues/54. Not clear if validation prior to upload could be performed.
2. Point to an updated dataset folder and upload all missing files (including new subject data). Similar to resuming upload. Requires users to have a copy of the whole dataset somwhere (not unreasonable).

jstiehl:
* resolves #159  and #26 
* adding ability to add a top level directory (subject) to a dataset using multi file upload
* if a directory of the same name already exists, the upload is not currently allowed (this is how adding existing files works)


nellh (changes requested):
I'm running into one issue with this. Once I add a new subject, I have to refresh the page before I can download or view the new files.

jstiehl:
I fixed the issue with having to refresh for the file links to not be broken. I had to force a reload of the dataset which required me having to add some flags to an existing method in the dataset store. I went ahead and also added the delete directory functionality into this branch as well as I think that is working pretty well at this point. Let me know if you want to discuss the reload or if you have a better idea on how to add the directory without a reload.

nellh (approved):
This is now working as expected and I've tested delete. Found a CSS issue but it's unrelated to this PR - I filed #281 to track that.

PR #280, opened 2017-12-19
Display S3 download URL and instructions for results

### LINKED ISSUE #38: Provide a better way to get large results
Providing the S3 URL for the results and instructions for using that URL to download your results will make getting results out of the system easier.

A dependency of solving this is we should use signed URLs to prevent excessive reuse of the S3 URLs (redownloading the same dataset many times) and protect private datasets.

nellh:
This breaks down the results component some and adds an S3 URL button next to the Download All for analysis results. The button is a regular link so you can copy it / manipulate it with the context menu but clicking it shows some basic instructions with the URL filled in for how to get the data with awscli.

Fixes #38


app/src/scripts/dataset/run/index.js
```diff
@@ -86,67 +91,6 @@ class JobAccordion extends React.Component {
     )
   }
 
-  _results(run, type) {
-    if (run[type] && run[type].length > 0) {
```
jstiehl, line 90: Looks like we are losing the conditional to test for results in the refactor. This is causing the results according to display when job is still in pending.
image

Also, I think it might be causing a crash when a job has no results 
image

jstiehl: image

this is probably a more useful error trace

jstiehl:
This looks good now. No longer seeing the crashes for empty results. Still waiting on a job to run to see some results.

PR #283, opened 2017-12-20
Adding spinner to top level dataset directory during directory upload

jstiehl:
* spinner was currently setup to spin in a non-existent directory. Fixed that issue so now there is a spinner at the top level of the dataset during directory add.  Also, set the file upload limit to 3 for add directory and also to 3 for deletion of directories.


nellh (approved):
Works well and limiting to 3 requests does make it harder to run into the timeout issue.

PR #291, opened 2017-12-21
Disable tools

jstiehl:
* adding disabling of controls during upload for a dataset


nellh:
Cherry picking the first two commits here.

PR #290, opened 2017-12-21
Add progress indicator and navigation warning for directory uploads in datasets

nellh:
screenshot from 2017-12-20 19-59-22

This hides validation until the new directory upload completes and shows a progress bar for the count of files uploaded.

If you navigate away from the page during an upload, you get a browser warning that says `You are currently uploading files. Leaving this page will cancel the upload process.` and you have to confirm this is what you wanted. There's no way to resume an upload if you do interrupt it like this but you can delete the directory and try again or add a missing file.

Fixes #285 and #289 

@chrisfilo This is on dev for testing.

PR #293, opened 2017-12-21
Cancel in-progress uploads if the user leaves the dataset page

nellh:
Also includes the fix for uploading into the wrong dataset if you browsed to another during upload.


nellh:
Adding these to #290 since they should be reviewed together.

PR #294, opened 2017-12-21
7 error boundaries

olgn:
implements method for fixing issues related to 7

create customizable error boundary component. 'message' property controls the text to display when something has gone wrong. 'className' property passes the class string to the div that contains the error message.

wrap items with loading spinner in ErrorBoundaries

add optional 'timeout' property to spinner component. passing timeout to a spinner causes it to wait that length of time, and then throw an error. ideally the spinner is enclosed within an error boundary so the app catches the error and renders the error message contained within the boundary.

### DIFF
--- a/app/src/scripts/common/partials/spinner.jsx
+++ b/app/src/scripts/common/partials/spinner.jsx
@@ -6,27 +6,55 @@ import PropTypes from 'prop-types'
 // component setup ---------------------------------------------------------------------------
 
 export default class Spinner extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { timedOut: false }
+  }
   // life cycle events -------------------------------------------------------------------------
+  componentDidMount() {
+    this._isMounted = true
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false
+  }
 
   render() {
-    let spinner = (
-      <div className="loading-wrap fade-in">
-        <div className="spinner">
-          <div className="spinnerinner" />
+    if (!this.state.timedOut) {
+      if (this.props.timeout) {
+        setTimeout(() => {
+          if (this._isMounted) {
+            this.setState({ timedOut: true }, () => {
+              this.forceUpdate()
+            })
+          }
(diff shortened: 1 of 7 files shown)


app/src/scripts/common/partials/spinner.jsx
```diff
@@ -6,27 +6,55 @@ import PropTypes from 'prop-types'
...
+  }
 
   render() {
-    let spinner = (
-      <div className="loading-wrap fade-in">
-        <div className="spinner">
-          <div className="spinnerinner" />
+    if (!this.state.timedOut) {
+      if (this.props.timeout) {
+        setTimeout(() => {
```
nellh: It's better to do this when the component mounts, instead of checking in render for the mounted state. componentWillReceiveProps is called on the prop change when the component is mounted.

nellh: It also should call clearTimeout when the component is unmounted to avoid stacking timeouts. It would be better to create a new timeout component and wrap the spinner in it, that way if we fix the components that don't bubble up errors later, we can remove the extra component without changing the spinner again.

nellh (changes requested):
The ErrorBoundary component looks good, just needs some fixes for the timeout part.

olgn:
not sure if i did that rebase correctly - tell me if it messes with the work flow. i can just revert to the last commit and do it over if this doesn't work for you.

app/src/scripts/common/partials/timeout.jsx
```diff
@@ -0,0 +1,44 @@
...
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({ timedOut: true }, () => {
+        this.forceUpdate()
+      })
+    }, this.props.timeout)
+  }
+
+  componentWillUnmount() {
+    clearTimeout()
```
nellh: setTimeout returns an id that should be used to clearTimeout(id) here.

nellh (approved):

PR #297, opened 2017-12-22
Rework requests to use promises

nellh:
Simplifying the app request lib to use promises so the client can interact with in progress requests.

### DIFF
--- a/app/__mocks__/config.js
+++ b/app/__mocks__/config.js
@@ -1,0 +1,48 @@
+export default {
+  scitran: {
+    url: 'localhost:9876/api/',
+  },
+
+  /**
+   * CRN
+   */
+  crn: {
+    url: 'localhost:9876/crn/',
+  },
+
+  /**
+   * Authentication
+   */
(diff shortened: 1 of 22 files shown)


nellh:
Access to the request promise is needed for situations where a request is in progress but the app needs to cancel or report on its progress.

This fixes the inconsistent callback / promises usage for the utils/crn and utils/scitran calls, others are left unchanged to keep this from being a bigger PR than this.

A few dependent changes to note: async/await is usable in the app now, config.js is mocked in tests, and Circle's version of node is closer to the right version to match with the built containers.

jstiehl:
Delete directory seems to stop after deletion of three files now. So if there are more than 3 files, the entire directory does not get deleted anymore.

jstiehl (approved):
This seems to be working as expected. I can upload datasets, resume dataset upload, run jobs, publish datasets, share datasets, block users, define apps, add directories, delete directories, login, login with new user...All of things that I could think to test are now working.

PR #299, opened 2018-01-03
154 replace reactable

### LINKED ISSUE #154: Replace Reactable Component
It seems that the Reactable component that we are using to display csvs in the app is no longer supported. This caused the viewing of csv results to break when we did the React 16 upgrade. A fix was implemented but we should explore alternatives to using this unsupported lib.  Perhaps https://github.com/schrodinger/fixed-data-table-2

olgn:
fixes #154 

-replace reactable with react-table 
-add react-table to package.json
-add styling to react-table to look similar to previous page
-update dataset.file-display.jsx to use ReactTable component instead of Reactable


nellh (approved):
I cleaned up the old library and changed the rows to match the old default of 100. This is working great :+1:

PR #301, opened 2018-01-03
#218 - display results only when all jobs are complete

### LINKED ISSUE #218: Handle delay between job events and results/logs updates better
More obvious than it was in the past, you now can see results before a job is completely done running. Sometimes this means a participant is done running on Batch and results are available, but it shows it is still running or hasn't even started in the job status.

The time that this delay occurs can be quite long, the worst case being about an hour. Improving the latency would make the first issue less obvious.

olgn:
fixes #218. 

only display job results tab when results.length == run.analysis.jobs.length


nellh (approved):
This looks good, working as expected.

PR #306, opened 2018-01-04
Return a 500 error and message when the validator exits unexpectedly

### LINKED ISSUE #99: Validation of deleted dataset causes a server crash
The validation endpoint crashes after not finding the files on disk.

nellh:
This catches the error and returns a 500 status code if validation is requested for a dataset that no longer exists or the validator exits unexpectedly.

Fixes #99

### DIFF
--- a/server/handlers/validation.js
+++ b/server/handlers/validation.js
@@ -19,32 +19,42 @@ export default {
     // get project id
     let datasetId = req.params.datasetId
 
+    const generic_failure =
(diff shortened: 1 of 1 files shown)


olgn (approved):
This looks good, I see the 500 error and the matching error message in the network monitor. 

I think it would be helpful to have the user see some sort of message about the validation failing, vs. a hanging spinner. opening a separate issue for this.

PR #309, opened 2018-01-08
Remove fade in class from warn button

### LINKED ISSUE #281: Dataset FileTree controls animate inconsistently
The first control in each set displays instantly and the others fade in.

jstiehl:
* resolves #281 
* removes the fade-in class from the warn button to avoid inconsistent animation between the option buttons (DELETE, VIEW, etc) in the file tree component.


nellh (approved):
I checked if removing this changes the animation in other places and it does but not significantly.

PR #310, opened 2018-01-08
Revalidate if upload has been canceled

### LINKED ISSUE #296: Validation is sometimes inconsistent when a directory is added to a dataset
If you add a directory and close the tab before completion, the draft dataset is not always revalidated. This is because part of this workflow depends on the browser to initiate validation when the upload completes. The server should check for this case and revalidate if no upload is in progress.

jstiehl:
* Resolves #296 
* added an updating tag to dataset during directory upload and checking for uploading state and updating tag to verify whether upload was started and then canceled without validation.


nellh (approved):
This is working and I think this is a good approach.

PR #315, opened 2018-01-09
36 resume s3 uploads

### LINKED ISSUE #36: Large uploads to S3 fail and are not automatically resumed
Sometimes a complete dataset is available in SciTran but it is not successfully copied to S3 when a job is started. This is more likely to occur as the dataset size goes up but can be worked around by canceling and retrying the job. It will usually succeed after this because the upload resumes where it left of and the underlying S3 or connectivity errors are infrequent.

To improve the upload performance and reliability we can move this SciTran to S3 upload to a queue and process the queue in the worker.

olgn:
fixes #36. 

uploads that have at least one part failed cause the job status to be flagged as 'FAILED'.

retrying a job after a failed attempt checks to see if the snapshot exists, so that if the failure came from s3 upload the snapshot is properly uploaded.

### DIFF
--- a/server/handlers/awsJobs.js
+++ b/server/handlers/awsJobs.js
@@ -196,7 +196,7 @@ let handlers = {
         queue.enqueue(
           'batch',
           'startAnalysis',
-          { job: preparedJob, jobId: jobId, userId: userId },
+          { job: preparedJob, jobId: jobId, userId: userId, retry: false },
           () => {
             // Finish the client request so S3 upload can happen async
             res.send({ jobId: jobId })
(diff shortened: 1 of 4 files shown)


server/handlers/awsJobs.js
```diff
@@ -518,35 +516,15 @@ let handlers = {
...
-          )
-          return
-        } else {
-          let jobLog = aws.batch.extractJobLog(job)
-          emitter.emit(
-            events.JOB_STARTED,
-            {
-              job: jobLog,
-              createdDate: job.analysis.created,
-              retry: true,
```
nellh, line 544: This is needed specifically to log retries.

nellh (changes requested):
Job retries are not logged as retries.

server/handlers/awsJobs.js
olgn: fixed with d6c2c84 below. now startAnalysis takes a retry parameter, and emits logs accordingly

nellh (approved):

PR #318, opened 2018-01-09
274 orcid id admin

### LINKED ISSUE #274: Human readable ORCID identifiers in admin tab
replace ORCID id with email in the following places:

Admin / Users

Admin / Blocked Users

Admin / Event Logs

Admin / All Jobs

olgn:
fixes #274.

add orcid email to the admin panel where applicable

- admin > jobs
- admin > users
- admin > blocked users
- admin > event logs

### DIFF
--- a/server/handlers/eventLogs.js
+++ b/server/handlers/eventLogs.js
@@ -1,5 +1,6 @@
 import mongo from '../libs/mongo'
 import config from '../config'
+import scitran from '../libs/scitran'
 
 let c = mongo.collections
 let events = Object.keys(config.events)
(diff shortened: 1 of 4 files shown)


server/handlers/eventLogs.js
```diff
@@ -24,11 +25,27 @@ let handlers = {
...
+
     c.crn.logs
       .find({ type: { $in: events } }, { sort: [['date', 'desc']] })
       .toArray((err, logs) => {
         if (err) return next(err)
-        res.send(logs)
+
+        const userPromises = logs.map(log => {
+          return new Promise(resolve => {
+            scitran.getUser(log.user, (err, response) => {
```
jstiehl, line 36: This works but do we really want to make an api call for every log? Is the goal to find the ORCID entries in the logs and replace those with an actual email? If so, should we only make this api call for the logs that have an ORCID? Seems like over time this could get expensive but perhaps it is not a big deal.

olgn: yea, you're totally right on this one. will update with conditional re: ORCID users

nellh: I filed #322 as a backlog issue to improve this further in the future.

jstiehl: So you think this is ok as is then?

nellh: Yeah, I think it is cleaner to do it this way and fix the performance consideration with #322 since Google users technically have an email field as well (and it could be different in the future).

jstiehl: OK sounds good.

jstiehl:
This looks good! 👍

jstiehl (approved):

Issue #259, opened 2017-12-08
Default value of boolean parameters passed as string


chrisgorgo:
If an app has a default boolean parameter it is passed as string to the command line. For example: `--ica True` instead of  just `--ica`. 

https://openneuro.org/datasets/ds001178/versions/00001?app=MRIQC-ICA&version=3&job=d3f72446-a866-4e47-bf25-60aaf78e80bc

Issue #244, opened 2017-12-05
Add a ability to provide default values for MultiCheck inputs


chrisgorgo:
Currently MultiCheck inputs use default value as "potential" options, but do not allow to preselect certain boxes.

Issue #243, opened 2017-12-05
TypeError: u.defaultValue.split is not a function


chrisgorgo:
I was trying to add a multicheck input with for FMRIPREP with following default options "template fsaverage".

https://sentry.io/openneuro/openneuro/issues/413504414/

```
TypeError: u.defaultValue.split is not a function
  at t.submitJobDefinition (./scripts/admin/admin.store.js:327:47)
  at a (../node_modules/reflux-core/lib/PublisherMethods.js:53:1)
  at a.emit (../node_modules/eventemitter3/index.js:116:1)
  at Function.t.trigger (../node_modules/reflux-core/lib/PublisherMethods.js:71:1)
  at e (../node_modules/reflux-core/lib/createAction.js:81:1)
...
(18 additional frame(s) were not displayed)
```

chrisgorgo:
This turned out to be a weird confluence of incorrect memory requirements (with the error message blurred in the background) and the modal trying to "fix" the values. Unlikely that anyone will run into this again.

Xesme:
I have recreated this bug message. It occurs when there is an error on a new app submission. On the second attempt to submit app definition it will throw error. Looking into how to resolve this issue.

Issue #149, opened 2017-10-24
Default Values not being recognized as values for required parameters


jstiehl:
If a parameter is set as required in app definition and given a default value, the default is not being recognized as a value for that parameter during job submission (this disables start button)
![image](https://user-images.githubusercontent.com/5668826/31972906-f76d1c78-b8d7-11e7-9af7-fc99bc21d5be.png)

PR #324, opened 2018-01-11
[WIP] Input update

Xesme:
This pull request makes the follow updates: 

- [x] #149 Default Values not being recognized as values for required parameters. 
- [x] #256 Unset input parameters of any kind look like boolean 'false.

- [x] #244 Add a ability to provide default values for MultiCheck inputs.
- [x] #243 TypeError: u.defaultValue.split is not a function. 
- [x] #259 Default value of boolean parameters passed as string


Xesme:
The following bugs/issues that are still open. 
- [x] On edit values are not display from the model. 
- [x] The wrong value is pulled through on default value checked. 
- [x] #149  Default Values not being recognized as values for required parameters.
 
- [x] refactor array-input

app/src/scripts/common/forms/array-input.jsx
```diff
@@ -84,6 +94,14 @@ class ArrayInput extends React.Component {
         <div className="text-info">{this.state.helper}</div>
         <div className="form-inline">
           <span>{inputFields}</span>
+          <ParamController
```
nellh, line 63: This might work better if it was a higher level replacement for ArrayInput when dealing with parameters. We're making ArrayInput too general - better to simplify it for what it was originally for (editing the array values in datasets) and have another parameter component with simpler requirements.

app/src/scripts/common/forms/paramTypes/paramController.jsx
```diff
@@ -0,0 +1,280 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Input from '../input.jsx'
+
+class ParamController extends React.Component {
```
nellh, line 5: This should be a parent component and a type specific one since most of the logic is 'if (type == something) { do something behavior }'

nellh (changes requested):
A few general comments, this would be clearer if it was a component for each type, it really could use some unit test coverage to ensure it hasn't resulted in side effects, and this can probably be simplified to depend less on state and event callbacks passed into the input components.

nellh:
One other issue I found is this version of ArrayInput does not work from the datasets page anymore (try to add an author, for example).

app/src/scripts/common/forms/paramTypes/paramController.jsx
Xesme: I have started breaking up all of these input types. It will take me a day or two to get full functionality back up and running.

Xesme: It is on a separate branch called: input-update-refactor. Will merge it into this branch for final review?

PR #325, opened 2018-01-11
305 fix bsStyle

### LINKED ISSUE #305: Fix bsStyle prop warnings
This is a minor regression from the React 16 upgrade that sometimes makes it harder to debug real issues in components triggering this warning. It occurs when react-bootstrap props are passed to non-react-bootstrap components.

olgn:
fixes #305. 

introduces the '<Wrapper>' component.

the react-bootstrap package applies unused properties to their children native DOM elements. this causes the console to output all kinds of annoying messages.

Placing a <Wrapper> component between a react-bootstrap component and any SINGLE native DOM child element will solve this issue. 

If the react-bootstrap component has an array of children, then this will be fixed by placing a '<Wrapper><div class='wrapper'> array-like dom content </div></Wrapper>' as a buffer between bootstrap component and its content.

### DIFF
--- a/app/src/scripts/common/partials/wrapper.jsx
+++ b/app/src/scripts/common/partials/wrapper.jsx
@@ -1,0 +1,18 @@
+// dependencies ------------------------------------------------------------------------------
+
+import React from 'react'
+import PropTypes from 'prop-types'
+
+// component setup ---------------------------------------------------------------------------
+
+export default class Wrapper extends React.Component {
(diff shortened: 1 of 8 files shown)


olgn:
I'm open to a more descriptive name for the <Wrapper> component, such as <FixBootstrapWrapper> or something like that.

olgn:
this was the wrong approach entirely!

PR #335, opened 2018-01-12
286 - App parameters are removed in the wrong order when deleted.

### LINKED ISSUE #286: App parameters are removed in the wrong order when deleted
If you have multiple parameters configured, removing one in the middle of the list deletes the last one.

olgn:
fixes #286.

actually the problem is that when we delete a parameter, the loading spinner on the \< ArrayItem \> below was triggered and did not go away. 

<WarningButton> requires a callback on whatever action is assigned to toggle(). \<ArrayItem\> _remove() was not providing one. 

solved by adding arbitrary callback to the <ArrayItem> _remove() function.

### DIFF
--- a/app/src/scripts/common/forms/array-input.jsx
+++ b/app/src/scripts/common/forms/array-input.jsx
@@ -208,10 +208,11 @@ class ArrayInput extends React.Component {
     this.setState(initialState)
   }
 
-  _remove(index) {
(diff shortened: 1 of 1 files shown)


Xesme:
Looks good! Tested it locally.

Xesme (approved):

PR #336, opened 2018-01-12
implement redis queue for obtaining usermetadata

### LINKED ISSUE #322: Cache server side user lookups from SciTran
When looking up any kind of user profile information, an HTTP request is made to SciTran for each user lookup. This is a small but significant performance penalty for endpoints referencing multiple users. Instead of making the request each time a cache should be maintained for this after the first request and expired as user information changes.

olgn:
fixes #322 

there was a small but somewhat significant performance hit when obtaining user metadata on datasets / jobs / events logs. 

this pr introduces a new getCache() function to the request library, and uses the cacheing middleware in server/libs/cache/userCache.js to return cached user data when server/scitran.getUser() is called (if available), or make the original request and cache the response (if cache unavailable). 

the length of time that a cached response persists is determined by the new setting in server/config.js:44  - config.redis.cacheTimeout.

### DIFF
--- a/server/libs/cache/userCache.js
+++ b/server/libs/cache/userCache.js
@@ -17,10 +17,11 @@ export default {
     })
   },
   store(res, callback) {
-    const data = res
-    const data_store = JSON.stringify(data)
-    const userId = data.body._id
-    redis.setex(userId, config.redis.cacheTimeout, data_store)
+    if (res.statusCode && res.statusCode == 200) {
+      const data = JSON.stringify(res)
+      const userId = res.body._id
+      redis.setex(userId, config.redis.cacheTimeout, data)
+    }
     callback()
   },
 }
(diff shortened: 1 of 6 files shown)


nellh:
This should use cache invalidation instead of a timeout so the data is always correct. The cache will be small, it only needs to change on user data updates.

server/libs/cache/userCache.js
```diff
@@ -0,0 +1,25 @@
+import { redis } from '../redis'
+
+export default {
+  get(req, callback) {
+    const userId = req.json.userId
+    redis.get(userId, (err, data) => {
```
nellh: Instead of just keying this as userId, it would be better to do something like `scitran:user:${userId}` to prevent collisions and allow for future clearing of parts of the cache.

nellh (changes requested):

server/libs/cache/userCache.js
olgn: changed!

nellh (approved):

Issue #313, opened 2018-01-09
Public snapshot redirects


chrisgorgo:
In case someone is not logged in and goes to a URL corresponding to a dataset which draft is not public, but which includes public snapshots a "You are not authorized to view this dataset" is displayed instead of redirect to the snapshot. Example https://openneuro.org/datasets/ds001105 and https://openneuro.org/datasets/ds001105/versions/00001

Originally reported in https://github.com/BIDS-Apps/SPM/issues/11#issuecomment-356255991

PR #348, opened 2018-01-17
Fix missing public datasets for logged out users

nellh:
Fixes a similar issue with getDataset as well (though this still results in the wrong behavior for #313).


olgn (approved):
'tis fixed, indeed

PR #349, opened 2018-01-18
add panel wrapper to null app results

olgn:
continuation of fix for #305 in the case of null app definitions


Xesme (approved):
Tested. It looks good.

PR #351, opened 2018-01-19
redirect to public snapshots when dataset permissions restrict access

### LINKED ISSUE #313: Public snapshot redirects

olgn:
Fixes #313 

In the case a user is not signed in and navigates to a dataset url that is not publicly available, we redirect them to the most recent public snapshot (if it exists).

Similarly, access to non-public snapshot versions will redirect the user to the most recent public snapshot (if it exists).

### DIFF
--- a/app/src/scripts/dataset/dataset.jsx
+++ b/app/src/scripts/dataset/dataset.jsx
@@ -3,7 +3,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import Reflux from 'reflux'
-import { Link, withRouter } from 'react-router-dom'
+import { Redirect, Link, withRouter } from 'react-router-dom'
 import moment from 'moment'
 import { ProgressBar } from 'react-bootstrap'
 import Spinner from '../common/partials/spinner.jsx'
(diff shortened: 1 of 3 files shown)


jstiehl:
This works great for the redirect for users who are not logged in.  If the dataset is public, I am redirected to the latest snapshot that is public. If the dataset is not public I get the not authorized message.  In either of these scenarios, if I proceed to login, nothing happens after I login. I still see the not authorized page or, if I have access to the entire dataset, only the public snapshots show up. I have to refresh to get things back to normal.
view after login
image
on refresh
image
Not sure if we should deal with this now or make another issue for this. I'll let @nellh make the call on that.

nellh:
I think that's another issue. I filed #354 to track it.

Issue #35, opened 2017-09-02
Google indexing


chrisgorgo:
Currently content of OpenNeuro is not indexed by Google properly

![image](https://user-images.githubusercontent.com/238759/29997848-d09e73a4-8fd2-11e7-9e60-163a10f52872.png)

nellh:
Looks like Google renders the page just fine but decides the browser check is the most important text. We can disable it for the Google indexer.

![screenshot from 2017-09-02 11-47-45](https://user-images.githubusercontent.com/11369795/29997936-9885f3b4-8fd4-11e7-95cb-72437944e6c4.png)

nellh:
This is now fixed on dev and will be fixed in the next production update.

![screenshot from 2017-09-05 11-58-07](https://user-images.githubusercontent.com/11369795/30078295-9d7f7584-9231-11e7-9fc0-ac5bdbf387cf.png)

chrisgorgo:
Google still seems to be fixated on the browser check. We might need to wait longer for the index to recreate.

chrisgorgo:
Unfortunately google still does not seem to index the main page properly. Further more very few users come to OpenNeuro via search

nellh:
I think the issue is we're showing that message for the version of Chrome that [Googlebot uses for some indexing](https://developers.google.com/search/docs/guides/rendering), even though this was fixed for the Googlebot user agent. To work around this, I whitelisted Chrome 41 even though the upload feature does not work on Chrome 41. Any other version before 49 will still show the upgrade warning.

chrisgorgo:
Indexing still seems to be problematic
![image](https://user-images.githubusercontent.com/238759/34729008-7f1386d4-f529-11e7-9765-0d9ca078a9b4.png)

nellh:
I've requested indexing again. Google does not show the warning when using fetch as Googlebot:

![screenshot-2018-1-9 search console - fetch as google - https openneuro org](https://user-images.githubusercontent.com/11369795/34734894-54e26170-f522-11e7-9f2d-d5b6c7f8e781.png)

![screenshot-2018-1-9 search console - fetch as google - https openneuro org 1](https://user-images.githubusercontent.com/11369795/34734899-5922d7d8-f522-11e7-994c-34b96e978f21.png)

nellh:
It might be because this text is only hidden with CSS. I've opened #314 to hide it from the markup in this case to prevent indexing of the invisible text.

nellh:
[...]

PR #356, opened 2018-01-23
Add meta tag descriptions to front page and datasets

nellh:
Also fixes handling of page titles to use the same system and adds distinct page titles for dashboards.

Improvement to #35

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -35,6 +35,7 @@
     "react-bootstrap": "0.31.3",
     "react-dom": "16",
     "react-ga": "^2.2.0",
+    "react-helmet": "^5.2.0",
     "react-router": "4",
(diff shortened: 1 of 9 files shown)


nellh:
@chrisfilo For the dataset descriptions, I'm using the README unless it's not provided, then the label is used. This is less than ideal though, Google only looks at the first 300 characters. Duplicating the label also isn't that useful though - since it is already indexed as part of the page title.

chrisgorgo:
I think this is the best solution for now. Hopefully README's will start with general introduction to a dataset. Another option would be to include  the summary produced by the validator in the description.

olgn (approved):
* home page title / description meta tag as expected

* dataset title / description meta tag as expected 

* jobs title / description meta tag behave as expected

PR #360, opened 2018-01-25
330 editable .tsv

### LINKED ISSUE #330: Ability to edit TSV files
Similar to #329. Online JS based editor of TSV files.

olgn:
fixes #330 

replaces the react-table with react-data-grid as a .tsv and .csv viewer.

also implements work flow for editing / saving files. 

features: filter rows, sort rows, edit cell, add row, delete row (must have a row selected), paste data from external spreadsheet file (must use same separator as original file), cancel edit, and save changes.

saved files will be subject to revalidation. 

only files associated with unsaved datasets are editable.


nellh (approved):
This is pretty good. The react-data-grid import could cause some issues later but we can update that once upstream is fixed.

The usability of adding/removing rows could also be improved (it is pretty awkward) but the rest is pretty solid and understandable.

PR #367, opened 2018-01-26
Input update refactor

Xesme:
This is pull request makes updates to branch: **input-update**


nellh:
@Xesme Let's just close the old PR. I pointed this one at master. Looks like there's also some linting issues to fix.

Xesme:
@nellh sounds good. I will delete the old PR now. 

The linting issues are outlined within ci/circleci, right?

nellh:
One issue I've just run into, you shouldn't be able to add or remove 'participant_label' as a parameter if it's implicit.

Xesme:
@nellh indeed! I didn't think of that. I will set up an error for tying to add participant_label as a parameter on app definition

nellh:
Adding a radio parameter with three options, I get this warning when selecting the options from the job run dialogue.

> Warning: A component is changing an uncontrolled input of type radio to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

Xesme:
Hmm. I will look into that. They should have an assigned value. I will see why they are triggering uncontrolled.

nellh:
One more regression from master is that participant_label should use all participants if none are selected. This might just be that the implicit version of the parameter is set to required when it shouldn't be.

Xesme:
oh! I thought it was a required parameter. updating now.

Xesme:
All prior bug and linting issues have been addressed.  Please let me know if more occur. 
@olgn & @nellh

nellh:
There's some issues with checkboxes and multi-select inputs. Each parameter should relate to one command line argument and booleans with only one option are currently not working. These two should probably not be the same type.

A boolean parameter should produce an argument like this `--key` if true or leave it out if false. This is represented on the server side by {key: 'string', value: true || false}.

A multiselect with one option should produce `--key option` if the only option is selected. Otherwise empty. With two options you should be able to input any combination of `--key one`, `--key two`, `--key one two`, or no argument. This is represented by {key: 'string': value: ['one, two']} to indicate the selected choices.

Trying to build both of these into the boolean parameter type is overloading the intended behavior and deprecating the multiple choice option is going to break some apps. Taking a current app and editing the boolean arguments leads to some broken values as well - so this is hard to fix even if it were done manually.

nellh:
I went ahead and put together a fix for the last issue I found. I reintroduced the boolean / single checkbox parameter and simplified the default value setup. I also split out the Radio input since most of the logic in the CheckboxOrRadio component was detecting which version it was.

nellh (approved):

Issue #358, opened 2018-01-24
https://openneuro.org/ leads to a blank page on Firefox Q


chrisgorgo:
On the console:

```
TypeError: o.default is undefined
[Learn More]
app-7ddb7633.bundle.js:10:44795
getORCIDProfile
https://openneuro.org/app-7ddb7633.bundle.js:10:44795
getCurrentUser
https://openneuro.org/app-7ddb7633.bundle.js:433:355759
init
https://openneuro.org/app-7ddb7633.bundle.js:433:355516
init
https://openneuro.org/app-7ddb7633.bundle.js:10:22145
init self-hosted:948:17 t
https://openneuro.org/app-7ddb7633.bundle.js:44:45538
t.createStore
https://openneuro.org/app-7ddb7633.bundle.js:44:46122
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:10:21553
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:433:349830
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:19:33452
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:14:26482
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:14:44302
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:433:332617
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:433:330345
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:433:324866
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:433:311926
t
propTypes.js:10:23
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:433:84949
t
propTypes.js:10:23
<anonymous>
propTypes.js:53:2
<anonymous>
https://openneuro.org/app-7ddb7633.bundle.js:1:2
```

nellh:
This looks like it might be a bug in ORCID login refresh, rather than a Firefox issue. Is this a browser session you might not have accessed the site for a while with? Firefox 57.0.4 is working for me.

chrisgorgo:
[...]

PR #381, opened 2018-01-31
Give the user feedback if their ORCID account is missing name info

nellh:
First name / last name fields are required to register for OpenNeuro via ORCID. This fixes the failure by providing feedback if either of these is missing or not public and avoids a dead end in the login flow.

This is the second fix for #358

### DIFF
--- a/server/libs/orcid.js
+++ b/server/libs/orcid.js
@@ -3,7 +3,6 @@ import xmldoc from 'xmldoc'
 import config from '../config'
 
 export default {
-
   getProfile(token, callback) {
     let data = token.split(':')
     if (data.length != 2) {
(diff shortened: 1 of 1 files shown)


olgn:
* feedback for missing orcid lastname is present

* feedback for missing orcid email is present

* feedback for missing firstname is impossible, as it is not possible to remove first name from user profile in orcid. but the logic is the same

olgn:
one thing I think would be useful is to see is links to edit the ORCID profile, similar to here: https://github.com/OpenNeuroOrg/openneuro/blob/42d78a54060a80b62e1c583f0dd2b2b2e00984da/app/src/scripts/front-page/front-page.jsx#L144

olgn (commented):
one thing I think would be useful is to see is links to edit the ORCID profile, similar to situations where email is missing / not public: https://github.com/OpenNeuroOrg/openneuro/blob/42d78a54060a80b62e1c583f0dd2b2b2e00984da/app/src/scripts/front-page/front-page.jsx#L144

@nellh if you consider this a separate issue, feel free to close and merge

Issue #380, opened 2018-01-31
Chrome/Google login issues


chrisgorgo:
See https://neurostars.org/t/events-files-unrecognised-by-openneuro/1249/9?u=chrisgorgolewski

nellh:
A guess for this, if you have an invalid refresh token in localStorage, it looks possible to get stuck rather than being logged out (which is the correct action to take if the refresh token request fails). Each page load tries to refresh the token, fails, doesn't log you out. Clearing local storage would temporarily fix it if this is the cause here.

nellh:
The proposed fix for this is updated on dev, failures during any part of the login for Google accounts will now log you out (clearing bad refresh tokens) and let you try again.

PR #383, opened 2018-02-01
Prevent a hang when refresh tokens fail during login.

nellh:
This is a potential fix for #380.

### DIFF
--- a/app/src/scripts/utils/google.js
+++ b/app/src/scripts/utils/google.js
@@ -46,15 +46,25 @@ let google = {
   },
 
   signIn(callback) {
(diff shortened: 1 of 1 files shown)


olgn:
this looks good. 

I've also committed a change with error handling for cases where signIn and signOut attempts fail. i believe these cases could potentially cause the login screen to hang, as well.

olgn (approved):
this looks good.

I've also committed a change with error handling for cases where signIn and signOut attempts fail. i believe these cases could potentially cause the login screen to hang, as well.

Issue #353, opened 2018-01-22
Create client side archiving prototype


nellh:
The idea here is to generate archives on the client by downloading each file from S3. This requires some investigation for feasibility with large datasets but an idea is presented [here](331#issuecomment-358135570).

nellh:
I've been working on a prototype implementation of this with the intention of having it implemented but not enabled for downloads until the new datasets store is live. Since the S3-only datasets implementation depends on the client getting these files from S3 directly, this is important to validate before more effort is spent there.

The implementation should be possible to use with any S3-API compatible backed datasets storage but is manually hooked up to the S3 bucket used for Batch jobs to test it.

chrisgorgo:
You can also use data in s3://openneuro bucket for testing.

Best,
Chris

JohnKael:
Creating a phase 2 ticket for remaining work

nellh:
This works for small datasets with the branch here. How it's hooked up to the datasets will have to change with #331 but we'll keep this around to implement on top of the new datasets backend with #395 and #396.

PR #384, opened 2018-02-01
Add a service worker and support client side archiving for S3 URLs

nellh:
This is an in progress prototype for #353.

Requests for an S3 bucket URL that returns objects are converted into requests for the actual files and streamed back to the client.


nellh:
@jstiehl Can you take a look at this? I am not aiming to merge it yet but I'd like a second opinion on the approach before we go further with it.

The main limitation of the prototype that would need to be solved for real use is the S3 objects are streamed into browser memory instead of piping them to disk as soon as they start downloading. This means it only works for small datasets and progress isn't shown (the data is retrieved from S3 to memory and then appears to download as fast as your browser can compress the zip file).

jstiehl:
Yep I will take a look today....

nellh:
I've removed the demo links from analysis results and narrowed the scope of URL handling to the configured datasets bucket. The only changes here that should impact the current site are the new transform runtime and the service worker more aggressively caching some files, so this should be able to be merged now.

PR #385, opened 2018-02-02
Fix failure to load analysis due to a race condition

nellh:
This happens when datasets are done loading but the apps request is waiting. This is easier to hit if datasets load very quickly, so local environments can reproduce more easily.


olgn (approved):
looks good!

* replicated problem by refreshing a snapshot with analysis several times until race condition met

* verified that problem no longer occurs with this change

PR #389, opened 2018-02-06
add feedback with position highlighting for failed json validation

### LINKED ISSUE #378: JSON editor feedback is hard to interpret
image

It's not clear how "position" translates to particular place in the file.

I wonder if using http://jsoneditoronline.org/ could lead to better user experience. (It has been recently adopted by G-Node - https://web.gin.g-node.org/G-Node/Info/wiki/News#december). More at https://github.com/josdejong/jsoneditor

olgn:
temporarily fixes #378

should tide users over until there is a react 16 compatible version of jsoneditoronline.org (v6).

now errors in json validation give messages with detailed position information.

### DIFF
--- a/app/src/scripts/dataset/tools/json/jsoneditor.store.js
+++ b/app/src/scripts/dataset/tools/json/jsoneditor.store.js
@@ -44,6 +44,7 @@ let JsonEditorStore = Reflux.createStore({
       errorMessage: null,
       errorDetail: null,
       showErrorDetail: false,
+      highlightPosition: null,
     }
     for (let prop in diffs) {
       data[prop] = diffs[prop]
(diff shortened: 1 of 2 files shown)


olgn:
@chrisfilo I've added a bit more detailed validation messages - let me know if this will be sufficient until we can implement the editor from jsoneditoronline.org

app/src/scripts/dataset/tools/json/jsoneditor.store.js
```diff
@@ -118,8 +119,60 @@ let JsonEditorStore = Reflux.createStore({
...
+    // split the data into separate  lines and take the last
+    const lines = subsection.split('\n')
+
+    // get the last line
+    const lastLine = lines[lines.length - 1]
+
+    // trim off the tabs at the beginning of each line
+    const tabSplit = lastLine.split('\t')
+    const lineContent = tabSplit[tabSplit.length - 1]
+    let caratPosition = lineContent.length
```
nellh, line 147: This is minor but you're inconsistently using `let` and `const`. `let` should be reserved for when you need to reassign a variable, which is pretty rare (only one of the `let` declarations in this function looks needed).

nellh (approved):
Works well.

Issue #372, opened 2018-01-29
Search


chrisgorgo:
With growing number of datasets we need a better search feature. To avoid addition effort and maintenance cost and increase indexability I suggest using [Google Custom Search](https://developers.google.com/custom-search/) (even though we will be forced to display ads). The fact that everything will need to be indexed by Google should help with #35. We can use Structured Data (related to #352, but might required embedding the same metadata in additional format) to provide search result tabs for datasets with different modalities as well as allowing users to perform more specialized searches. Those additional filtering options can be implemented gradually as time permits.

Issue #288, opened 2017-12-20
App catalog view


chrisgorgo:
A page accessible from the top nav bar with tiles representing apps available on the platform. 

Each tile should include:
- name of the app
- rating (1-5 stars) and number of raters
- number of successful analyses run (among private and public datasets)
- tags

There should be an ability to filter and sort over those values.

After clicking on a tile a app page should show up that includes:
- All text fields of the app
- Listing of all the versions and their coresponding docker images
- rating (1-5 stars) and number of raters - editable (logged in users can leave a score)
- number of successful analyses run (among private and public snapshots)
- Listing of all successful analyses (limited to public snapshots)

Some inspiration. 

Google Play
![image](https://user-images.githubusercontent.com/238759/34231927-d4652cc4-e592-11e7-98e0-039e98d6707c.png)

Open Stack
![image](https://user-images.githubusercontent.com/238759/34231968-06b52094-e593-11e7-9de9-577c83176100.png)

Brain Life:
![image](https://user-images.githubusercontent.com/238759/34232009-267292c2-e593-11e7-9662-91aff188efe0.png)

Please add your comments/suggestions below!

PR #392, opened 2018-02-07
[WIP] search

rwblair:
#372 

@chrisfilo put the search bar only where datasets are listed:
screen shot 2018-02-07 at 9 37 07 am

Is this ok? or should it be some where more prominent?

@nellh I have the gcse code in the didmount call, is there a more appropriate place it should live?

Switching between the dataset list and the analyses tab and back again will sometimes cause the searchbar to not show up, but toggling again it will come back. Still trying to figure out why this is happening.


chrisgorgo:
Search results in the future will also include available apps (pending #288). Might be worth putting the search panel (maybe folded?) in the nav bar. @Xesme @jstiehl @olgn any ideas how this could be done in aesthetically pleasing way?

nellh:
This is going to cause issues with the virtual DOM. It appears/disappears as you navigate sometimes because React removes the markup on re-render and this is timing dependent.

It would be better to use the [custom search REST API](https://developers.google.com/custom-search/json-api/v1/using_rest) and add a component that renders an input text box for search, then takes you to a new route with the results once you submit your search (/search/[query]). You would have two components GoogleSearch and GoogleSearchResults. The simplest way to do this is pass the query as part of the route to the results component, make the search request in the GoogleSearchResults componentDidMount  and display a loading state until the query results came back.

You could make a new search store but it isn't needed and would complicate the implementation.

nellh:
I took a look at this so far and I have some suggestions.

The page load can be avoided when you make a search by using a controlled input component instead of a form. You use withRouter to get access to the history prop, then calling history.push() with the query on submit. React router provides a helper here so you can define the route as `<Route name="search" path="/search/:query" component={SearchResults} />` then get the query in the SearchResults component with a `match.params.query` prop.

To avoid the page load when clicking a result, you have to extract the relative URL and use the Link component. I'd check if `window.location` is defined and return an anchor if the search result URL doesn't match the current protocol/hostname and Link with the relative URL if it does. Both are needed in case search results include a URL outside of the React app (like a file inside a dataset or testing the site on alternate origins).

One other thing I noticed is `{search()}` is not equivalent to `<Search />` because the JSX transform actually calls React.createElement when you specify a JSX component. It will work either way but `<Search />` is preferred - it makes it easier to debug what's going on with tools that recognize components and easier to write tests for.

rwblair:
Still need to add the window.location changes.

for the route I'm having the query be appended directly after '/search/' I think this still needs to be put into a GET param; right now searching for a full url does not work.

I moved the search component out of common/partials since that wasn't hot reloading on my local computer. Where should this component live in the end?

How should I trigger a re-render when the search button is hit? It works as expected on the dataset list pages, but on the search page it does not.

Finally still need to differentiate between waiting for a request to come back and no results.

nellh:
I sent you some example code for the search button.

Partials was a good place for it. Are you developing on macOS? You might need the polling configuration from #386 to get webpack to pick up changed files automatically inside of Docker.

rwblair:
closing due to #419

PR #394, opened 2018-02-07
Make the front page accessible for logged in users

### LINKED ISSUE #390: Make landing page accessible after log in
Currently after logging in it is not possible to access the landing page which is quite informative. It would be good to keep the landing page as the default for accessing http://openneuro.org even if one is logged in. Clicking on OpenNeuro logo should take users there.

nellh:
Fixes #390


olgn (approved):
* login stays on home page

* clicking the openneuro brand icon returns logged-in users to the home page

PR #398, opened 2018-02-08
Add a response from cache fetch handler to the service worker

nellh:
This enables some basic offline response for cached assets. The site isn't usable in offline mode but this should improve load stability for unstable connections by using cached assets.

### DIFF
--- a/app/src/scripts/sw.js
+++ b/app/src/scripts/sw.js
@@ -33,16 +33,8 @@ self.addEventListener('fetch', event => {
     } else {
       // Respond from cache, then the network
       event.respondWith(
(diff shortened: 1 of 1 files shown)


olgn (commented):
cache might be returning too much data - the site is returning caches of the requests to almost everything. for example, the requests to datasets/id/jobs are getting cached and returned when i navigate to another page while offline.

nellh:
Updated to not cache on every request, only those specified in the install event.

olgn (approved):
This works much better now. the server responses themselves are no longer being cached, only bundles.

PR #402, opened 2018-02-09
328 dataset comments

### LINKED ISSUE #328: Dataset comments
Users should be able to ask questions and report issues with the dataset via a public commenting system. Email notifications should be sent to the dataset submitter.

olgn:
fixes #328 

creates comment functionality for datasets.

features: 
* add, update, and delete comments
* reply to comments 
* syntax highlighting for code blocks in a few languages
* permalinks to comments are supported via hash in url
* comments are sortable by date - ascending or date - descending
* comments are styled according to site design guidelines
* admin and dataset owners have elevated permissions


app/src/scripts/common/partials/comment-tree.jsx
```diff
@@ -0,0 +1,263 @@
+import React from 'react'
+import Reflux from 'reflux'
+import PropTypes from 'prop-types'
+import moment from 'moment'
+import { withRouter } from 'react-router-dom'
+import Comment from './comment.jsx'
+import WarnButton from '../forms/warn-button.jsx'
+
+class CommentTree extends Reflux.Component {
```
nellh: Is this supposed to be a Reflux component?

nellh (changes requested):
I have some concerns here. When another user deletes their comment, all child comments can disappear. This should probably delete the content and hide the name, not remove the comment entirely. That way child comments remain accessible.

The syntax highlighting support is pretty awkward still. It doesn't always remember the selection made, the controls aren't clearly associated with the code block they affect. If it is complicated to make it autodetect, it would be better to remove it and revisit it later.

We should take a second look at the styling for the actual comments. The body of the comments and tree layout looks good but the comment body doesn't fit with the controls / metadata / user icon. If you take a look at the openfmri comments, the controls are more tightly connected to the content they manage and it feels a lot less like your comment body is floating unrelated.

app/src/scripts/common/partials/comment-tree.jsx
olgn: i) deleted comments now show the [deleted] tag above the comment. admins still have the ability to remove / edit / reply to deleted comments

ii) code syntax select dropdown is now tied to the syntax state of the comment. while this is only single syntax per comment, it now works more along the lines of intuition.

iii) the styling has been tightened up

iv) no, that wasn't supposed to be a reflux component

nellh (approved):

PR #403, opened 2018-02-13
Fix extra snapshot endpoint request bugs in JobMenu component

nellh:
This fixes the extra requests issue discovered when testing the comment branch.


olgn (approved):
* navigating to a snapshot now makes only one call to /api/datasetid/snapshots/snapshotid per visit to snapshot page.

* tested both for users that are signed in / not signed in

* tested for public + private datasets

PR #413, opened 2018-02-14
admins can no longer see deleted comment avatars

### LINKED ISSUE #407: Deleted comments show avatars
image

olgn:
fixes #407 

* admins can no longer see the avatars of deleted comments


nellh (changes requested):
This works but it points out that deleted comments are available from the endpoint anyway. The server should not return the "deleted" comment content or user.

olgn:
@nellh good point - latest commit implements server side filtering on deleted comments

nellh (approved):

PR #417, opened 2018-02-14
allow sign in from comments thread

### LINKED ISSUE #409: Make "login" immediatelly actionable
image

"login" should link to the sign in dialog.

olgn:
fixes #409 

* message displayed on comment thread when user is not logged in: 'Please sign in to contribute to this discussion' now has 'sign in' actionable - the link brings up the sign-in modal.


nellh (approved):
This is good. We might want to make a login route so we can just link to it but this is a good way of doing this for now.

PR #418, opened 2018-02-15
follow datasets feature

### LINKED ISSUE #337: Follow dataset feature
Users should be able to "follow" a dataset and thus receive email notifications about:

- new snapshots of the dataset (the email should include the change log if included - see #327)
- new comments and replies (the email should include the content of the message - related to #328)

Users should be encourage to follow a dataset when clicking the download button.

olgn:
fixes #337

users now have the ability to follow datasets. users that follow a dataset will receive notifications about activity, including snapshot creation and comments

* new item - follow / unfollow - added to the dataset toolbar
* new status - Uploader is Following - added to the status bar and the comments section to notify users when the uploader of the dataset is actively monitoring the dataset
* backend - CRUD operations for followers (subscriptions) added
* new email notifications -  snapshot-created.html, and comment-created.html, fueled by server/libs/notifications.js. These new emails will be sent to all followers of a dataset, and contain links to the appropriate content as well as instructions on how to unfollow the dataset.
* new download workflow - on dataset download, the user will be prompted to sign in (if not signed in already) and then to follow (if not following already). users have the ability to choose to not follow, of course.
* when a dataset is uploaded, the uploading user will automatically be following that dataset


nellh (approved):
I removed one extra dependency from the app. This is working great, tried it as a couple kinds of users, after uploading, sub/unsub, and several kinds of comments.

Xesme (approved):

PR #419, opened 2018-02-15
Search

rwblair:
Adds search and searchResult components to display results from the google custom search api.


rwblair:
Merged in the changes from your branch @nellh and added a catch in case the request fails. From my understanding the request will trigger the catch on any non 200 status code. I have tested it with 403s.

nellh (approved):
Nice, this looks pretty good now.

Issue #421, opened 2018-02-19
ORCID notifications are sometimes sent with wrong email field


chrisgorgo:
When submitting a job on https://openneuro.org/datasets/ds001235 (infinite spinning ball)
![image](https://user-images.githubusercontent.com/238759/36392989-959f1130-1562-11e8-93e9-a28d129a8600.png)

I could not find a relevant entry in Sentry.

chrisgorgo:
The dataset page is not accessible now due to follow up errors. Neither is the log of all jobs: https://openneuro.org/admin/jobs

chrisgorgo:
Actually all dataset pages seem to be inaccessible now.

nellh:
This is a server crash caused by a badly generated notification.

PR #422, opened 2018-02-19
Fix sending emails for dataset sharing to user ids instead of emails.

nellh:
Bug fix for #421.


olgn (approved):
works great! 

* sending email to orcid user sends to account email address

Issue #425, opened 2018-02-20
Snapshot creation modal should be its own view


olgn:
the snapshot workflow is well on its way to being complex enough to warrant its own view. 

links to this view will be accessible through other modals that allow / encourage snapshot generation (eg publish)

Issue #424, opened 2018-02-20
Dataset modals should use routes


olgn:
navigating to different dataset modals should use routes, so we can provide links to modal states.

for example, the publish modal should be accessible through a url like /datasets/:datasetId/publish.

PR #437, opened 2018-02-22
dataset modal routing and snapshot changelog

### LINKED ISSUE #327: Add ability to annotate dataset snapshots with changelogs
When creating a new snapshots users should be able to provide description of changes. Those descriptions should be used to autogenerate CHANGES (see 8.1 http://bids.neuroimaging.io/bids_spec1.0.2.pdf).

olgn:
fixes #327, #424, #425, and #458. 

reason for combining these: when working on #327, it became apparent that the create snapshot modal should be broken into a separate view. 

however, there were modals (publish, jobs) that contained 'create new snapshot' buttons within them.  if a user navigated to the snapshot creation view from a modal, it would be very difficult to keep track of what modal it came from. 

the solution is to create routes for each modal, and utilize these modals to control navigation between snapshot creation, as well as control the activation / deactivation of modals.

changes to modals:
* toggling a modal from the dataset toolbar now links to a separate url, /dataset/datasetId?modal=modalname
* closing a modal links the user back to the /dataset/datasetId url

changes to new snapshot workflow:
* creating a new snapshot now involves specifying a version number (locked at next major version until we implement major, minor, and point versioning) 
* creating a new snapshot now involves specifying a changelog
* saving a snapshot will format changes and attach the new changes to the old changelog, along with version number and date
* the snapshot notification email now contains the updated changelog, to inform followers about updates
* user will not be able to see the 'create new snapshot' link from within publish / jobs modals unless the dataset has been modified


nellh (changes requested):
The query string should be reserved for queries, a verb is better structured as '/datasets/ds000001/create-snapshot'. Hyphens are also preferred so that individual words are more clearly delineated in a URL (especially for search engines). These can also be implemented as actual Route components instead of parsing in componentWillReceiveProps.

Ran into one bug - deleting a snapshot suggests and lets you reuse version numbers.

We may want to hide the extra version parameters if it is not available with the SciTran backend. The disabled boxes are confusing.

olgn:
@nellh all three points of feedback have been addressed. 

* modals now use datseturl/modalname routing methods
* the modals themselves have been turned into components which are rendered within the dataset page
* version numbers are locked at incrementing by one above the last snapshot version, regardless of if it has been deleted or not
* version is set in stone until we implement a method for utilizing a point versioning system

app/src/scripts/common/forms/warn-button.jsx
```diff
@@ -186,3 +196,5 @@ WarnButton.defaultProps = {
   warn: true,
   tooltip: null,
 }
+
+export default withRouter(WarnButton)
```
nellh: It doesn't look like this is needed.

app/src/scripts/common/partials/file-tree.jsx
```diff
@@ -16,8 +16,9 @@ class FileTree extends React.Component {
     let editable = this.props.editable
     let tree = this.props.tree
     let topLevel = this.props.topLevel
-
+    let history = this.props.history
```
nellh: This probably isn't needed either.

nellh (changes requested):
A couple extra router dependencies but this looks pretty good now. yarn.lock might need to be fixed and merge conflicts need to be resolved.

nellh (approved):

PR #441, opened 2018-02-22
update node alpine version

olgn:
this is a test to see if the upgraded node alpine version will compile in circle ci


olgn:
@nellh it worked!

nellh (approved):

PR #445, opened 2018-02-26
Skip waiting to update service worker sooner

nellh:
This replaces the service worker when a new one has loaded, instead of waiting for the next page load.


olgn (approved):
works as advertised

* shutting down a service worker and refreshing the page generates a new service worker.
* stopping a service worker and refreshing the page resumes the stopped service worker

PR #446, opened 2018-02-26
Dependencies and testing changes for new dataset backend

nellh:
This is part of the new dataset backend work, updating NodeJS to 8.9 and adding a test environment for writing model dependent tests.


nellh:
Circle does not work with the mongodb tests, I'm fixing this.

olgn (approved):
looks good! 

* jest tests of mongodb look good
* optional datalad backend does not interfere with current backend
* node upgraded without any apparent issues

PR #449, opened 2018-02-27
Avoid cross origin only-if-cached requests from service worker

### LINKED ISSUE #435: TypeError: Failed to execute 'fetch' on 'ServiceWorkerGlobalScope': 'only-if-cached' can be set only with 'same-origin' mode
Could be related to the fact that icons showing how many jobs have run for each snapshot are not showing up. https://openneuro.org/datasets/ds001236/versions/00002?app=FMRIPREP&version=53&job=7415ba16-d476-4d62-ad64-fae9da437943

image

nellh:
Fixes #435


olgn (approved):
indeed, 'tis fixed

PR #457, opened 2018-03-02
Report unexpected ORCID auth failures to sentry.

### LINKED ISSUE #456: TypeError: Cannot read property 'body' of undefined
https://sentry.io/openneuro/openneuro/issues/468395047/

```
TypeError: Cannot read property 'body' of undefined
  File "/srv/server/libs/orcid.js", line 21, in Request._request2.default.get [as _callback]
    let doc = new xmldoc.XmlDocument(res.body)
...
(8 additional frame(s) were not displayed)

TypeError: Cannot read property 'body' of undefined
```

nellh:
Also shows a generic warning in this case.

Fixes #456

### DIFF
--- a/server/libs/orcid.js
+++ b/server/libs/orcid.js
@@ -1,5 +1,6 @@
 import request from 'request'
 import xmldoc from 'xmldoc'
+import Raven from 'raven'
 import config from '../config'
 
 export default {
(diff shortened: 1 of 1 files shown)


olgn (commented):
Maybe i'm being a bit to careful, but is there any case where the server would not send back an error or a response? if so, maybe if(err || !res) would cover that case? if not, this looks good to go

PR #460, opened 2018-03-06
455 orcid avatars

### LINKED ISSUE #455: ORCID accounts yield empty <img> avatar tags
I logged in with my ORCID account and noticed that my avatar was not displaying correctly

image

It seems that the  tag is missing src

```
<img alt="Krzysztof Jacek Gorgolewski" class="user-img-thumb">
```

olgn:
fixes #455 

users without an avatar on their user profiles (AKA ORCID users) now have a generic thumbnail provided - both for the navbar and the comments section

### DIFF
--- a/app/src/scripts/user/user.store.js
+++ b/app/src/scripts/user/user.store.js
@@ -423,6 +423,29 @@ let UserStore = Reflux.createStore({
         }
       })
   },
+
+  generateThumbnail(profile) {
+    let thumbnail, thumbnailUrl
+    if (profile.imageUrl) {
+      let username =
+        profile && profile.firstName && profile.lastName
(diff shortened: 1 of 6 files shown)


app/src/scripts/user/user.store.js
```diff
@@ -423,6 +423,29 @@ let UserStore = Reflux.createStore({
         }
       })
   },
+
+  generateThumbnail(profile) {
```
nellh: The user store is an odd place for this. It doesn't interact with the data in the store. It really looks like it should be a component named <AvatarThumbnail> or similar.

nellh (changes requested):

nellh (approved):

PR #461, opened 2018-03-07
allow embedded filenames in the file display url

### LINKED ISSUE #404: Encode file previews in URLs
To facilitate sharing specific aspects of each dataset (and job results) when user opens a file preview the URL should be updated. This would allow someone to open a file preview copy the URL and share it with a colleague who would be able to see the same file preview.

olgn:
fixes #404 

viewing files now navigates the user to datasetURL/file-display?file=filename.fileextension

this allows the user to copy the link from the browser, and share with other users (given they have access to the dataset).

### DIFF
--- a/app/src/scripts/dataset/dataset.file-display.jsx
+++ b/app/src/scripts/dataset/dataset.file-display.jsx
@@ -22,11 +22,42 @@ class FileDisplay extends Reflux.Component {
   constructor() {
     super()
     refluxConnect(this, datasetStore, 'datasets')
+    this.state = {
+      fileRequested: false,
+    }
+  }
+
+  componentWillReceiveProps() {
+    let datasets = this.state.datasets
+
+    const file = datasets ? datasets.displayFile : null
+    const fileName = file ? file.name : null
+
+    if (!fileName) {
+      let params = new URLSearchParams(this.props.location.search)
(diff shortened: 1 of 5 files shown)


app/src/scripts/dataset/dataset.file-display.jsx
```diff
@@ -22,11 +22,42 @@ class FileDisplay extends Reflux.Component {
   constructor() {
     super()
     refluxConnect(this, datasetStore, 'datasets')
+    this.state = {
+      fileRequested: false,
+    }
   }
 
-  render() {
+  componentWillReceiveProps() {
```
nellh, line 51: This could just be a route.

nellh (changes requested):
The route approach for these would be better so that inline files are more easily indexed by search engines as their own pages.

app/src/scripts/dataset/dataset.file-display.jsx
olgn: /datasetUrl/file-display/filename instead of /datasetUrl/file-display?file=filename?

nellh: Yes, since each file is its own page in this case.

nellh (approved):

PR #463, opened 2018-03-08
DataLad API integration [1/3]

nellh:
This adds some dependencies and test fixtures for the DataLad APIs and very simple versions of the dataset and snapshot models. Lots of mongo, express, and node testing improvements.


olgn:
i've got an error - cannot find module './handlers/openfmri' originating from the /server/routes.js file

olgn (approved):
* app doesn't appear to be broken by the segmentation of server routes into datalad / scitran

* all new tests successfully passed

PR #465, opened 2018-03-09
json-ld dataset microdata

### LINKED ISSUE #352: Embed schema.org JSON-LD in dataset landing pages
As described in point 7 of https://www.biorxiv.org/content/early/2017/10/09/097196.full.pdf+html. The `@id` field should be set to doi URL when #333 will be implemented (however work on schema.org can start earlier).

When deployed pleas comment on https://github.com/datacite/freya/issues/8

olgn:
fixes #352 

helmet now includes tagged dataset metadata in json-ld format for search indexing. the format conforms with the Dataset schema from schema.org, found [here](http://schema.org/Dataset)

### DIFF
--- a/app/src/scripts/utils/json-ld.js
+++ b/app/src/scripts/utils/json-ld.js
@@ -1,0 +1,63 @@
+const schemaGenerator = datasets => {
+  const dataset = datasets ? datasets.dataset : null
+  const context = 'http://schema.org'
+  const type = 'Dataset'
+  let publisher = 'OpenNeuro'
+
+  if (dataset) {
+    let description = dataset.description
+    let name = description ? description.Name : null
+    let authors = formatAuthors(dataset.authors)
+    let datePublished = dataset.created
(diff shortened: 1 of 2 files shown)


app/src/scripts/utils/json-ld.js
```diff
@@ -0,0 +1,63 @@
...
+    let datePublished = dataset.created
+    let dateModified = dataset.modified
+    let license = description ? description.License : null
+    let version = dataset.snapshot_version
+    let datasetDescription = dataset.README ? dataset.README : dataset.label
+
+    let schema = {
+      '@context': context,
+      '@type': type,
+      '@id': 'linktodoi',
```
nellh: This should probably be removed if there isn't data to uniquely identify it yet.

nellh (changes requested):
Looks good besides the @id field.

nellh (approved):

PR #472, opened 2018-03-12
Snapshot email fixes

### LINKED ISSUE #471: Link from the "new snapshot" email notification leads to an empty page
https://openneuro.org/datasets/ds000115/versions/

image

nellh:
This should redirect any /dataset/:datasetId/versions URLs to a snapshot (which may redirect again if the user does not have access or it has been deleted).

The emails should also have a version number as expected instead of the null URL going forward.

Fixes #471


olgn (approved):
* emails now have proper version number attached
* navigation to datasets/:datasetId/versions now redirects to first snapshot

Issue #453, opened 2018-03-01
"public" icon is not displayed in firefox


chrisgorgo:
![image](https://user-images.githubusercontent.com/238759/36873151-726d9c48-1d5c-11e8-8226-7f7f6decaf70.png)

https://openneuro.org/datasets/ds001245/versions/00001

olgn:
this turned out to be not a mozilla issue, but one dependent on whether the user was signed in or not. \(fixed #480)

PR #480, opened 2018-03-13
Issue 453 - "public" icon is not displayed in firefox

Xesme:
This branch implements the following changes into branch **master**:
 - [x] Removed isSignedIn condition from rendering public icon on datasets

Reference: issue #453


olgn (approved):
looks good!

PR #486, opened 2018-03-14
Issue 371

### LINKED ISSUE #371: Selecting multiple participants closes react-select list between each one
It used to be fairly quick to select a subset of participants but now the menu has to be opened between each selection, making this fairly tedious if you want to run a larger subset.

Xesme:
fixes #371


nellh (approved):
Nice, this was simpler than I thought it might be :+1:

PR #494, opened 2018-03-16
dataset stars

### LINKED ISSUE #414: Dataset stars
⭐️ ⭐️ ⭐️ 

Logged in users should be able to express their admiration for a given dataset. Each user can give (and take back) one ⭐️ per dataset (whole not a snapshot). ⭐️ should be displayed in a prominent place on dataset page as well as the listing of public datasets.

We are going full on social.

### LINKED ISSUE #462: add dataset star count to public dataset list
dataset stars should be visible on the public dataset dashboard

olgn:
fixes #414 
fixes #462 

* users now have the ability to add stars to a dataset (equivalent to a like). 
* the star count of a dataset is visible both on the dataset home page and the dataset dashboard (my, public, and admin dashboards). 
* dataset dashboards are sortable by star count
* there is a new component, \<Metrics>, similar to \<Statuses>, that will hold metrics for a dataset (stars, followers, etc)


nellh (approved):
Tested with several users, adding and removing stars works, sorting is good.

I'm not sure if confirmation is really needed for unstaring but I think this is good as is.

PR #497, opened 2018-03-19
Copy comment link to clipboard

### LINKED ISSUE #427: Links to comments are not easy to find
When clicking 
image
it might be easy to miss the fact the URL was updated in the address bar. A modal with the URL ready to copy to the clipboard would be more obvious.

olgn:
fixes #427 

* clicking the link button on a comment will copy the comment link to the user's clipboard. 
* clicking the link will no longer redirect the user to the comment url
* the link has a useful tooltip
* when the link is copied, a useful 'copied!' flag appears to inform the user they have copied the link to clipboard, and then fades out after a couple of seconds.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -38,6 +38,7 @@
     "raven-js": "^3.20.1",
     "react": "16",
     "react-bootstrap": "0.31.3",
+    "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "16",
     "react-ga": "^2.2.0",
     "react-helmet": "^5.2.0",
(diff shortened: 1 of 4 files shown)


nellh:
I think the icon here could still be updated to be more obvious somehow. The tooltip helps but it's not obvious it is an action now instead of a link. Context menu is also weird for something that looks like it should be a link (since it isn't a link).

olgn:
check it now?

nellh:
I added one more change, to avoid the page jump when you click the button on a page with many comments.

Issue #76, opened 2017-09-18
Add read-only public API


chrisgorgo:
Backwards compatible versions of https://openfmri.org/dataset/api/

Listing all public datasets and their metadata.

This feature is used by nilearn and datalad. We need to make sure we don't break those tools.

KamalakerDadi:
Hi OpenNeuro team, cc @chrisfilo 

My general thoughts on having these metadata like:

- Data versions (because we have multiple versions for each dataset)
   - Example: For dataset ds000001 we have 4 different versions.
   - And adding few more deatils in README document saying what is the difference
between each versions.

- which pipeline a user is interested ?
  - In this pipeline again, which version ? Here may be we can propagate through Job ID.
 Because Job ID is assigned to download url.

Could help users query the data easily to download by writing their own scripts (downloaders).

My thoughts are based on what I have seen so far. I might also be understood wrong. 
Thanks a lot for your amazing work.

chrisgorgo:
https://github.com/INCF/apine

rwblair:
Above link is an idea for a specification for serving more detailed information about files in datasets than is currently provided by the openfMRI api.

rwblair:
Notes on what the old public api returns and what currently in OpenNeuro can be used to give similar results:

|OpenfMRI api fields| Corresponding fields currently in OpenNeuro |
|-------------------|-----------------------------------------------|
| accession_number | dataset id |
| project_name | name |
| summary | readme |
| sample_size | len(subject_list) |
| scanner_type | |
| acknowledgements | acknowledgements |
| license_title | license (license name not explicit) |
| license_url | license (url for license not explicit)|
| ~~curated~~ | |
| publicationdocument_set | digital documents |
| publicationpubmedlink_set | references and links |
| task_set | available tasks (no cognitive atlas id) |
| revision_set | |
| investigator_set | authors |
| link_set | |
| contacts | |

 - Some [places](https://scicrunch.org/scicrunch/data/source/nlx_144048-1/search?q=&l=) make use of the cognitive atlas id so the ability to store that along side the task name might be necessary to add back at some point.
 - link_set is typically a list of links to tarballs on s3. Could this be replaced with links to openneuro.org/api/snapshots/download?
 - the curated field is no longer needed.

nellh:
@chrisfilo Would this still be useful? We added a stub for this during the DataLad implementation that could be finished to provide these APIs on top of the GraphQL backend.

[...]

PR #500, opened 2018-03-19
Datalad API 2/3

nellh:
This includes additional testing improvements, an framework of a schema, but mostly stub resolvers.

Also adds a really basic (it returns one field) version of #76.


olgn (approved):
* new jest tests pass
* GraphiQL visual editor works
* GraphQL server endpoint is responsive to all defined and adequately configured queries + mutations 
* new client-side datalad routes are still inactive

PR #523, opened 2018-03-27
Script to update accession numbers.

rwblair:
Since this was going to be a one off I wrote it as a script that could be called with babel-cli and taking a json file as an argument with the numbers to update. Put it in server/libs not sure if there is a more appropriate place for it to live.

Does not update downloads and analytics collections, have some notes implying these would collections might be replaced by google analytics. If they should be updated let me know and I can update the script.


rwblair:
Here's the working list of accession numbers that will need to be updated on production:
https://gist.github.com/rwblair/b73772b989f94a4c042fb640cf140aca

Unfortunately the accession numbers ds000113b, ds000113c, and ds000113d do not currently work with the script when ObjectId is called since their hex encoding is a character to long.

Issue #85, opened 2017-09-21
Support server side pagination in the client


nellh:
Most endpoints return all data and are only paginated client side. This is part of the issue with #6 but it's a more general issue causing slow performance (especially if you are bandwidth constrained at all).

nellh:
Moving the API side of this to #108

nellh:
This is implemented for datasets.

PR #528, opened 2018-03-30
Issue 503 - Restyle Search

### LINKED ISSUE #503: Improve search styling
button height is not equal to input height: 
screen shot 2018-03-20 at 2 56 18 pm

Search results  page could be improved:
screen shot 2018-03-20 at 2 57 42 pm

Xesme:
closes #503 

Add the following to branch **master**:
 - Fix .search-group button height and alignment.
-  Fix component code structure (ex: life cycle methods) and indentation.
-  Add basic error handling for search input and on navigation null results.
-  Refactored and/or simplified some of the rendering conditionals. 
-  Add title and type of results to result-block.
- Add simple styling to improve readability of result-block.

### DIFF
--- a/app/src/scripts/dashboard/dashboard.searchresults.jsx
+++ b/app/src/scripts/dashboard/dashboard.searchresults.jsx
@@ -43,7 +43,9 @@ class SearchResults extends React.Component {
             <div className="admin clearfix">
               <div className="row">
                 <div className="col-md-5">
-                  {query && this.state.results ? (
+                  {query &&
+                  this.state.results &&
+                  this.state.results[0].link != 'No results' ? (
                     <React.Fragment>
                       <h2>Search Results</h2>
                       <span className="sub-title">
@@ -56,18 +58,11 @@ class SearchResults extends React.Component {
                   <Search />
                 </div>
               </div>
-              <div className="panel-group">
-                {this.state.loading ? (
-                  <Spinner active={true} />
-                ) : query ? (
-                  this._results(this.state.results)
-                ) : (
-                  <h2>
-                    No results found. Please check the url for accuracy and try
-                    again.
-                  </h2>
-                )}
-              </div>
+              {this.state.loading ? (
+                <Spinner active={true} />
+              ) : (
+                this._checkLoadingStatus(this.state.results, this.state.query)
+              )}
             </div>
           </div>
         </div>
@@ -92,7 +87,7 @@ class SearchResults extends React.Component {
           let parsedResponse
           let noResults = [{ link: 'No results', snippet: '' }]
           let searchResults = {}
-          if (!response || !response.text) {
+          if (!response || !response.text || !response.links) {
             searchResults = noResults
           } else {
             parsedResponse = JSON.parse(response.text)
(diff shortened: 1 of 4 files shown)


app/src/scripts/dashboard/dashboard.searchresults.jsx
```diff
@@ -79,58 +122,57 @@ class SearchResults extends React.Component {
...
+    } else if (arr.includes('versions')) {
+      type = 'Dataset - Snapshot'
+    } else if (!arr.includes('public') && arr.includes('datasets')) {
+      type = 'Dataset - Draft'
+    } else if (arr.includes('openneuro.org') && !arr.includes('public')) {
+      type = 'Page - Home'
+    } else {
+      type = 'Page'
+    }
+
```
nellh: This part seems a bit fragile as it will break as pages change. I think it would be better to take this out and improve the titles for the pages indexed, which would improve the search visibility with or without viewing them via OpenNeuro.

app/src/scripts/dashboard/dashboard.searchresults.jsx
```diff
@@ -32,14 +34,55 @@ class SearchResults extends React.Component {
...
+                <div className="col-md-7">
+                  <Search />
+                </div>
+              </div>
+              <div className="panel-group">
+                {this.state.loading ? (
+                  <Spinner active={true} />
+                ) : query ? (
+                  this._results(this.state.results)
+                ) : (
```
nellh: We should avoid nested ternary expressions, this is hard to read. Better to make a function with if statements or a small component if the logic is complex enough.

nellh (changes requested):
Looks good except for the path detection part.

nellh:
One more small issue I noticed is that if your search has zero results, the page shows one result and clicking the "Page" link takes you to a search for "No results"

screenshot from 2018-03-30 12-26-37

app/src/scripts/dashboard/dashboard.searchresults.jsx
Xesme: Ah. Indeed. I was doing the best that I knew how with the limited amount of information being returned from the API. Should I reach out to Ross about updating the titles and possibly adding a type property? I tried it without the type, and the results are very confusing. There isn't a way to tell the difference between a dataset draft, a snapshot, or a file display.

nellh: #85 is part of the issue since some pages are orphaned by it and the index is updated slowly or not at all. Mostly the titles for pages just don't include useful information and I think that's a separate issue, I filed #529 for that.

Xesme: - [x] removed

app/src/scripts/dashboard/dashboard.searchresults.jsx
Xesme: - [x] refactored to be nested conditionals within templating methods

Xesme:
The above issue with null results should be fixed now. I have tested it with numbers, jumbled, symbols, caps, and single letters through the input and the url. Please let me know if I may have missed testing it in any way.

nellh (approved):

olgn (commented):
look good

PR #532, opened 2018-04-02
Check for basename when updating files and add to correct parent path.

### LINKED ISSUE #531: Cannot update files
Steps to replicat:
1. Download a file
2. Try to update the file using the one just downloaded

Result:
image

nellh:
Fixes #531


olgn (approved):
looks good! 

* updating files in a subdirectory is now fixed (given the file has the same name), and files with different names than the original are not successfully uploaded.

Issue #200, opened 2017-11-08
Support CLI uploads


nellh:
Creating a CLI wrapper for the upload feature would let users unable to use Firefox or Chrome still upload datasets and make it easier to script uploading multiple datasets.

The main thing needed to make this happen is a way to authenticate the CLI client. We could either support oauth or an API key for this.

JohnKael:
@chrisfilo how much of a priority is this feature. It's a relatively heavy lift. Was originally geared toward Chinese users, but we don't have a sense for how much of a barrier this actually is (i.e. are they just downloading Chrome, etc)

chrisgorgo:
It's low priority. We upload should be primary mean to get data in.

chrisgorgo:
The CLI would enable us to build a BIDS App for uploading data to OpenNeuro. Such app could be integrated into other data management systems such as Flywheel or Brain Life to facilitate data transfer.

PR #533, opened 2018-04-02
Lerna support

nellh:
This migrates the project to support Lerna packages so some functionality can be more easily shared between the CLI and React clients for issue #200 

WIP because CircleCI is still broken.


nellh:
CircleCI is kind of slower (tests are slower, images are faster) but working now. I'll address the performance in another PR.

olgn (approved):
looks good. the functionality of the app remains intact, and the app and server are now being provided from the lerna packages.

PR #535, opened 2018-04-04
518 references and links

### LINKED ISSUE #518: References and Links produces incorrect formatting
The "References and Links" field on OpenNeuro appears to produce incorrect formatting. Here is the result from just copying and pasting a link into the field, when hide is selected. Behavior is the same whether the text is a link or not:
screen shot 2018-03-27 at 10 32 37 am

olgn:
fixes #518 

* references and links are now updated in the state when changes are made to the description.json file in the json editor
* references are now saved as strings, rather than objects, in the state
* saving the references from the description editor now creates updates to the description file, and update the state with the proper value
* required fields now yell at the user when they don't put in data (for single model input fields)
* this should be fixed for other single-field array inputs, as well

### DIFF
--- a/packages/openneuro-app/src/scripts/common/forms/array-input.jsx
+++ b/packages/openneuro-app/src/scripts/common/forms/array-input.jsx
@@ -425,28 +425,39 @@ class ArrayItem extends React.Component {
     let data = {}
     let types = ['radio', 'multi', 'checkbox']
     for (let field of model) {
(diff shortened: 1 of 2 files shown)


nellh:
Rebased this due to #533

nellh (approved):
:+1: 

Any existing datasets like this are still a bit weird but are easily fixed by editing the messed up row after this patch, so I think we can manually fix a couple datasets.

PR #540, opened 2018-04-05
CI build improvements

nellh:
This fixes several build and test issues related to the #533 merge.

* Upgrades Jest and Husky.
* Uses the Jest projects configuration to test all packages from one Jest process. Fixes --watch at the project root.
* Limits concurrency for Jest processes on CircleCI to two worker processes to avoid out of resources issues.
* Reduces the number of mongodb instances launched during tests.
* Enables type checking errors for JS files, this only affects code editors since we don't use the TypeScript compiler anywhere in our build.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,0 @@
-app/dist
-app/coverage
-app/node_modules
-server/node_modules
-server/coverage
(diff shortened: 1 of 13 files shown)


olgn (commented):
looks good! 

* reconfigured yarn commands work
* 'yarn test tests' the contents of each lerna package (openneuro-app + openneuro-server)

PR #541, opened 2018-04-05
Service worker cache fixes

### LINKED ISSUE #502: Service worker is too aggressive with caching webpack-dev-server bundles
This only affects dev. The bundles are sometimes cached by the service worker when the underlying content changes in webpack-dev-server.

nellh:
This fixes #502


olgn (approved):
can i approve and suggest that you get a raise?

Issue #534, opened 2018-04-03
Some ORCID logins are failing


nellh:
A user reports they get stuck part way through login with the spinner. An account is never created in SciTran, suggesting this failure is negotiation between OpenNeuro and ORCID, prior to creating the user record.

Possibly related to #271 

https://openneuro.freshdesk.com/a/tickets/50

chrisgorgo:
Please close at will - I was not able to replicate the initial issue.

chrisgorgo:
I see on freshdesk that at least one user is still running into this.

chrisgorgo:
https://openneuro.freshdesk.com/a/tickets/50

nellh:
This might be the same issue as #597.

nellh:
Looks like this isn't fixed in #597 - so there is still an open login issue unrelated to that fix.

PR #554, opened 2018-04-09
add timeout to login flow to prevent never-ending spinner

### LINKED ISSUE #271: Login should time out if third party request does not complete
For Google and ORCID, you can get stuck and be forced to refresh the page if your authentication workflow in another window goes away (such as a crash or being blocked by a browser extension). The login flow should time out so you can try again in this case.

olgn:
fixes #271 
potentially addresses #534 

some users are experiencing login errors that cannot be caught due to cross-origin window errors. this serves as a catch-all for login authentication cycles that last too long.

### DIFF
--- a/packages/openneuro-app/src/scripts/user/user.store.js
+++ b/packages/openneuro-app/src/scripts/user/user.store.js
@@ -137,7 +137,22 @@ let UserStore = Reflux.createStore({
       { persist: true },
     )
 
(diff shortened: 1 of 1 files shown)


nellh (approved):
Works great, tested with ORCID and Google logins and existing errors are handled correctly as long as they occur before the timeout.

PR #556, opened 2018-04-10
Openneuro CLI package

nellh:
This adds a WIP OpenNeuro CLI package with command line argument handling.

Login gives a link to directions for getting an API key (directions do not yet exist) and saves it to a common configuration path (platform user directory -> .openneuro) in JSON format.

Upload is a stub in this PR.

This depends on another new package, openneuro-client which is shared logic for the CLI tool and the React frontend.

Example usage for testing:
```shell
cd openneuro/packages/openneuro-cli
yarn run openneuro -h
```

A more complete version will be published to NPM and be installed via `npm install openneuro-cli` or `yarn add openneuro-cli`.

Related to issue #200

### DIFF
--- a/packages/openneuro-cli/.babelrc
+++ b/packages/openneuro-cli/.babelrc
@@ -1,0 +1,15 @@
+{
+  "plugins": [
+    "transform-object-rest-spread"
+  ],
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "node": "current"
+        }
+      }
+    ]
+  ]
+}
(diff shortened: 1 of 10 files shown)


olgn (approved):
this looks good!

I've added the following changes:
* jest mocks now use mkdirp.sync instead of fs.mkdirSync, making config.spec.js tests os agnostic
* update the login prompt to encourage users to obtain an api key from the proper url (openneuro.org/keygen) as per #560

PR #558, opened 2018-04-10
openneuro-client dataset queries

nellh:
This follows #556 - This adds some basic pieces of openneuro-client. Two dataset queries and a create dataset mutation.


olgn (approved):
looks good. looking forward to the implementation of getDataset(), getDatasets(), and createDataset() in the cli actions

PR #560, opened 2018-04-11
api key generator

### LINKED ISSUE #557: Allow users to create API keys
This is a requirement for #200 - users need a way to identify themselves to the site and a very simple way to handle this is to provide a unique key for each user. A user should be able to request a key, see it once, and then a hashed copy is stored with the associated user id.

A client uploading can hash the user's secret key and include the hash to provide authentication without implementing an oauth workflow in the CLI utility.

olgn:
fixes #557 

creates the methods and UI for creating an API key for the openneuro CLI.

* add link to "/keygen" page in the dropdown created by clicking the user avatar (in the navbar)
* add api component for generating new api keys
* add routes and functions for generating api keys on the server side

### DIFF
--- a/packages/openneuro-app/src/sass/crn_app/_api.scss
+++ b/packages/openneuro-app/src/sass/crn_app/_api.scss
@@ -1,0 +1,60 @@
+.api {
+    max-width: 1000px;
+    margin: 50px auto;
+    width: 100%;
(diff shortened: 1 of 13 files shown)


nellh (approved):
This is great! :+1:

PR #561, opened 2018-04-11
increase .html file heights

### LINKED ISSUE #545: HTML files are previewed in a tiny iframe
image
https://openneuro.org/datasets/ds001334/versions/00002/results/openneuro.outputs%2Faad434281fdca1a32bc30ca3c65ccc9e%2Fd34b5b80-3a9e-4cea-afe4-9fb3489e1dce%2Ffmriprep%2Fsub-3003.html

This is on a 1920 × 1080 screen

olgn:
fixes #545 
increase the minimum height of the .html iframes to 750px

### DIFF
--- a/packages/openneuro-app/src/sass/crn_app/_common.modals.scss
+++ b/packages/openneuro-app/src/sass/crn_app/_common.modals.scss
@@ -253,11 +253,8 @@ body > div:nth-child(9) > div.fade.in.modal > div > div > div.modal-body > div >
         }
 
         .display-file-body {
-            padding: 2em;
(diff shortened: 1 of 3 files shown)


nellh:
Is there a way we can fit the iframe to the height of the content? This is definitely better but I think if the page scrolled instead of the iframe, it would be easier to browse these. Full width of the view would also be an improvement.

olgn:
alright @nell that should be good to go. it still scrolls by the iframe but the height is adjustable to the screen height and there is much less padding on the width

PR #568, opened 2018-04-13
Issue 530 - Make dataset accessibility status more prominent

### LINKED ISSUE #530: Make dataset accessibility status more prominent
A large subset of users uploads data to make them publicly available. We had a few users thinking that uploading was enough to publish a dataset. To avoid such confusion we should indicate somewhere on the draft (dataset landing page) that no snapshots are public and ask user if they want to publish them.

Xesme:
Closes #530 

1. Adds a constant status on the dataset (snapshot) page 'Published' or 'Unpublished'.
2. Adds an alert to dataset (draft) page when a dataset has no published snapshots.
3. Clears the alert when the component unmounts.

### DIFF
--- a/packages/openneuro-app/src/scripts/common/partials/status.jsx
+++ b/packages/openneuro-app/src/scripts/common/partials/status.jsx
@@ -26,6 +26,13 @@ export default class Status extends React.Component {
         title = 'Published'
         iconClass = 'fa fa-globe'
         break
+      case 'unpublished':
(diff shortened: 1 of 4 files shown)


olgn (changes requested):
this looks really good. one small change - if you publish the dataset, the notification should go away.

nellh:
Closing this, we can revisit it later.

PR #570, opened 2018-04-13
DOI generation

rwblair:
So, we only want to generate DOIs for public datasets so that the link that the DOI is register with is viewable by anyone. Similarly we wont generate a DOI if one is present in the description.json, the BIDS only allows for one in that field.

An attempt to generate a DOI is made when a snapshot is made public. If this fails for any reason an icon will be present on the tool component to attempt to generate a DOI again.

I did not find a convenient way to update the description.json file with this new DOI, snapshots seem to be immutable right now. To work around this a 'dois' collection was added to openneruo-server where the datasetId and doi are stored. When a dataset is being loaded by the app if no DOI is present in the dataset.description a call to the server is made to see if a DOI is present there.

@nellh Will it be possible with the new backend to make an update to a description.json without having to generate a new snapshot version?

### DIFF
--- a/packages/openneuro-server/handlers/doi.js
+++ b/packages/openneuro-server/handlers/doi.js
@@ -6,24 +6,27 @@ let c = mongo.collections
 
 export default {
   async createSnapshotDoi(req, res) {
+    let doiRes = null
     if (!config.doi.username || !config.doi.password) {
-      return res.send({ doi: null })
+      return res.send({ doiRes: null })
     }
     const datasetId = req.params.datasetId
     let exists = await c.crn.dois.findOne({ datasetId: datasetId })
     if (exists) {
-      let doiRes = exists.doi
+      doiRes = exists.doi
+      return res.send({ doi: doiRes })
     } else {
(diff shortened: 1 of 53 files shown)


packages/openneuro-server/handlers/doi.js
```diff
@@ -0,0 +1,40 @@
...
+      return res.send({ doiRes: null })
+    }
+    const datasetId = req.params.datasetId
+    let exists = await c.crn.dois.findOne({ datasetId: datasetId })
+    if (exists) {
+      doiRes = exists.doi
+      return res.send({ doi: doiRes })
+    } else {
+      await doi.registerSnapshotDoi(datasetId).then(doiRes => {
+        if (doi) {
```
jstiehl: I think we actually want to check doiRes here correct?

jstiehl:
Other than that one conditional check I pointed out, this looks good. I was able to generate a DOI on publish as well as on a dataset that was already public but had no DOI.
image
I checked metadata on metadata store and everything is there. 
image

jstiehl (approved):
This looks good to me.

PR #574, opened 2018-04-16
Basic CLI uploads

nellh:
* Adds the DataLad service to docker-compose
* Uploads to new and existing datasets additively, e.g. `openneuro upload dataset-dir` or `openneuro upload -d ds000025 local-ds25-dir`

See #200


nellh:
This now successfully uploads to the new backend. I will make a few more PRs to finish up #200 but this is the bulk of the implementation. Planning on adding upload progress and the delete function in their own branches.

rwblair:
Ran into an issue with the nginx location for /crn was using the default client_max_body_size and , caused 413 errors when attempting to upload with cli. Also seeing 503 timeouts so proxy timeout settings might also need to be set for location /crn

olgn:
instructions for how to access the cli are given in #556

olgn:
works well. 

* upload -d feature is not active yet

olgn (approved):

PR #598, opened 2018-04-26
App container build paths fix

nellh:
This fixes the repo internal imports made from webpack. This bloats the image but is needed because Docker does not follow the symlinks made in the node_modules directory. With this change you can import sibling dependencies in the docker-compose environment without having to publish on every change.


jstiehl (approved):
Looks like circle build is passing and I am able to run the app locally without issues.

PR #607, opened 2018-05-01
Add a catch to revalidate and clear validating tag

jstiehl:
There was no error handling client side for validation http errors (different than validation failures). This adds a catch to clear the validation flag if there is an error from validation request.

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/dataset.store.js
+++ b/packages/openneuro-app/src/scripts/dataset/dataset.store.js
@@ -1277,21 +1277,30 @@ let datasetStore = Reflux.createStore({
    */
   revalidate() {
     let dataset = this.data.dataset
-    scitran.addTag('projects', dataset._id, 'validating').then(() => {
-      dataset.status.validating = true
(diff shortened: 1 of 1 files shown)


packages/openneuro-app/src/scripts/dataset/dataset.store.js
```diff
@@ -1277,21 +1277,30 @@ let datasetStore = Reflux.createStore({
...
-          validation.errors &&
-          (validation.errors == 'Invalid' || validation.errors.length > 0)
+    scitran
+      .addTag('projects', dataset._id, 'validating')
+      .then(() => {
+        dataset.status.validating = true
         this.update({ dataset })
-        this.updateModified()
+        return crn.validate(dataset._id).then(res => {
+          let validation = res.body && res.body.validation
```
olgn: this will cause validation to be a boolean, but it needs to be the object that sits at res.body.validation

let validation = (res.body && res.body.validation) ? res.body.validation : {}

PR #610, opened 2018-05-01
Add basic DataLad datasets page

nellh:
This adds a limited version of the datasets page using the DataLad backend.

* Tools are disabled.
* Files are a read only preview.
* Summaries and validation results are mocked waiting on https://github.com/OpenNeuroOrg/datalad-service/issues/10

This is enabled with `CRN_SERVER_DATALAD=true`


olgn (approved):
cli:
* cli uploads now require user
* cli uploads now prompt for host
* cli uploads attach user to the datalad dataset

app:
* crn datalad / scitran backend config works splendidly
* minimal dataset page exists
* snapshots can be created with the graphiql page, and those snapshots exist on the dataset page
* uploaded user info is applied to the dataset page
* file tree is generated (albeit not very stylishly) and added to the dataset page
* adds apollo wrapper around the app

PR #616, opened 2018-05-03
Print location instead of file name

chrisgorgo:
Closes https://github.com/INCF/bids-validator/issues/357. I could not fully test it due to local deployment issues. Working on it.


chrisgorgo:
Ready to merge.

Before:
screenshot_72

After:
screenshot_71

Tests are passing, but circle fails due to attempted docker hub push from a fork. This is fixed in #617

Issue #627, opened 2018-05-06
"error Command failed with exit code 1." when uploading a dataset via commandline


chrisgorgo:
```
$ yarn openneuro upload /d/data/ds000051_small/
yarn run v1.6.0
$ node packages/openneuro-cli/src upload D:/data/ds000051_small
? This will create a new dataset, continue? Yes
"ds001182" created with label "ds000051_small"
{ errors: [],
  warnings:
   [ { key: 'SLICE_TIMING_NOT_DEFINED',
       severity: 'warning',
       reason: 'You should define \'SliceTiming\' for this file. If you don\'t provide this information slice time correction will not be possible.',
       files: [Array],
       additionalFileCount: 0,
       code: '13' } ] }
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
Logged in to dev. The dataset did not show up on web UI. For testing purposes I uploaded the same dataset via web UI - it's available here: https://openneuro.dev.sqm.io/datasets/ds001183 (note that the server did use a new accession number).

nellh:
The exit code is because of the validation warning. You can skip warnings with `upload -i` and #581 will make the validator errors clearer.

#629 fixes the issue with a dataset being created even though local validation failed.

chrisgorgo:
I see! It did get slightly further after adding `-i`.

chrisgorgo:
I'm not sure if #629 was suppose to fix this, but I'm still getting this:

Still seeing validation happening after dataset is created:
```
filo@MSI MINGW64 /d/drive/workspace/openneuro (master)
$ yarn openneuro upload /d/data/ds000051_small/
yarn run v1.6.0
$ node packages/openneuro-cli/src upload D:/data/ds000051_small
? This will create a new dataset, continue? Yes
{ errors: [],
  warnings:
   [ { key: 'SLICE_TIMING_NOT_DEFINED',
       severity: 'warning',
       reason: 'You should define \'SliceTiming\' for this file. If you don\'t provide this information slice time correction will not be possible.',
       files: [Array],
       additionalFileCount: 0,
       code: '13' } ] }
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
I'm on eca792620fc4fc09d4b6c7f3a2be7bf624e4a4e6, did `yarn install` and `yarn bootstrap`

nellh:
[...]

PR #629, opened 2018-05-06
CLI: Run validation before creating a dataset.

nellh:
This reorders the steps for a command line upload to avoid creating a dataset if validation has failed.

Related to #627


olgn (approved):
looks good!

* cli uploads trigger validation before creating dataset

Issue #636, opened 2018-05-14
Port web uploader to DataLad backend


nellh:
Tracking issue for remaining fixes for the web uploader on the DataLad backend.

PR #659, opened 2018-05-30
DataLad web uploader

nellh:
DataLad compatible web uploader.

It uses a virtual router to manage the modal state. State is consolidated to the UploadClient component. One major feature of the existing uploader is missing, you cannot see which individual files are in progress, just the upload as a whole. This is due to the limitations of using a multipart request for the uploads, it could be worked around by extending the xhrFetch shim or observing the file objects but I didn't want to get too far into that right now.

See #636

WIP for styling and some navigation issues (you can't jump back as easily as before).


nellh:
Styling is fixed. Navigation should work but is not tied to browser events, so we may need to include some links back to steps if that's needed. Generally simplified the flow so it's always a loop.

olgn (changes requested):
the 'select your folder again' button should reset the state of the uploader to the first step, not redirect to 'uri/#' as it does in this implementation. mostly for this reason - if i go to a page other than the home page (eg dashboard) -> upload dataset -> add invalid dataset -> click 'select your folder again' -> redirects to home page, with the upload modal hidden until clicking upload dataset again

nellh:
@olgn That should be fixed and the prompt when leaving the page (under most circumstances) should be back.

olgn (approved):
looks great! 

tested the following:
* web uploads successfully implemented
* they can be cancelled, too
* switching around within the page does not interfere with web uploads
* warnings are thrown if the user tries to navigate away from the page during an upload
* upload appears more stylish, indeed
* validation happens before upload, and warns / gives user options for invalid sets
* tested in firefox and chrome

PR #698, opened 2018-06-15
Client side streaming downloads for DataLad backend

### LINKED ISSUE #395: Streaming file support for client side archiving
Improvement for the prototype implementation from #353.

Files should not be saved in memory while being compressed, instead they should be streamed to the zip file or written to disk and streamed from disk.

nellh:
This implements all the changes to enable downloading from the new backend.

Requires a cache clear, so remove the Redis container before testing.

* Drafts and private datasets are always streamed via OpenNeuro itself.
* Public snapshots are streamed from S3 once published (as long as S3 is reachable).
* Adds a metadata endpoint that the CLI can use to implement downloads later.

Fixes #395

One optimization that is not included here is draft files that are available on S3 still always come from OpenNeuro itself.

### DIFF
--- a/packages/openneuro-app/src/scripts/datalad/download/download-link.jsx
+++ b/packages/openneuro-app/src/scripts/datalad/download/download-link.jsx
@@ -26,8 +26,10 @@ const downloadClick = (datasetId, snapshotTag) => callback => {
     } else {
       // Something has gone wrong with the service worker
       // or the browser does not support it
-      console.log('An unexpected issue occurred downloading.')
-      console.log(serviceWorker)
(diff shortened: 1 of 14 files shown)


nellh:
#700 is a known issue not intended to be fixed in this PR.

nellh:
Added a fix to save zip files as filenames like `ds001024.zip` and `ds001024-1.0.1.zip`

olgn (approved):
works very well

* downloads are zipped by browser's service worker
* downloads have the name of the dataset + snapshot tag
* tested in chrome
* downloads cannot be cancelled
* helpful error message in firefox if config is not correct

PR #707, opened 2018-06-20
Toggleable analysis

### LINKED ISSUE #596: Allow a site administrator to disable/enable the analysis tools
This should hide the analysis functionality from snapshot pages and disable any other user facing analysis information.

olgn:
fixes #596 

allows for a site admin to toggle analysis features on / off by setting the ANALYSIS_ENABLED config parameter to a boolean. 

ANALYSIS_ENABLED=true -> analysis features are shown. 
ANALYSIS_ENABLED=anything_other_than_true -> analysis features are disabled

* dashboard 'my / public analyses' tab is toggleable
* admin 'all jobs', 'app definitions', and 'job stats' tabs + panels toggleable
* datasets page -> job panel is toggleable

disables the 'run analysis' button for datasets page, regardless of ANALYSIS_ENABLED config, until it is refactored to work with scitran.

fixes bug where boolean params were not being set to the right truthy value

### DIFF
--- a/config.env.example
+++ b/config.env.example
@@ -63,4 +63,4 @@ ENVIRONMENT=
 
 # client -----------------------------------------------------------------------
 
-ENABLE_ANALYSIS=true
+ANALYSIS_ENABLED=true
--- a/config.env.example
+++ b/config.env.example
@@ -60,3 +60,7 @@ DOI_URL=
 
 SENTRY_DSN=
 ENVIRONMENT=
+
+# client -----------------------------------------------------------------------
+
+ENABLE_ANALYSIS=true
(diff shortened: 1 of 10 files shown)


nellh:
This looks pretty good but before we merge it, I think we should try it with the analysis specific configuration flags set to defaults and check if that works. There may be some hidden gotchas where the server doesn't start without AWS Batch services for example.

olgn:
@nellh i poked around with the aws batch vars set to defaults - everything is starting up just fine. the places in the app where aws config variables would be required have been disabled with the analysis flag, so those are unaffected as well.

PR #710, opened 2018-06-20
Handle client disconnects during uploads.

### LINKED ISSUE #709: Handle FileStreamDisconnectUploadError when upload is interrupted
Sometimes an interrupted upload generates this error and when unhandled it can lead to nginx deciding the backend is down.

nellh:
This catches both exceptions that can occur when a client disconnects while uploading, without catching any other errors to prevent silent failures. A warning is logged when a client disconnects.

Fixes #709


olgn (approved):
works great. server no longer breaks on cancelled uploads.

PR #711, opened 2018-06-21
More info section on DataLad for the front page

mih:
@chrisfilo I hope this is approximately correct, as I wasn't able to actually run the app, but pieced this together in the debugger in chrome.

Thanks!


nellh:
Thanks! I fixed the JSX syntax and this looks good :+1:

mih:
thx!

PR #713, opened 2018-06-21
set apollo fetch policy to no-cache on datasets page

olgn:
fixes a bug where the datasets page was fetching from an expired cache, causing partial (incomplete) flags to be out of date


nellh (approved):
Too bad we have to disable the client cache here but this looks like the best solution for launch.

PR #733, opened 2018-07-05
External site content files

olgn:
initial work to make the site content configurable for differences between nimh / crn openneuro branches.

fixes #595 
fixes #691 
fixes #720 

* faq page content is written in markdown and stored in a configurable file - packages/openneuro-app/src/faq/faq-config.js. 
* home page content is configurable, the various settings are stored in the packages/openneuro-app/src/front-page/front-page-config.js file. 
* privacy policy link added to the front page

PR #734, opened 2018-07-06
Configurable site theming

### LINKED ISSUE #594: Configurable site theming
The color palette, logo, and other brand images should be configurable.

olgn:
makes the various css theme values configurable, and puts them in well-named .scss files. 

* _variables.scss has been split into three components:
-_colors.scss - contains all of the colors for the site
-_theme.scss - contains all variables that effect the site's theming (besides colors)
-_bootstrap.override.scss - containes all variables which server to override bootstrap theming

* all references to colors that are configurable have been updated to utilize variables set in above mentioned .scss files
* all references to fonts that are configurable have been updated to utilize variables set in _theme.scss

fixes #594


nellh (approved):
Similar to #733 this needs a way to load an alternate set of variables but otherwise great.

Issue #740, opened 2018-07-10
users with google providers have emails as ids


olgn:
pending #739, google users should have their email as an id, to allow for successful migration of scitran users into auth 2.0 system

nellh:
This is no longer relevant since #739 is implemented.

Issue #739, opened 2018-07-10
Implementing generic user IDs


ckrountree:

nellh:
This just adds a generic ID string to users so we can reference them independent of login provider.

olgn:
additionally handles the edge case where we do not have a google id stored for a user, and need to use an email for lookup

Issue #738, opened 2018-07-10
Migrate SciTran users to internal authentication system


ckrountree:

nellh:
This is done and in production.

PR #742, opened 2018-07-10
MongoDB migrations for DataLad + new auth system

nellh:
This adds the missing migrations for #738, #739, and #740 

Migrations do not run on server startup in this implementation because I think there's an open question about how to decide if a database with no migration history requires updates or not. For now, they are manually run with `node upgrade.js`.


olgn (approved):
this looks splendid!

* confirmed that users are migrated with new provider and provider id fields
* permissions models are shifted to new format and matched on user ids
* associated comment, datasets, jobs, stars, and subscriptions user fields are updated with new generic userid
* logging in with a google account replaces existing email-based providerId with the google id

PR #760, opened 2018-07-13
Graphql analytics

### LINKED ISSUE #750: fix analytics sans scitran
the analytics used to be stored in scitran, and are now broken since the removal of scitran in lieu of datalad.

olgn:
fixes #750 

* moves the analytics counters (views and downloads) from scitran into graphql.
* includes a script for the analytics migration from scitran to gql.
* updates the corresponding front-end components to use the new methods


nellh (commented):
This works as a direct replacement but I am wondering why we store an entry for each event and sum those instead of just incrementing the value? It doesn't look like we're storing any extra metadata that makes keeping the individual entries around useful.

nellh (approved):

PR #763, opened 2018-07-19
Runtime configuration support

nellh:
This separates the environment variable configuration from the app container, allowing configuration of the app without rebuilding. To do this, the nginx and app containers are merged, since the app will be static at runtime except in the development configuration. The old app container is still present in the dev configuration to run webpack-dev-server and it uses the same Dockerfile but skips the nginx stage.

CSS variables can also be injected via the same configuration path by setting values for config.theme.variables

```javascript
const config = {
  theme: {'--primary': 'palevioletred'}
}
```

Will override the primary color, for example.

Need to finish a few things here still:
- [x] CircleCI 2.0 update is needed to get this working on CI again.
- [x] Provide a way to merge configuration from outside of the Git tree, just a JSON file that can override any values specified in the defaults or by environment variables.
- [x] Port over content hooks from #733


olgn (approved):
a journey, indeed.

* circle ci config updated to 2.0 for openneuro
* external content containers implemented
* configurable frontPage, faq, and site themes

PR #774, opened 2018-07-23
Restore blacklist functionality and ignore git and DataLad metadata.

### LINKED ISSUE #731: Datasets containing a git repo uploaded via cli/web lead to corrupt git repos
Uploads are not protected from writing to .git and .datalad directories, which allows the repo to become corrupted by overwriting files in those directories.

nellh:
Fixes #731 and some related issues when users attempt to upload git or datalad metadata as part of their dataset. Later we'll support some kind of push mechanism but for now we filter out these files.

### DIFF
--- a/packages/openneuro-server/datalad/dataset.js
+++ b/packages/openneuro-server/datalad/dataset.js
@@ -139,6 +139,9 @@ const fileUrl = (datasetId, path, filename) => {
   return url
 }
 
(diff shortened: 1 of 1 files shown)


olgn (approved):
looks splendid 

* any file/folder with a name that includes text from the blacklist is not uploaded.

PR #812, opened 2018-07-31
Re-enable production source maps by upgrading to Webpack 4

nellh:
We disabled source maps in production because of a babel-minify-webpack-plugin bug. Upgrading to webpack 4 we no longer need this plugin and source maps can be re-enabled to make debugging production issues easier.


nellh:
Started to enable hot module reloading too but there is some integration needed to get loadable-modules and webpack-hot-reload working together. Much closer to working than it was before v4.

Issue #663, opened 2018-05-30
Allow resuming uploads with CLI and new uploader


nellh:
These share some core code that needs to be updated to resume instead of overwriting datasets.

- [x] CLI support - see #824
- [x] Browser support

chrisgorgo:
Partial implementation at https://github.com/nellh/openneuro/tree/cli-resume-support

chrisgorgo:
> The datalad-service only returns files in the git index, so I was testing this by making commits, but to work for interrupted uploads on OpenNeuro.org it would need to get the list of files in the working tree including those not yet committed.

> I guess we could make commits after each file, but this could slow things down

> Yeah, and I think the working tree is a fairly good way to track if your bulk edit (uploads being a kind of bulk edit) was interrupted. We have a flag on the datasets, partial, which indicates if it is in that state where there are some files uncommitted.
> The first version of the datalad-service you tried out on dev committed after every file and was pretty slow

nellh:
In both uploaders, the file list just needs to filter out any selected files that are already complete before it's passed into the Apollo client. The old uploader did this by comparing the file lengths based on path and including any missing or truncated files.

nellh:
- [x] CLI support - see #824
- [ ] Browser support

PR #824, opened 2018-08-03
CLI resume upload support

nellh:
Fixes the CLI half of #663

Updated documentation will be published with the npm package for the 2.1.0 release.

I removed the 'sync' alias because this seemed a bit confusing. Uploads now always try to resume without reuploading files if you've specified an accession number.

### DIFF
--- a/packages/openneuro-cli/README.md
+++ b/packages/openneuro-cli/README.md
@@ -28,6 +28,6 @@ Your dataset must pass validation to upload but warnings can be skipped with `op
 
 To resume an interrupted upload or add files to an existing dataset:
 
-`openneuro upload --resume --dataset <accession number> <dataset directory>`
+`openneuro upload --dataset <accession number> <dataset directory>`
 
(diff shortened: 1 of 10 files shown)


olgn (changes requested):
this looks great for the most part - a couple small changes:

* snapshots are not created when the upload resume is completed, but they should be.
* attempting to use the --dataset flag on a dataset that does not exist should have a more useful error message. currently it is the gql 'cannot read property draft of null' which might not be super intuitive to users that are not familiar with the code base
* i think we could emit a 'draftFilesUpdated' subscription notification that would update the browser when a new dataset is added via cli (not necessary to this pr, but easily implemented at this point)

nellh:
I've fixed the snapshot bug.

You should get 'You do not have access to read this dataset' for datasets which do not exist, which is a bit misleading but necessary to prevent leaking information about private datasets. It isn't displayed very nicely but fixing that is out of scope for this.

PR #827, opened 2018-08-06
811 missing summary

### LINKED ISSUE #811: Summary information is missing
We need to calculate the snapshot size based on the expected content. There's a size value as part of the annex key and non-annexed git objects can just be checked with stat, these just need to be summed to show an accurate size.

olgn:
fixes #811 

* adds summary field to the openneuro-client getDatasets query
* updates the bids.js formatDataset constructor to use the validation summary provided by the new dataset backend

### DIFF
--- a/packages/openneuro-app/src/scripts/utils/bids.js
+++ b/packages/openneuro-app/src/scripts/utils/bids.js
@@ -46,6 +46,7 @@ export default {
     const followers = (await crn.getSubscriptions()).body
     let resultDict = {}
     for (let project of projects) {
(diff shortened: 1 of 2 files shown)


packages/openneuro-app/src/scripts/utils/bids.js
```diff
@@ -159,7 +159,7 @@ export default {
             project.snapshot_version = snapshot ? snapshot.tag : null
             project.analytics = snapshot
               ? snapshot.analytics
-              : project.analyticsa
+              : project.analytics
```
olgn: d'oh

PR #828, opened 2018-08-06
Web resume upload

nellh:
Depends on #824 

* Re-enables resume upload button.
* Only uploads files with different sizes than the remote copy.
* Refactors the uploader to decouple state from UI. This lets us move the uploader up the component tree and subscribe to state in different places (like the resume buttons).
* Disables the resume upload buttons with a tooltip while an upload is happening.


olgn:
'cancel upload' button no longer works as a regression w/ new uploader component

PR #842, opened 2018-08-16
Fix files resolver returning untracked files by default

### LINKED ISSUE #840: "You are not authorized to view this dataset"
When accessing https://openneuro.org/datasets/ds001454 by the owner or an admin.

Context: https://openneuro.freshdesk.com/a/tickets/107

nellh:
A query with `draft { files {} }` would pass in an empty object instead of undefined or null.

This means you would get untracked files with no id if the argument was unspecified.

Fixes #840


olgn (approved):
* only a `null` value of options would allow an `undefined` value for the `untracked` variable, which is just perfect... because `options==null` -> bad things are happening in gql

Issue #816, opened 2018-08-02
Minting a DOI on a snaphot fails with "Can not mint DOI for draft"


chrisgorgo:
https://openneuro.org/datasets/ds001302/versions/1.0.0
![image](https://user-images.githubusercontent.com/238759/43595848-8db1c02a-9632-11e8-9562-b13b1cc5e242.png)

Happens on this dataset as well https://openneuro.org/datasets/ds001421/versions/00002

chrisgorgo:
Another user run into this. @rwblair what's the ETA for the fix?

rwblair:
I'll have a PR ready for review today.

ckrountree:
Fixed by the DOI minting refactor.

PR #853, opened 2018-08-24
Fix doi minting

rwblair:
Fix for #816 Attempts to automatically mint doi before snapshot creation, continues normally on failure, otherwise updates description accordingly. Also removes manual minting button.


rwblair:
Annnd I forgot to lint it.

Issue #863, opened 2018-08-31
Updating a dataset via CLI creates a new snapshot automatically


chrisgorgo:
https://openneuro.org/datasets/ds001205

This mechanism does allow users to add a meaningful commit message. We should either ask for a commit message in the uploader or instruct users to create a snapshot via the web.

nellh:
I think we should change the automatic first snapshot creation to be triggered only when there are no snapshots and the first data commit is made. That would allow it to happen on initial upload or resume of the initial upload, but not for any other edits.

chrisgorgo:
Sounds like a plan!

PR #870, opened 2018-09-02
Don't create snapshots after uploading to an existing dataset via CLI

nellh:
See #863 - this has the consequence of never creating the first snapshot for a resumed upload but that's better than creating bogus snapshots on every CLI update.


olgn (approved):
works well. eventually we can have a cli snapshot feature ;)

PR #872, opened 2018-09-04
Faster snapshot creation

nellh:
This reuses the files cache from the commit when a tag is created, avoiding an extra walk of the full git-annex tree during snapshot creation.

A fallback is included in cases where the cache is not available. Those cases don't benefit at all here, but they are rare.

Depends on https://github.com/OpenNeuroOrg/datalad-service/pull/58


olgn (approved):
smooth. Almost every create snapshot utilizes the cache vs checking out snapshot tag again (which wastes valuable datalad-service resources)

PR #873, opened 2018-09-04
Avoid duplicate get files request to datalad-service

nellh:
A few clients can stack up slow requests for files while waiting on another request to finish. This limits the running requests to one at a time based on the URL and query options provided. The same promise is returned to each call.


nellh:
Closing this as too complex for the benefit. There's other ways to deduplicate these requests without getting into promise generators.

PR #878, opened 2018-09-06
Fix snapshot files cache miss

nellh:
Fixes a regression where there's a cache miss when the draft is accessed during initial snapshot publish to a remote.


olgn (approved):
indeed!

PR #881, opened 2018-09-06
Add publish celery worker and create implicit snapshots server side

nellh:
This adds the publish worker to docker-compose.yml for https://github.com/OpenNeuroOrg/datalad-service/pull/60

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,6 +72,18 @@ services:
     env_file: ./config.env
     init: true
 
+  # publish-only celery worker
+  publish:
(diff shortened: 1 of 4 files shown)


nellh:
I've added one more commit here to move the first snapshot creation over to commitFiles. If a snapshot does not exist, one is created before the commit returns now. The should ensure it is always created on uploads regardless of what happens on the client, resolving a number of bugs where the required first snapshot is not created.

olgn (approved):
concurrent publishes?! yes please

PR #892, opened 2018-09-10
Clear snapshot index only after snapshot is created

### LINKED ISSUE #888: snapshot order and dates
https://openneuro.org/datasets/ds000114
image

after clicking on 1.0.1
image

after clicking back on draft
image

nellh:
This moved the snapshot index delete call to the end of snapshot creation, once metadata is available. This prevents a stale cache from being created while the snapshot request has been made but the snapshot is not yet complete.

Fixes #888 - particularly with DOI snapshots which were almost always causing this order of operations.


olgn (approved):
this feels better

nellh:
Closing this as obsoleted by #897

PR #893, opened 2018-09-10
ORCID error messages

### LINKED ISSUE #770: ORCID login message for no email / no name does not redirect
This is mainly a problem with ORCID as the other login methods will show you an error on the providers site and only callback in successful logins. The ORCID version should redirect you back to the React App and display the error, so you can try again without manually navigating.

nellh:
This gives useful error pages during failed ORCID authentication. This is the top error in Sentry right now.

screenshot from 2018-09-10 16-30-59

You can either set a "credit name" or both a given name and family name, so the errors messages ask you to just make sure you have a public name of some kind. Hopefully that's enough guidance for users.

Fixes #770

### DIFF
--- a/packages/openneuro-server/libs/orcid.js
+++ b/packages/openneuro-server/libs/orcid.js
@@ -43,7 +43,7 @@ export default {
           if (!name) {
             if (!firstname) {
               reject({
-                url: 'given',
+                type: 'given',
                 message:
                   'Your ORCID account does not have a given name, or it is not public. Please fix your account before continuing.',
               })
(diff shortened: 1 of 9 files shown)


packages/openneuro-server/libs/orcid.js
```diff
@@ -41,22 +42,28 @@ export default {
 
           if (!name) {
             if (!firstname) {
-              reject(
-                'Your ORCID account does not have a given name, or it is not public. Please fix your account before continuing.',
-              )
+              reject({
+                url: 'given',
```
olgn: should this be `type`?

nellh: Yep, thanks!

olgn (approved):

PR #897, opened 2018-09-11
Don't block on Snapshot.files from createSnapshot mutations

nellh:
This makes the files field on createSnapshot async by returning a promise to the resolver, saving the long wait if files are not immediately available from cache and the client hasn't requested this field.


nellh:
Added in #892 changes to avoid a merge conflict between these two.

olgn (approved):

Issue #911, opened 2018-09-17
Refactor datasets page to use Apollo natively


nellh:
This is a tracking issue for the datasets portion of #905 - partially complete.

nellh:
This is complete!

Issue #693, opened 2018-06-12
Dashboard uses the wrong label for datasets


nellh:
It reads the obsolete label field when it should be the value from dataset_description.json

nellh:
This makes sense to fix with the dataset part of #905.

PR #917, opened 2018-09-21
Add draft { description } and snapshot { description } API fields

### LINKED ISSUE #910: Provide API fields for dataset_description.json content
Add mutations and queries to support the dataset and dashboard refactors in #905.

#693 is related here since the dashboard queries this info different to avoid loading hundreds of files client side just to get dataset titles. Doing this on the server is a requirement to consolidate these two.

nellh:
This fixes #910 and #693 , and provides the content of datasets_description.json as server side fields.

A few corner cases are handled better by this API, if datasets description is entirely missing or non-JSON data, a default "Unnamed Dataset" and current BIDSVersion fields are generated. This lets other edits which alter the dataset_description.json file include the minimum needed to validate. "Unnamed Dataset" is better than an empty string in this rare case.

Additionally this PR refactors many of the resolvers to be more composable and cleans up the root tree to avoid as many explicit named imports. 

The app here is still writing the entire file and in some cases reading it directly as well. I made the minimum changes to consolidate the label field with the Name in dataset_description. I want to add the mutation(s) for this with the new components in #911 and remove any direct reads in the same refactor.


olgn (changes requested):
* changing the dataset name on a draft page causes the name to be changed on the snapshots, too... that seems wrong.
* the uploader asks for a dataset name - and then that name never gets used (because the name is inherited from the dataset description file). If we leave a name entry here, it should update the dataset description file after upload. alternatively we could just take that field out of the uploader form.

nellh:
@olgn Thanks, those are both addressed now.

For the rename during upload, I went with the version that edits the dataset_description.json file during upload. Included is a couple tests to validate the new function added to the UploadClient component for this. If you do edit the name during the initial upload, you can overwrite the name accidentally by resuming later, but resume should work. This also ensures the dataset_description.json file is always formatted.

nellh:
@chrisfilo FYI, this also changes how datasets are named during upload.

* The Name field in dataset_description.json is suggested during upload (instead of directory name).
* If Name can't be read, the directory name is suggested instead. This will also throw validation in the next step (you can't upload with a bad dataset_description.json in your local copy, even if the only problem is no name).
* If you edit the name during upload, the change is reflected in the server copy of dataset_description.json.
* dataset_description.json is always pretty printed on initial upload now, making it more consistently human readable / diffable.

chrisgorgo:
Thanks for the update. What do you mean by "The Name field in dataset_description.json is suggested during upload"? There is an extra dialog box?

nellh:
> Thanks for the update. What do you mean by "The Name field in dataset_description.json is suggested during upload"? There is an extra dialog box?

For `Step 2: Name Dataset`, it was prefilled with the directory name instead of the dataset_description.json Name value.

olgn (approved):

olgn:
looks great!

PR #924, opened 2018-09-26
Generate changes on upload

### LINKED ISSUE #891: 'undefined' is sometimes added to CHANGES file
There's a case where the generated `CHANGES` file contains `undefined` at the end.

olgn:
fixes #891 

* cli + web uploader - if CHANGES file does not exist, adds it with the text
```
1.0.0    YYYY-MM-DD
    - Initial snapshot
```
* fixes a bug where (when CHANGES file not present on draft) the word 'undefined' is added to the bottom of the CHANGES file


nellh (approved):
Working with some regular test datasets and a couple torture tests from both CLI and web. An existing dataset missing CHANGES does not get undefined added after snapshot creation. :+1:

PR #925, opened 2018-09-26
Avoid holding onto invalid docker-compose dns in nginx

### LINKED ISSUE #809: Nginx backend never recovers after server restart
From #808 - the server restarting should recover on its own but it did not, nginx never reconnected.

nellh:
Fixes #809 

The original fix here did work for the reproduction case tested but there's a secondary issue which looks almost same but only happens when the container exits unexpectedly (not just when the backend connection drops). The reproduction scenario is:

1. All containers start normally. Nginx resolves the DNS value when it sets up the upstream directive.
1. Server container exits and nginx marks the upstream down.
1. Server container is restarted by docker-compose. This gives it a new virtual network device - sometimes changing the IP address and DNS is updated.
1. If the IP changes, nginx never notices the backend is at the new address because the DNS resolved and this is cached internally by nginx as an already resolved server address (it helpfully round robins our one dead IP address).

The resolver cache can be forced to update at the TTL interval by [specifying the server name in a variable](https://www.nginx.com/blog/dns-service-discovery-nginx-plus/#domain-name-proxy_pass) which forces it to be rechecked in case the variable has been altered by another rule in the request. The docker-compose TTL is 600 seconds but nginx allows this to be overridden so I've lowered it to ten seconds which gives some pretty reliable recovery after a container IP change.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -1,6 +1,5 @@
-upstream nodejs {
-    server server:8111 fail_timeout=10s max_fails=0;
-}
+# Override docker-compose 600s TTL
+resolver 127.0.0.11 valid=10s;
 
 server {
     listen 80;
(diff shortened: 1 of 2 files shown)


olgn (approved):
nice!

PR #927, opened 2018-09-27
getDraftFiles - If no revision is specified, return untracked files.

### LINKED ISSUE #926: More performance woes
I'm getting 504 on graphql endpoint when trying to load https://openneuro.org/datasets/ds001522. The user tried to upload this dataset four times - most likely because they could not access the resume functionality.

Loading the page a second time worked (I guess cache kicked in).

nellh:
No revision means there's been no commits made yet, such as during initial upload. This avoids writing the incomplete files to cache and includes untracked files (since the draft only consists of untracked files in this case).

Fixes #926

### DIFF
--- a/packages/openneuro-server/datalad/draft.js
+++ b/packages/openneuro-server/datalad/draft.js
@@ -18,7 +18,7 @@ const uri = config.datalad.uri
  */
 export const getDraftFiles = async (datasetId, revision, options = {}) => {
   // If untracked is set and true
-  const untracked = 'untracked' in options && options.untracked
+  const untracked = ('untracked' in options && options.untracked) || !revision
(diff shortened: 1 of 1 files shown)


chrisgorgo:
Is the assumption that no commits means ongoing/interrupted upload always
true? I've seen (and fixed) datasets that finished uploading without
triggering a commit. Could be due to another bug, but maybe a more robust
check is necessary.

nellh:
@chrisfilo Do you mean snapshots? One commit is created for every edit and the first is always created when the initial upload completes, or is resumed and finishes.

chrisgorgo:
Ah yes - I confused commits and snapshots.

olgn (approved):

PR #942, opened 2018-10-03
Email notifications fix

olgn:
fixes broken email notifications that still relied on scitran

* scitran user lookups are replaced with Mongoose User model queries
* snapshot notification email is now triggered via createSnapshot workflow


nellh:
:+1:

Issue #551, opened 2018-04-09
Add download support via the CLI tool


nellh:
From #549, add support for downloads without requiring a browser.

Following #200

chrisgorgo:
Most people will be interested in downloading latest snapshot of a public dataset - this should be straightforward with AWS CLI. Downloading historical snapshots of public datasets should be possible with datalad.

That leaves us only with CLI downloads of private datasets which might not be a high priority.

PR #950, opened 2018-10-13
CLI Download Support

nellh:
I needed to script this up to debug download issues, so implemented #551 in the process

It can download any draft or snapshot you have access to - using the same technique as browser downloads. It currently does not try the fallback URLs if the first URL fails but is otherwise identical.

If you do not specify a snapshot, it gets the list of snapshots and prompts you to choose one. If you have existing files locally, it will skip downloading files you already have.

### DIFF
--- a/packages/openneuro-cli/README.md
+++ b/packages/openneuro-cli/README.md
@@ -22,6 +22,8 @@ You can manually configure custom servers by editing this file.
 
 # Usage
 
+## Uploading datasets
+
 To upload a new dataset:
 
 `openneuro upload <dataset directory>`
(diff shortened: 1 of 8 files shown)


chrisgorgo:
This is very cool - I'm sure it will come handy for many users.

olgn (approved):
looks good!
* downloads dataset to specified location
* creates destination folder if it does not exist
* throws error when there is no source dataset
* throws error if the user does not have permission to access dataset

One thought - perhaps there should be a separate error for 'the dataset does not exist' versus 'you do not have access to this dataset'

PR #951, opened 2018-10-13
Fix bug with resumed uploads losing dataset names

nellh:
@chrisfilo I reproduced the missing name after resume upload issue - this is a fix for that.


olgn (commented):
the dataset will be called 'Unnamed Dataset' on an incomplete upload. after resume upload has completed the dataset has the proper name. @nellh is this the desired behavior?

nellh:
Yes, the bug this is fixing was leading to no dataset_description.json being uploaded after resuming, if it gets "Unnamed Dataset" or the original one the fix is correct.

olgn:
sweet!

PR #954, opened 2018-10-14
Fix rendering for Googlebot

### LINKED ISSUE #835: Fetch as Google renders blank page
This likely has some impact on indexing. Chrome 41 is currently working, so this is a new kind of Googlebot issue.

nellh:
Fixes #835 

This turned out to be due to ES6 arrow functions in bids-validator not being transpiled to a syntax supported by Chrome 41. The production build obscures the error here so this took a while to track down!

This used to not be an issue since the validator only contained ES5 syntax. To fix this, I've updated everything to use Babel 7 (the new config format supports transpiling node_modules better) and enabled Babel processing on the node_modules dependencies. This is kind of a good idea anyways because other dependencies aren't guaranteed to run on Chrome 41 / Googlebot, so this protects the React app build from that. It does increase build times significantly and impacts bundle size though, might require some optimization there.

I've also added some missing CSS fallbacks that Googlebot needs to make nicer previews.


nellh:
Found one more issue here which was Googlebot does not have websockets even though Chrome 41 does. This is now also fixed so all pages should render as expected and if Googlebot isn't finding something it's a page structure issue now.

chrisgorgo:
Nice! https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fopenneuro.dev.sqm.io%2Fdatasets%2Fds001023%2Fversions%2F1.0.0

olgn (approved):
LGTM

* googlebot indexing improved
* css fallbacks for variable names

PR #961, opened 2018-10-17
Moving LICENSE to the root of the repository

chrisgorgo:
Consistent with `datalad-service` repo. Pending approval from Eric.


nellh:
We should remove the LICENSE files from the sub-packages, having one up to date at the top level is best.

chrisgorgo:
I see! So no need to discuss the actual license - it's just moving it to the roo. I pushed a commit removing the subpackage licenses. This is RTM.

PR #962, opened 2018-10-17
Clarifying support information

chrisgorgo:


nellh:
The CI issue here is that the branch name isn't a valid docker tag and we build branch images for every source repo branch.

PR #964, opened 2018-10-18
Remove conflicting `cache-control: public` from index.html

nellh:
This should weed out caches that are just looking for `public` to decide if they can cache index.html longer than expected.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -79,7 +79,7 @@ server {
     # crn-web app
     root /srv/app/dist;
     location /index.html {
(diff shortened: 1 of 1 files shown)


olgn (approved):

nellh:
> Let's deploy this and see if it makes a difference. Still plenty of clients stuck with the old version: https://sentry.io/openneuro/openneuro/releases/

This was deployed yesterday but we can't control how long caches will take to notice the update, I'd expect most to evict the bad header version in 24-48 hours but we may see this longer in Sentry for some.

Issue #935, opened 2018-09-28
Some snapshots are missing S3 indexes


nellh:
Looks like a few snapshots did not index the S3 files once exported. Short term fix: we should rerun the indexing for the affected snapshots.

Better fix will be to get the S3 object versions from git-annex following https://github.com/OpenNeuroOrg/datalad-service/issues/61

chrisgorgo:
I wonder if it would be time-consuming to add a "sentry assertion" to monitor such situation. When a snapshot is requested, check if it's a) older than 2 days b) not on S3 - if so send an error report to Sentry, but keep serving local files.

nellh:
We'll fix this by enabling https://github.com/OpenNeuroOrg/datalad-service/issues/61 on all datasets. That way we can simplify this implementation pretty significantly.

There's three classes of snapshots to update:

- [x] Exported and indexed snapshots - these will be updated from the existing metadata and this should be transparent to users.
- [ ] Exported but not indexed snapshots - these will be re-exported. This will temporarily make these unavailable on S3.
- [x] Not exported at all - these just require the remote updates and the export can be kicked off.

After all are updated we can update the GitHub exports.

Monitoring with Sentry is a good idea. We have a status endpoint for datasets that could be extended to include a check for this.

PR #968, opened 2018-10-19
implement s3 versioning migration

olgn:
partial implementation of fix for #935. Run this script with node from within the openneuro-server Docker container to generate $PERSISTENT_DIR/crn_server/persistent/fileVersions.json. Then use [this script here](https://github.com/OpenNeuroOrg/datalad-service/pull/71) to update the dataset git-annex metadata and allow for s3-remote versioning.

* loops though all datasets
* loops through all tags of this dataset
* grabs the files of each tag
* obtains the file version info for each tag, as well as some tag metadata(eg. which bucket it is stored in)
* outputs fileVersions.json into the /persistent/ directory ($PERSISTENT_DIR/crn_server/persistent/fileVersions.json)

### DIFF
--- a/packages/openneuro-server/migrations/s3-versioning.js
+++ b/packages/openneuro-server/migrations/s3-versioning.js
@@ -1,0 +1,76 @@
+/* eslint-disable no-console */
+import config from '../config.js'
+import mongo from '../libs/mongo.js'
+import mongoose from 'mongoose'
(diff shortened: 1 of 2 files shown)


nellh (approved):
I think this looks good, going ahead to merge it and we can test things a bit more on dev before deployment.

PR #976, opened 2018-10-24
Resume upload in upload error toast

nellh:
This adds a resume upload button to the interrupted upload error. This takes you back into the upload workflow for the dataset upload which triggered the error. Toasts are now styled to match site styles better as well.

image


olgn (approved):
works nicely

PR #993, opened 2018-11-06
RF: Remove "Advanced Methods" decoration around cmdline and datalad download types

yarikoptic:
- Generally that section is collapsed and makes those methods not visible
  to a regular user, so unlikely they would be interested to go "advanced route"
  and thus would not please greedy me to try to DataLad

- "Your advanced is my basic" or "Your basic is my crippled"
  aws cmdline is more advanced to me than DataLad invocation.
  All options besides "Download" button could be considered "Advanced",
  and "Download" button could be considered a "Crippled" way

So why discriminate? ;-)

### DIFF
--- a/packages/openneuro-app/src/scripts/datalad/download/download-dataset.jsx
+++ b/packages/openneuro-app/src/scripts/datalad/download/download-dataset.jsx
@@ -35,19 +35,17 @@ const DownloadDataset = ({
         <DownloadS3 datasetId={datasetId} />
       </PaddedDiv>
     </div>
-    <ExtraPaddedDiv className="col-xs-12">
-      <Panel header="Advanced Methods" collapsible>
-        <PaddedDiv className="col-xs-6">
-          <DownloadCommandLine
-            datasetId={datasetId}
-            snapshotTag={snapshotTag}
-          />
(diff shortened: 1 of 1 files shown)


chrisgorgo:
This will make the layout very busy. I thought we alredy provided
sufficient promotion for the datalad project...

yarikoptic:
thank you for promoting!  Even though I added my "greedy" comment, the main drive for the change was not entirely personal greed/promotion but rather giving openneuro users a better presentation of what they could use.

re business: It will be a layout with 4 boxes instead of 3 (if Advanced ...  is  collapsed) or 5 (2 are embedded inside one) when it is not collapsed. Altogether to me it perceives busier than a simple uniform listing.
Moreover, "Advanced .." has no indicator of any kind (down error) which hints that clicking on it would expand it.  For some users it might be perceived as just an empty stub for something you planned to be added in the future.

olgn:
@yarikoptic I agree that the advanced tab is not obviously clickable - I think adding a down/up carat and/or a (Show/Hide) Advanced Methods would probably do it justice

yarikoptic:
I agree to agree! ;-)

PR #998, opened 2018-11-08
Log download button click to Google analytics

nellh:
Label is set to the URI for the draft or snapshot. This should already be tracked by the page tracking but just to make sure it's logged with the download event as well.

Tested this with Google analytics blocked/unblocked to make sure it doesn't interrupt downloads for anyone who blocks requests to GA.


nellh:
Made one change, the label is accession_number for drafts and accession_number:snapshot_tag for snapshots. That seems like it will be easier to categorize in GA, especially if the download URL changes at some point.

chrisgorgo:
This is perfect - thank you.

chrisgorgo (approved):

PR #1000, opened 2018-11-09
Use new 'key' field for obtaining git object hash or git-annex key

nellh:
This is required to work with https://github.com/OpenNeuroOrg/datalad-service/pull/76


olgn (approved):
works great. 

* tested uploading duplicate files (with different names) into the same folder 
* tested uploading duplicate files into different folders
* tested viewing / downloading individual files

Issue #905, opened 2018-09-14
Remove Reflux stores


nellh:
Some major components on the React frontend are still using [Reflux](https://github.com/reflux/refluxjs) which has been largely abandoned. As we've seen from using it heavily, the multiple stores and many dispatchers lead to confusing and overly complex stores and components. A lot of bugs could be resolved by simplifying the components to no longer use Reflux. This is a requirement to fully take advantage of client side caching and subscriptions within Apollo as well, Apollo's cache becomes a new single store organized by the GraphQL schema.

There's a few areas of updates needed to make this happen. In descending order of complexity:

- [x] Datasets
- [x] Dashboards
- [x] Admin
- [x] Front Page
- [x] Notifications
- [x] JSON Editor

The first two are most of the work here, the others are relatively simple and can easily be refactored to not use a store at all. The datasets refactor is partially done, with a very minimal implementation [here](https://github.com/OpenNeuroOrg/openneuro/blob/v2.2.2/packages/openneuro-app/src/scripts/datalad/datalad.dataset.jsx).

PR #1001, opened 2018-11-09
Reduce subscription spam

### LINKED ISSUE #919: API subscriptions are spammy
API subscriptions publish irrelevant data to many clients which leads to extra data reload by unconcerned clients. It looks like this is caused by a lack of filtering on the keys specified in the subscription schemas. This is a minor issue at the moment but it will cause scaling issues with increases in traffic since it makes the cost of each update apply to every active subscriber.

nellh:
This narrows subscriptions to datasets users have actively open, removes one extra subscription to any dataset changes, and avoids an update query entirely for dataset deletion (though this does not update in the Reflux side of things).

There's a lot more fixes here but they depend on #905 - it's not possible to do optimistic updates without first removing the Reflux stores, so this is about all we can do to reduce the load generated until that's complete.

Fixes #919 and depends on changes in #1000 - so that will need to be merged first.

### DIFF
--- a/packages/openneuro-app/src/scripts/dashboard/dashboard.datasets.jsx
+++ b/packages/openneuro-app/src/scripts/dashboard/dashboard.datasets.jsx
@@ -62,7 +62,9 @@ class Datasets extends Reflux.Component {
     this.props.datasetsQuery.subscribeToMore({
       document: gql`
         subscription {
-          draftFilesUpdated
+          datasetCreated {
+            id
+          }
         }
       `,
       updateQuery: () => {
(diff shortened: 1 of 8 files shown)


nellh:
Updated the base so the list of commits here is clearer.

olgn (approved):
following subscriptions tested:
* draft files update works + filters on id
* snapshot added works + filters on id
* snapshot deleted never worked (apparently) and still does not work
* permissions update works + filters on id
* validation updated works + filters on id
* dataset added / deleted intentionally broken to reduce subscription spam
** these will work properly when dashboard refactor is complete and tied to apollo query components

Issue #1024, opened 2018-12-28
No indication that a dataset is public on its landing page


chrisgorgo:
https://openneuro.org/datasets/ds001618
![image](https://user-images.githubusercontent.com/238759/50498733-365b4000-09f8-11e9-9625-248fed6b2166.png)

There used to be a green icon - possibly a regression? Related to #530

PR #1030, opened 2019-01-08
Display draft published status and remove notifications Reflux store

### LINKED ISSUE #530: Make dataset accessibility status more prominent

nellh:
The store change is included here because of conflicting styles from the store and bootstrap for alerts.

This fixes #530, #1024, and removes the Reflux notifications store for #905. The message for #530 is only visible on drafts.


packages/openneuro-app/src/scripts/common/forms/warn-button.jsx
```diff
@@ -115,11 +116,13 @@ class WarnButton extends React.Component {
...
           if (validation.check) {
-            notifications.createAlert({
-              type: validation.type,
-              message: validation.message,
-              timeout: validation.timeout,
-            })
+            toast.error(
+              <ToastContent
+                title={validation.type}
+                message={validation.message}
```
olgn: i think the property for the toast message is 'body'

packages/openneuro-app/src/scripts/common/forms/warn-button.jsx
```diff
@@ -152,7 +155,7 @@ class WarnButton extends React.Component {
       this.setState({ loading: true })
       action(e => {
         if (e && e.error) {
-          notifications.createAlert({ type: 'Error', message: e.error })
+          toast.error(<ToastContent title="Error" message={e.error} />)
```
olgn: property for toast content is body (not message)

olgn (approved):
looks great! toasty

PR #1036, opened 2019-01-16
Ensure only unique results are merged in fast scrolling dashboard loads

### LINKED ISSUE #1027: Dataset dashboard lists duplicates, skips datasets
https://openneuro.org/public/datasets
Lists the same datasets multiple times and does not list all public datasets. This happens even if user is not logged in.
screenshot_51

screenshot_50

nellh:
This is needed to prevent multiple scroll events from loading duplicate rows if they request overlapping ranges. Fixes #1027


olgn (approved):
works great! tested on openneuro.dev.sqm.io

Issue #1034, opened 2019-01-12
"Public dataset" alert is visible to non-admins non-owners


chrisgorgo:
![image](https://user-images.githubusercontent.com/238759/51074783-30e53300-1651-11e9-8c90-0ac945b19a4d.png)

This alert should be visible only to owners of the dataset.

nellh:
This is now fixed, it is visible to users with write access only.

PR #1037, opened 2019-01-16
Link to snapshots directly from public dashboard

nellh:
This changes the dashboard links to point directly at the most recent snapshot for a given dataset as long as one is available. This will avoid an extra redirect, fixing one of the causes of #1034, and provides real URLs to crawlers for the public dashboard.

I'm not sure if we should redirect to the snapshot or draft from "my dashboard" though. I've kept the current behavior of "My Dashboard" linking to the drafts but it would be more consistent for it to also link to the newest snapshot. @chrisfilo Thoughts?


olgn (approved):
looks good - my dashboard links are to the draft pages, and public dashboard links are to the most recent snapshot

PR #1040, opened 2019-01-25
Replace Raven with Sentry-SDK

nellh:
Sentry is moving away from the Raven library we are using and while it says it's maintained, not every feature is working anymore. This ports all Sentry usage to the new SDK.


olgn (approved):
works - server side errors now captured in sentry, and existing error handling remains identical throughout openneuro-app

PR #1066, opened 2019-03-19
Refactor datasets page to remove Reflux dependency

### LINKED ISSUE #900: Make comment section wider
The comment section is quite narrow right now making it hard to read.
image

It would be better to extend it to full width of the the right panel. This would limit the maximum height of the file tree potentially requiring adding a scroll bar.

### LINKED ISSUE #1041: Deleting a dataset results in a failed page load
Reported by a user, this happens for any deleted dataset.

nellh:
This is the main set of changes for #911 - below is the work in progress tasks remaining.

- [x] Readme editor
- [x] Description fields - cache update fragment
- [x] Snapshot creation - Changelog is lost
- [x] Dataset status icons (e.g. public icon)
- [x] Comments - Authoring
- [x] Share dataset - Access level selector
- [x] Share dataset - Access removal button
- [x] DOI minting button
- [x] CSS fixes for new markup

This excludes any work on the file listing, that will be in a follow up PR to keep backend changes to a minimum in this PR.

This fixes #1041 and fixes #900 as side effects of those features being rewritten.


nellh:
@franklin-feingold @chrisfilo A couple questions that have come up here:

The README in the specification is plain text (ASCII or UTF-8) but we assume it is markdown in OpenNeuro. Should we continue parsing it as markdown?

It's possible to manually edit the DOI field, should we show the minting button instead? Do you think users edit the field via the web very often? It looks like most cases were either minted or uploaded with the dataset.

franklin-feingold:
Regarding the README - we can continue parsing it as markdown. Is there a way to have the markdown parser, but if it fails it can render as a regular text. Is there a performance increase by changing the parser?

nellh:
> Regarding the README - we can continue parsing it as markdown. Is there a way to have the markdown parser, but if it fails it can render as a regular text. Is there a performance increase by changing the parser?

It's negligible for performance. Markdown parsing could mangle some text formatting but I doubt it's a big issue since we've been doing this since the first public release of OpenNeuro, I just wanted to double check before reimplementing it on the refactored datasets page :+1:

david-nishi (approved):
Aside from the comment reply persisting after submit, it all looks good to me!

franklin-feingold (approved):

PR #1109, opened 2019-04-30
Remove unused worker container from docker-compose

nellh:
This used to submit jobs for AWS batch but is no longer used.


david-nishi (commented):
should the openneuro-server/worker.js file also be removed?

nellh:
Thanks, removed that as well.

david-nishi (approved):

PR #1111, opened 2019-04-30
add prominent dataset download button link

david-nishi:
Screen Shot 2019-04-30 at 16 18 37


nellh (changes requested):
The old button should be removed and the style of the new button does not match the design spec.

david-nishi:
@nell I talked with Kate about removing the old button and whether the new button styles should be consistent with the others or follow the design spec. Might be good to hash this out together @kate

ckrountree:
@DaNish808 we can remove the download icon.  The style of the new button should mimic the old look in terms of radius - for consistency, I think (@nellh  - thoughts?).   However, the size should remain equivalent to the design spec.

david-nishi:
Screen Shot 2019-05-02 at 10 27 25
Toolbar icon removed and style transitioned to new spec.

nellh (approved):

Issue #1107, opened 2019-04-29
Add error message for CLI TypeError: path must be a string or Buffer


franklin-feingold:
reported by user - could not replicate this error locally
dataset: 1769

This dataset is also not present on the openneuro github repo so limits options to download. Locally I was able to download dataset via the Node.js client.

```$> node_modules/.bin/openneuro download --snapshot 1.1.0 ds001769 ds001769-download/

/home/usr/proj/openneuro/node_modules/openneuro-cli/src/config.js:1
TypeError: path must be a string or Buffer
at Object.fs.openSync (fs.js:646:18)
at Proxy.fs.readFileSync (fs.js:551:33)
at readConfig (/home/usr/proj/openneuro/node_modules/openneuro-cli/src/config.js:14:13)
at getUrl (/home/usr/proj/openneuro/node_modules/openneuro-cli/src/config.js:45:29)
at getDownloadMetadata (/home/usr/proj/openneuro/node_modules/openneuro-cli/src/download.js:41:22)
at getDownload (/home/usr/proj/openneuro/node_modules/openneuro-cli/src/download.js:56:3)
at Command.download (/home/usr/proj/openneuro/node_modules/openneuro-cli/src/actions.js:145:5)
at Proxy.listener (/home/usr/proj/openneuro/node_modules/commander/index.js:315:8)
at emitTwo (events.js:126:13)
at Proxy.emit (events.js:214:7)```

**Comment by danieltomasz** [comment]
I am getting the same error


```
➜ openneuro download --snapshot 1.0.3 ds000117 ds000117-download/
/home/daniel/.config/yarn/global/node_modules/openneuro-cli/src/config.js:1
TypeError: path must be a string or Buffer
    at Object.fs.openSync (fs.js:646:18)
    at Proxy.fs.readFileSync (fs.js:551:33)
    at readConfig (/home/daniel/.config/yarn/global/node_modules/openneuro-cli/src/config.js:14:13)
    at getUrl (/home/daniel/.config/yarn/global/node_modules/openneuro-cli/src/config.js:45:29)
    at getDownloadMetadata (/home/daniel/.config/yarn/global/node_modules/openneuro-cli/src/download.js:41:22)
    at getDownload (/home/daniel/.config/yarn/global/node_modules/openneuro-cli/src/download.js:56:3)
    at Command.download (/home/daniel/.config/yarn/global/node_modules/openneuro-cli/src/actions.js:145:5)
[...]

PR #1118, opened 2019-05-08
Added TypeError solution to README

franklin-feingold:
Addresses https://github.com/OpenNeuroOrg/openneuro/issues/1107

Changes: Added the solution into the README file


nellh:
@franklin-feingold Thanks but I think this is best in the openneuro-cli readme? Or both places? https://github.com/OpenNeuroOrg/openneuro/blob/master/packages/openneuro-cli/README.md

franklin-feingold:
I think you are right that this should be put into the CLI README. That is the place that is more useful and more appropriate. I will open the PR over there. Thank you!

Issue #1127, opened 2019-05-16
Add "Open in BrainLife.io" button to dataset page


effigies:
[BrainLife](https://brainlife.io) is making a number of OpenNeuro datasets available for their users at https://brainlife.io/project/. For those datasets which are available, we'd like to add a "Open in BrainLife.io" button next to the "Download" button.

I've mocked up a button for DS000003:

<img width="663" alt="Screen Shot 2019-05-16 at 13 45 36" src="https://user-images.githubusercontent.com/83442/57878638-74f76880-77e8-11e9-8db4-d906434f9448.png">

I did this by duplicating the download button, adding `margin-right: 1em` to the button CSS, and adding `display: flex` to the containing' `<div>`'s CSS.

The target URL for ds00ABCD would be https://brainlife.io/openneuro/ds00ABCD.

BrainLife is not automatically indexing all OpenNeuro datasets, so we'll need some way to periodically poll which are available.

cc @soichih

franklin-feingold:
Looks good!

@soichih - when you add openneuro datasets to brainlife (which looks great!) would it be possible to be notified or have a weekly/biweekly summary of additions so the updates can be better documented and tracked for everyone?

soichih:
@franklin-feingold Thanks! Yes, I can store and publish the information somewhere. Do you have any suggestion on format/location? 

Since I simply *link* to openneuro datasets, I can technically just import *all* datasets.. but for now I am only crawling datasets with >30 downloads to keep the list of openeuro datasets short-ish. The importing is triggered by cron right now and I can run it as often as necessary. In the near future, however, I am planning to import openneuro datasets on-demand - triggered by user wanting to access it. In a sense, you can just assume that all openneuro datasets are imported on brainlife.

For the "Open in brainlife.io" button, would it be better to call it "Analyze on brainlife.io" instead? Also, since most openneuro users won't know what brainlife is, is there any way to add a small text explaining what brainlife is? I think you could show something like..

(Analyze on brainlife.io)  * brainlife.io can run various data processing tools such as fmriprep, freesurfer, mrtrix and many other apps to analyze and visualize your data.

franklin-feingold:
Thank you! @ckrountree is there a format/place that would work best for you for tracking and documenting? 

I think that is a good plan to move toward full support.

+1 for "Analyze on brainlife.io" and small text explaining what brainlife is

ckrountree:
[...]

PR #1140, opened 2019-05-27
Brainlife button #1127

david-nishi:
#1127 
Check for dataset in Brainlife in dataset GraphQL query and present a link "Analyze on brainlife.io" next to Download Link if extant.


packages/openneuro-server/graphql/resolvers/dataset.js
```diff
@@ -9,11 +9,25 @@ import { permissions } from './permissions.js'
...
+    return false
+  }
+}
+
+export const dataset = async (obj, { id }, { user, userInfo }) => {
+  const dataset = await checkDatasetRead(id, user, userInfo).then(() =>
+    datalad.getDataset(id),
+  )
+  dataset.onBrainlife = await checkBrainlife(id)
+  return dataset
```
nellh: This forces queries which do not request onBrianlife to wait for the field to resolve and could lead to the field not being populated in some cases. Especially since this is an external service, we want it to only block queries which request it.

If you look at the end of this file, there's an object which defines the map of resolvers. You should just add a property onBrainlife which maps to the checkBrainLife function to fix this.

nellh (changes requested):

PR #1146, opened 2019-06-12
Add Kubernetes support and Helm template (chart)

nellh:
This adds a deployment chart for Helm 3 which can be used to deploy OpenNeuro on AWS EKS. A few changes would be needed to deploy on any Kubernetes cluster, this version depends on the ALB load balancer and EFS. This chart should be pretty easy to extend to support other load balancers and NFS.

@DaNish808 Can you help test this for regressions on docker-compose? There's several new options in config.env (documented in config.env.example) but otherwise the docker-compose setup should be unchanged. This also depends on the latest container image for datalad-service (3.1.5-19) so make sure you have that version to test.

@rwblair I'd love some feedback on the [README](https://github.com/OpenNeuroOrg/openneuro/blob/e57614fee24ffa40cf69add44b772c223f74f43c/helm/README.md)


david-nishi (approved):
Site functionality looks good and logs are clean.

rwblair:
README looks good, everything made sense and the glossary is much appreciated. That being said I have not executed the commands myself to verify they work. I'll try this sometime next week.

PR #1157, opened 2019-07-01
1133 generic error boundary

david-nishi:
Added error boundaries above dashboard dataset tab, dataset page, comments, validator, and editable fields to catch errors and prompt user with prepopulated freshdesk widget in modal. Error message is also presented in place of component with error.

### DIFF
--- a/packages/openneuro-app/src/scripts/datalad/fragments/freshdesk-widget.jsx
+++ b/packages/openneuro-app/src/scripts/datalad/fragments/freshdesk-widget.jsx
@@ -4,7 +4,7 @@ import config from '../../../../config'
 
 const prepopulatedFieldsQuery = prepopulatedFields => {
   const fieldQueries = Object.entries(prepopulatedFields)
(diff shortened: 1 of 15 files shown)


packages/openneuro-app/src/scripts/datalad/fragments/freshdesk-widget.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { getProfile } from '../../authentication/profile'
+import config from '../../../../config'
+
+const prepopulatedFieldsQuery = prepopulatedFields => {
+  const fieldQueries = Object.entries(prepopulatedFields)
+    .filter(([_, value]) => value)
```
nellh: I believe you can fix the linting error here with:

```
.filter(([, value]) => value)
```

nellh (changes requested):
This looks great, just needs a linting fix to merge.

Issue #1147, opened 2019-06-18
Add redirect dataset page to most recent snapshot


franklin-feingold:
when using the dataset url: (i.e. https://openneuro.org/datasets/ds001980) redirect to the latest snapshot (https://openneuro.org/datasets/ds001980/versions/1.0.1). This will redirect to the latest snapshot unless a specific version is given (i.e. https://openneuro.org/datasets/ds001980/versions/1.0.0)

nellh:
This redirect exists for users who do not have access to the draft page. I
believe there is an existing bug for an issue where you are taken to the
oldest snapshot instead of the newest one though.

effigies:
I think it makes sense to ensure that users with write access to the dataset still land on the draft page.

franklin-feingold:
Thank you for your messages. That makes sense, I found some that follow that behavior some that may be good to document that this is not the case: 
Document of some datasets that do not direct to latest snapshot:
[1978](https://openneuro.org/datasets/ds001978/versions/1.0.0)
[1984](https://openneuro.org/datasets/ds001984/versions/1.0.0)
[1972](https://openneuro.org/datasets/ds001972/versions/1.0.0)
[1299](https://openneuro.org/datasets/ds001299/versions/00001) - this one may be different issues from https://github.com/OpenNeuroOrg/openneuro/issues/1145
[1919](https://openneuro.org/datasets/ds001919/versions/1.0.0)

PR #1166, opened 2019-07-03
1147 redirect latest snapshot

david-nishi:
#1147
navigation to /datasets/dataset-id redirects to latest snapshot when user is not logged in or does not have edit privileges on dataset
only show readme edit button to users with edit privileges
fix regression: only show drafts to users with edit privileges


david-nishi:
reopen with branch on main repo

PR #1164, opened 2019-07-03
1138 dataset description changes

david-nishi:
extracts dataset description updates to datalad service

PR #1169, opened 2019-07-08
1138 dataset description changes

### LINKED ISSUE #1138: Changes to dataset_description.json are not saved in snapshot
Updating `dataset_description.json` through the web interface or by uploading a new copy will change values in the draft dataset, but snapshots will not contain the changes, instead reverting to the previous version.

Also, using the `Edit` option to edit the values produces a throbber that goes on for minutes (I didn't leave it to see if it ever refreshed).

Both of these issues were reported by user with a large dataset and reproduced on the very small ds001024 on the dev server. The user tried this with the "HowToAcknowledge" field, and I tried it with the "License" field.

david-nishi:
resolves #1138: extracts dataset description updates to datalad service
(reopened from PR #1164)

### DIFF
--- a/packages/openneuro-server/datalad/dataset.js
+++ b/packages/openneuro-server/datalad/dataset.js
@@ -13,7 +13,6 @@ import * as subscriptions from '../handlers/subscriptions.js'
 import { generateDataladCookie } from '../libs/authentication/jwt'
 import { redis } from '../libs/redis.js'
 import { updateDatasetRevision, draftPartialKey } from './draft.js'
-import { createSnapshot } from './snapshots.js'
 import { fileUrl } from './files.js'
 import { getAccessionNumber } from '../libs/dataset.js'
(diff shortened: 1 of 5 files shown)


packages/openneuro-server/datalad/dataset.js
```diff
@@ -290,7 +290,7 @@ export const addFileString = (datasetId, filename, mimetype, content) =>
 /**
  * Commit a draft
  */
-export const commitFiles = (datasetId, user) => {
+export const commitFiles = (datasetId, user, descriptionFieldUpdates) => {
```
nellh: I think adding this here confuses the goal of commitFiles. Maybe a better way is to remove the createSnapshot handling here and accomplish the automatic snapshot in another way.

nellh (changes requested):
Lets discuss the commitFiles change, otherwise this is good to go.

PR #1188, opened 2019-07-15
Prevent snapshot creation when Authors description field is empty

### LINKED ISSUE #1155: Block snapshot on DOI minting failure
Possibly the problem with ds001995 is that the curator did not add dataset authors, which @rwblair tells me is a required field for datacite.

If this is the case, then it should not be possible to snapshot without an author, in my opinion. I see a few options:

1) Fail validation without an author.
1) Automatically fill in Authors field with the user's name, if missing. Probably with a warning that we're doing that.
1) Block snapshots without an author.

david-nishi:
resolves #1155


david-nishi:
error is thrown when doi is not successfully created for snapshot and is gracefully handled on front end

Screen Shot 2019-07-29 at 10 37 29

Screen Shot 2019-07-29 at 10 37 52

PR #1204, opened 2019-07-24
bring back analysis info to homepage

### LINKED ISSUE #1185: Add ANALYSIS capability information back to homepage
Add the "we have analysis capabilities" information back to the landing page. This capability is part of a collaboration with brainlife.io; we have their button on some datasets already (will be scaled up to all over time and consistency reports from brainlife).

david-nishi:
Screen Shot 2019-07-24 at 15 15 51
resolves #1185


franklin-feingold:
Thank you for the rendering of the analysis page brought back! One difference from the previous set up to the current one is that the analysis happens off OpenNeuro (on BrainLife). Perhaps editing the current text to reflect that. In the future, we are hoping CBRAIN can be nicely integrated to help make the interface easy to use like before.

david-nishi:
Screen Shot 2019-07-25 at 13 42 18

nellh (approved):

PR #1217, opened 2019-07-29
1097-1148 Frontpage Updates #1

### LINKED ISSUE #1097: Add the dataset count and total participants shared in public DS to the ON main landing page
Add a section for more visually denoting the number of public datasets and total number of participants of those public datasets. The arrow denotes the space to use (spreading it out so there is enough room). The text size can be the same 'Get Data'. 

My thoughts were to have the number above the text. 

Screen Shot 2019-04-23 at 11 58 20 AM

david-nishi:
closes #1097 
Screen Shot 2019-07-29 at 05 40 22


nellh:
Participants are the study participants, not OpenNeuro users.

nellh (changes requested):
Participants here is the study participants, not OpenNeuro users.

packages/openneuro-server/graphql/resolvers/snapshots.js
```diff
@@ -21,6 +22,31 @@ export const snapshot = (obj, { datasetId, tag }, context) => {
...
+        foreignField: 'id',
+        as: 'summary',
+      },
+    },
+    { $project: { subjects: { $size: '$summary.subjects' } } },
+    { $group: { _id: null, participantCount: { $sum: '$subjects' } } },
+  ]).exec()
+  return Array.isArray(aggregateResult)
+    ? aggregateResult[0].participantCount
+    : aggregateResult
```
nellh: This seems like it would be unnecessary, when would the non-array result be useful? If it's the null case, I'd make it more explicit and return the null.

nellh (approved):
One question but otherwise this looks good.

PR #1232, opened 2019-08-05
1183 fix error boundaries

### LINKED ISSUE #1183: Partial upload throws error boundary on draft page
The partial page should be mostly available for datasets that have not finished uploading.

david-nishi:
closes #1183 

Error boundary not triggered by inaccessible dataset due to partial upload, instead renders resume and delete options:
localhost_9876_datasets_ds001047_versions_0
Hitting delete triggers confirm step:
localhost_9876_datasets_ds001047_versions_0 (1)

Dataset query error now given precedence over dataset page error, and query/mutation bound error boundaries now return more helpful errors:
Screen Shot 2019-08-05 at 16 01 34


packages/openneuro-app/src/scripts/datalad/routes/snapshot-content.jsx
```diff
@@ -60,72 +65,105 @@ const getSnapshotDetails = gql`
...
         <Helmet>
           <title>
-            {pageTitle} - {snapshot.description.Name}
+            {pageTitle} - {snapshotName}
           </title>
-          <meta name="description" content={snapshot.readme} />
-          <script type="application/ld+json">
-            {schemaGenerator(snapshot)}
-          </script>
+          {snapshot ? <>
```
nellh: I'm not sure about adding conditionals here for the entire snapshot, this component is meant to always receive some kind of snapshot and while making fields conditional might be necessary - if the whole snapshot is missing we should extract the logic to a parent component rather than add conditionals to the already pretty complex template here.

nellh (changes requested):
The "confirm" button with a delete icon is pretty confusing. Offering the delete option here is a good idea but we should make it very clear that clicking this button deletes your progress and will nullify the accession number and any existing snapshots / DOIs.

ckrountree:
@danish808 Maybe change text on "Confirm" button to be more explicit..."Confirm Deletion".  Also explain the result of deletion...

packages/openneuro-app/src/scripts/datalad/routes/snapshot-content.jsx
david-nishi: refactored

david-nishi:
Screen Shot 2019-08-14 at 09 27 20
updated with clearer warning on delete consequences

nellh:
@ckrountree @effigies @DaNish808 I think we should change this and delay it until the next release. It does what was requested in the issue but seeing it in action on a dataset gives me some concerns:

1. If you are uploading updated data to an existing dataset, this makes it very easy to accidentally delete the entire dataset losing the accession number and any DOI references.
1. Partial uploads are meant to be mostly visible (only things like summary data missing) so showing an alternative view should be a last resort.

ckrountree:
@nellh @effigies @DaNish808  I completely agree with the decision not to release this and to refactor it in this current sprint...especially in light of concern  no. 1.

effigies:
I'm okay with pushing off, if we're getting a normal-ish view of the partial dataset. Are we giving any advice on how to resume upload?

franklin-feingold:
That sounds good to me - may there be admin-level permissions setting that can be implemented so we can see this prompt and delete partial datasets while a stronger solution is designed?

david-nishi:
Screen Shot 2019-08-21 at 10 22 24

Freshdesk ticket opened with:
  - Subject - "Partial Upload Failure"
  - Description - "Missing snapshot in {dataset id} has resulted in inaccessible dataset page. Likely 
caused by partial upload error in dataset."
  - Email - "{user email}"

david-nishi:
simplified styles:
Screen Shot 2019-08-21 at 12 01 41

PR #1233, opened 2019-08-06
more helpful cli error messages

### LINKED ISSUE #642: Return a helpful error when uploading fails because of an invalid API key
When the server rejects the request, the uploader should print a useful error.

david-nishi:
resolves #642 

In case of unauthorized user:
Screen Shot 2019-08-05 at 18 32 36

In case of [missing .openneuro config file](https://github.com/OpenNeuroOrg/openneuro/blob/master/packages/openneuro-cli/README.md#errors): 
Screen Shot 2019-08-05 at 18 23 23


nellh (approved):

nellh:
This is merged but GitHub seems confused about that, closing it.

PR #1274, opened 2019-09-04
update Share blurb with BRAIN Initiative status

david-nishi:
The Publish text under Share Data now reads, "Publish your dataset in an NIH Brain Initiative approved repository where anyone can view, download, and run analyses on it."


packages/openneuro-app/src/scripts/front-page/front-page-tabs.jsx
```diff
@@ -142,14 +142,13 @@ class FrontPageTabs extends React.Component {
         firstImageClass: 'collabotate',
```
effigies, line 142: ?

```suggestion
        firstImageClass: 'collaborate',
```

PR #1276, opened 2019-09-09
1209 metadata form

### LINKED ISSUE #1209: Create metadata form from mockups
Refer to mockups in #1085

david-nishi:
resolves #1209 

- [x] metadata form has icon tool and page
- [x] metadata page is a form with editable (user input required) and disabled (validator/description/ON extracted data) fields
- [x] metadata page is redirected to from /publish if metadata does not exist for dataset
- [x] upon submission of metadata, user is redirected to /publish if user was redirected from /publish and back to previous page otherwise
- [x] edit dataset name as step 2 of upload is removed from process
- [x] metadata form sans disabled fields is presented as step 3 after validation and before license agreement
- [x] metadata has editable fields: associatedPaperDOI, species, studyLongitudinal, studyDomain, trialCount,  studyDesign, openneuroPaperDOI, dxStatus, and tasksCompleted
- [x] metadata has uneditable fields: datasetId, datasetUrl, firstSnapshotCreatedAt, latestSnapshotCreatedAt, adminUsers, datasetName, seniorAuthor, dataProcessed, ages, modalities
- [x] merge upstream


packages/openneuro-server/datalad/snapshots.js
```diff
@@ -67,26 +68,36 @@ const getSnapshotMetadata = (datasetId, snapshots) => {
...
-      tag,
-      oldDesc,
-    )
+    const snapshotDoi = 'mockDOI'
+    // const oldDesc = await description({}, { datasetId, revision: 'HEAD' })
+    // const snapshotDoi = await doiLib.registerSnapshotDoi(
+    //   datasetId,
+    //   tag,
+    //   oldDesc,
+    // )
```
nellh: This looks unintentional.

packages/openneuro-app/src/scripts/datalad/dataset/dataset-query-fragments.js
```diff
@@ -97,3 +97,67 @@ export const SNAPSHOT_ISSUES = gql`
...
+        }
+        ages {
+          type
+          container
+        }
+        modalities {
+          type
+          container
+        }
+      }
```
nellh: What's the reason for duplicating most of these fields in two formats via the context field?

nellh (commented):
Still testing this but the context question is the main thing I see.

nellh:
It doesn't need to be fixed for this PR but we should try to avoid the foxtrot merges and rely on rebase if possible.

packages/openneuro-app/src/scripts/datalad/dataset/dataset-query-fragments.js
david-nishi: As far as I can tell, the context is the most standard way to define terms in JSON-LD documents. We could convert it to the more literal syntax where IRIs are the keys, but I think that sequestering them in the context makes for easier data management. I'm not sure that we need the context on the front-end at the moment, but I thought it would be useful in case we wanted to make the metadata downloadable in its JSON-LD format.

nellh (approved):

PR #1289, opened 2019-09-23
581 format validator output

exbotanical:
**Abstract:** The outputs on the CLI tool are now formatted. Error outputs point the user to the specific files causing the failed validation (in the stead of the less-than-helpful "`[array]`" left unpacked and inaccessible). A broken URL has been updated.
***
I've implemented three changes, as follows:

**(1)** The staging URL that the OpenNeuro CLI points to was deprecated; it was being output to the user. I've updated it to its current, live incarnation.

I've imported the `consoleFormat` methods (already exposed given `bids-validator` is a dependency of OpenNeuro) into the uploads script. These methods were called to:

**(2)** format the summary output upon successful validation of a dataset upload
**(3)** format the errors, warnings output upon failed validation of dataset upload

### DIFF
--- a/packages/openneuro-cli/src/actions.js
+++ b/packages/openneuro-cli/src/actions.js
@@ -23,7 +23,7 @@ export const login = () => {
       type: 'list',
       name: 'url',
       message: 'Choose an OpenNeuro instance to use.',
-      choices: ['https://openneuro.org/', 'https://openneuro.dev.sqm.io/'],
+      choices: ['https://openneuro.org/', 'https://openneuro.staging.sqm.io/'],
       default: 'https://openneuro.org/',
     })
     .then(async answers =>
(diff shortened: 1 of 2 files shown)


nellh (approved):
Tested with some passing and failing datasets and output looks good!

david-nishi:
I'm wondering if theres anyone out there that's using the cli tool in a subprocess or something and is expecting the json output. Do you guys think it would be worth it to keep the json output available under an option (defaulting to off)?

nellh:
> I'm wondering if theres anyone out there that's using the cli tool in a subprocess or something and is expecting the json output. Do you guys think it would be worth it to keep the json output available under an option (defaulting to off)?

I don't think it is useful. The JSON was just mixed in with non-JSON output before so it wasn't easy to parse it out. If you want JSON validation results, use bids-validator directly. If you want to automate OpenNeuro uploads, you can still script this client or make your own based on the API pretty easily (though we should document this better).

david-nishi (approved):

PR #1293, opened 2019-09-25
Revise subjectmetadata

david-nishi:


nellh (approved):
Along with https://github.com/OpenNeuroOrg/datalad-service/pull/112 this fixes the subject metadata failure from the validation task with test datasets, schema changes look good :+1:

PR #1304, opened 2019-10-02
26 delete directory button

### LINKED ISSUE #26: Fix inability to delete directories
I want to remove one entire subject from a dataset.  I am able to delete the files within the subject folder, but unable to remove the subject directory, which causes a validation failure.  should include ability to delete folders.

david-nishi:
Included:
- delete option added to directories (including root dir; do we want to give this option to user?)
- files are removed from apollo cache when deleted
- tests for DeleteDir and DeleteFile

resolves #26

### DIFF
--- a/packages/openneuro-server/graphql/resolvers/dataset.js
+++ b/packages/openneuro-server/graphql/resolvers/dataset.js
@@ -178,17 +178,17 @@ export const deleteFile = async (
  */
 export const deleteFilesTree = (datasetId, fileTree) => {
   // drafts just need something to invalidate client cache
-  const { name, files, directories } = fileTree
+  const { path, files, directories } = fileTree
   if (files.length) {
-    const filesPromises = files.map(file =>
-      datalad.deleteFile(datasetId, name, file),
+    const filesPromises = files.map(({ filename }) =>
+      datalad.deleteFile(datasetId, path, filename),
     )
     const dirPromises = directories.map(tree =>
       deleteFilesTree(datasetId, tree),
     )
     return filesPromises.concat(...dirPromises)
   } else {
-    return [datalad.deleteFile(datasetId, name, { name: '' })]
+    return [datalad.deleteFile(datasetId, path, { name: '' })]
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/openneuro-server/graphql/resolvers/dataset.js
```diff
@@ -13,6 +13,7 @@ import { metadata } from './metadata.js'
 import * as dataladAnalytics from '../../datalad/analytics.js'
 import DatasetModel from '../../models/dataset.js'
 import fetch from 'node-fetch'
+import * as Sentry from '@sentry/browser'
```
nellh: This should be @sentry/node. browser is not installed for the server, so a clean node_modules leads to a server crash.

nellh (changes requested):
Looks pretty good but I haven't been able to test it end to end yet due to the Sentry module issue. Nice work on the tests.

packages/openneuro-server/graphql/resolvers/dataset.js
david-nishi: Ah thanks. The dangers of mindless copy-paste ;)

nellh (approved):

packages/openneuro-server/graphql/resolvers/dataset.js
```diff
@@ -148,10 +148,9 @@ export const deleteFiles = async (
   try {
     await checkDatasetWrite(datasetId, user, userInfo)
     await Promise.all(deleteFilesTree(datasetId, fileTree))
-    await Promise.all([
-      datalad.commitFiles(datasetId, userInfo),
-      pubsub.publish('draftFilesUpdated', { datasetId }),
-    ])
+    await datalad.commitFiles(datasetId, userInfo)
+    await pubsub.publish('draftFilesUpdated', { datasetId })
+    await Promise.all([])
```
nellh: This last line can be removed.

nellh (changes requested):

nellh (approved):

PR #1317, opened 2019-10-09
Avoid loading draft files when viewing snapshot routes

nellh:
The first commit here fixes a long standing bug where the snapshot page route was unused. Luckily we had nothing that depended on this until now :smile:

The second makes the draft files fragment conditional on the route entered by choosing the appropriate version of the query depending on whether the draft or snapshot route is entered. This nicely loads the draft once if you enter the page and then switch to a snapshot, but skips loading the draft if you enter via the snapshot. Going from snapshot -> draft now makes the extra query to obtain the draft files.


exbotanical (approved):
Didn't know the useQuery hook could accept a ternary. Nice solution :D

PR #1321, opened 2019-10-10
update core-js dependency, import polyfill

exbotanical:
Added core-js 3 as dependency so as to access and import the Object.fromEntries polyfill in the OpenNeuro App's entry point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,5 @@
       "!**/node_modules/**",
       "!**/vendor/**"
     ]
-  },
-  "dependencies": {
-    "core-js": "^3.2.1"
   }
 }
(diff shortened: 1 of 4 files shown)


exbotanical:
OK, fixed the package.json issue. Should be ready for approval!

exbotanical:
I see the test failure and will take a look at that later today.

exbotanical:
@nellh I'm not too familiar with codecov and how it interacts with / is affected by managing multiple packages. Any idea as to what might be causing this test failure?

David mentioned experiencing a similar issue with Lerna but doesn't recall the specifics, as it was quite a while ago. Maybe something to do with standardizing Yarn and not NPM?

david-nishi:
That issue cropped up when trying to upgrade lerna to v3

PR #1348, opened 2019-10-21
1210 custom support modal

exbotanical:
Should be up and running for all routes to the support modal. This is the first of many steps away from react-bootstrap and toward a consequentially smaller bundle size.

### DIFF
--- a/packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
+++ b/packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
@@ -2,49 +2,11 @@ import * as Sentry from '@sentry/browser'
 import React from 'react'
 import PropTypes from 'prop-types'
 import FreshdeskWidget from '../datalad/fragments/freshdesk-widget.jsx'
-import styled from '@emotion/styled'
-
-//styled components
-
-const Overlay = styled.div`
-  background: rgba(0, 0, 0, 0.8);
-  padding-top: 55px;
-  pointer-events: none;
-  bottom: 0;
-  left: 0;
-  position: fixed;
-  right: 0;
-  top: 0;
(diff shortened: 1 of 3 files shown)


packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -1,8 +1,50 @@
...
+  font-size: 34px;
+  padding: 12px 12px;
+  padding-top: 20px;
+  position: absolute;
+  right: 0;
+  text-align: center;
+  pointer-events: auto;
+  top: 0;
+  z-index: 100;
+`
```
nellh: These style components should probably be their own module, particularly since they are imported in both variants of the support modal and will likely see some reuse elsewhere in the future.

nellh (changes requested):

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
exbotanical: Good idea; I'll update in just a sec

david-nishi: Agreed, and it'd be convenient for them to be packaged together as a single Modal wrapper component.

david-nishi (changes requested):

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
exbotanical: @david-nishi Could do that, but do you think it'd be worth the opportunity cost? ( e.g. children, various props would still need to be called/passed) Also, theming requires a separate library. Might be misunderstanding you though

nellh (approved):

PR #1349, opened 2019-10-21
Validation Worker #1

david-nishi:
introduces a worker model via comlink


packages/openneuro-app/src/scripts/workers/validate.worker.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env worker */
+importScripts('https://unpkg.com/comlink/dist/umd/comlink.js')
```
nellh: I think we should stick to webpack es6 modules unless there's a good reason for this method.

nellh (changes requested):

nellh (approved):

PR #1353, opened 2019-10-23
1331 responsive homepage

exbotanical:
These changes make the front page responsive -> the containers are now flex and the gray offset on the right side of the page is now gone. I've also added a media query that draws a small inset border between the green "recent activity" cards; via mobile, they now appear as two discrete entities and the typographic hierarchy of the headers remains apparent (whereas before, this emphasis was lost on mobile). 

Last, I've added a media query that works in concert with the inset border -> this styling limits the prepopulated 'top datasets' and 'most active' lists to 3 items on mobile (as opposed to 5). This makes the vertical height of the view a bit shorter and conserves some real estate for the mobile view.

### DIFF
--- a/packages/openneuro-app/src/sass/crn_app/_elements-wrappers.scss
+++ b/packages/openneuro-app/src/sass/crn_app/_elements-wrappers.scss
@@ -52,7 +52,8 @@ body {
     font-family: $font-family-cabin;
     font-family: var(--font-family-cabin);
     letter-spacing: .03em;
-    background: $e5;
+    background: $white;
+    display: flex;
 }
 
 .main {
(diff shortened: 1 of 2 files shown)


david-nishi (approved):
Tested on Galaxy S6 Chrome: looks great!

nellh (changes requested):
This decenters the viewport for desktop in Chrome and Firefox:

Screenshot from 2019-10-29 15-19-17

On mobile, things look good in Chrome and Firefox. I think the condensed top datasets view is great. In Safari, the page horizontally scrolls but does look better on initial load.

exbotanical:
@nellh 
> This decenters the viewport for desktop in Chrome and Firefox:

Fixed! Thanks

nellh (approved):

Issue #1340, opened 2019-10-16
Fix FAQ title element arrows on mobile view


ckrountree:
The FAQ page is surprisingly unadulterated in mobile view. It's rather responsive, with the minor exception of the >> arrows that precede each title element. This isn't a pressing issue.

ckrountree:
Looks good, just needs code review.

Issue #1333, opened 2019-10-16
Fix dataset title sizing


ckrountree:
The h4 "dataset-name" elements (dataset titles) overflow their containers.

ckrountree:
Looks good, just needs code review...

PR #1354, opened 2019-10-23
responsive dataset title and faq page

exbotanical:
These changes address issues:

 - [[ON-1333]](https://github.com/OpenNeuroOrg/openneuro/issues/1333)
 - [[ON-1340] ](https://github.com/OpenNeuroOrg/openneuro/issues/1340)

Note: The sort, dataset summary, and filter (on private dashboards) issues are being postponed for further review of certain design decisions that need to be made. For now, the preceding two issues are addressed in this PR.

### DIFF
--- a/packages/openneuro-app/src/sass/crn_app/_dashboard.datasets.scss
+++ b/packages/openneuro-app/src/sass/crn_app/_dashboard.datasets.scss
@@ -61,7 +61,7 @@
     border-bottom: 1px solid $ddd;
 
     .filters {
-        float: right;
+        float: left;
         text-align: right;
         padding: 0;
         margin: 0;
@@ -74,7 +74,7 @@
             font-weight: 600;
             color: $c-77;
             text-align: right;
-            padding: 0 10px 0 0;
+            padding: 0 10px 0 20px;
 
         }
 
@@ -179,6 +179,7 @@
     }
 }
 
+
 // Dashboard teaser layout for Datasets ========================
 
 .dashboard-dataset-teasers {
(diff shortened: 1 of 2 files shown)


packages/openneuro-app/src/sass/crn_app/_dashboard.datasets.scss
```diff
@@ -362,20 +370,25 @@
...
             .summary-data {
                 display: inline-flex;
-                padding: 5px 20px;
+                padding: 2px 15px;
                 font-size: 12px;
-                line-height: 12px;
-                color: $c-33;
-                text-align: left;
-                border-right: 1px solid $f4;
+                color: #333;
```
david-nishi, line 384: i think this will get inherited from .minimal-summary

david-nishi (approved):

nellh (changes requested):
The FAQ change works great.

I'm running into some problems with the dashboard changes. Screenshots from Firefox but it seems about the same in Chrome.

Sometimes the summary data is not visible at all:
Screenshot from 2019-10-29 15-51-36

It is also breaking the alignment of the design and text is clipped at the bottom (see the G characters):
Screenshot from 2019-10-29 15-46-52-2

Looks like this introduces some variable height in some cases to these rows, which will result in rendering issues with the virtual scroll implementation.

exbotanical:
> Looks like this introduces some variable height in some cases to these rows, which will result in rendering issues with the virtual scroll implementation.

I should've marked this WIP. After meeting with Eric, there are going to be some more changes on the dashboard (e.g. removing most of the summary data tabs on mobile, removing the 'public dashboard' tab at the top left, et al); I'll make sure the issues you cited are fixed when I address them. Thanks for bringing that to my attention

Issue #1356, opened 2019-10-23
Implement background fetch for downloads


nellh:
This would allow for better download performance and reliability but is currently only implemented by Chrome and other Blink browsers.

The implementation needs to detect if a browser supports this feature and fall back to fetch or xhr if it does not.

ckrountree:
Related: Deeper discussion how this plays in the big picture. https://github.com/OpenNeuroOrg/openneuro/issues/1344#issuecomment-545605251

ckrountree:
Background fetch has been implemented but it is not being used because it didn't work well in Chrome 85 (current release of Chrome). 
Background fetch is ~75% of the implementation. So, with the current implementation, it wouldn't show any upload progress but turning it on should successfully upload once Chrome fixes it.

ckrountree:
After implementing a rough version of this, we found it isn't really possible with the current browser APIs without making multiple copies of each file. This is incredibly inefficient.  Instead we should implement a progress indicator for the native file downloads.

PR #1362, opened 2019-10-24
Native file API download support

nellh:
This adds a new web download method which uses the newly proposed [native file API](https://wicg.github.io/native-file-system/) to write downloaded streams to disk directly, skipping the zip bundling step.

This requires Chrome 78 or later and the [native file API flag enabled](https://web.dev/native-file-system/#enabling-via-chrome:flags).

This version does not show any progress during download as that will be handled by #1356 - you should see files successfully written to the selected directory.


nellh:
Key limitation of this version is https://crbug.com/998074 which requires you to select your target directory in 5 seconds or less.

PR #1369, opened 2019-10-29
Raise NO_AUTHORS to error

david-nishi:


nellh (changes requested):
Looks like you have an extra file in the PR.

nellh (approved):

Issue #1335, opened 2019-10-16
Fix view of Data Summary on mobile


ckrountree:
At first glance, it looks as though some summaries are just not being loaded.  However, the Dataset Summaries may actually be present/loaded but are being pushed below container parameters by the metrics.  If needed, increase container size for mobile.  All of summaries need to be present.

ckrountree:
I will figure out the best metadata to keep on mobile "list" view after looking into Google analytics.

nellh:
We could add some instrumentation for this in analytics but I believe the only way this is currently recorded is Apollo Engine.

ckrountree:
Over the past year - Google Analytics suggests that the following are the top 5 most requested search criteria (in order of most to least):

- eeg/EEG
- fmri 
- depression
- mri
- schizophrenia

From this list, it would seem that only the _Modalities_ is represented here (from the data summary list).  Let's discuss @MatthewZito .

ckrountree:
Based on this information, for the list of datasets on the mobile interface, perhaps we only display dataset:
- Title
- Tasks
- Modalities 

We can ellipse the longer text for tasks and modalities, as was done for titles.  Also, keep the boxes around each listed dataset the same height.

ckrountree:
Over the past year - Google Analytics suggests that the following are the top 20 most requested search criteria (in order of most to least):

- eeg/EEG  (_Available Modalities_)
- fmri/fMRI  (_Available Modalities_)
- depression  (_DX metadata_ as 'other')
- mri/MRI  (_Available Modalities_)
- schizophrenia  (_DX metadata_)
- brain
- alzheimer  (_DX metadata_)
- resting state/resting/rest
- parkinson  (_DX metadata_ as 'other')
- stroop
- memory
- tumor
- sleep
- emotion
- autism   (_DX metadata_ as 'other')
- dwi/diffusion/dti  (_Available Modalities_)
- bipolar  (_DX metadata_ as 'other')
- ucla
- ds000030
- midnight

@MatthewZito

ckrountree:
@MatthewZito For mobile, remove "available" on tasks and modalities for saving real estate. Make sure that it reads "no available..." if none are present.

Issue #1334, opened 2019-10-16
Fix Responsiveness of dataset metrics container


ckrountree:
The metrics and accompanying icons overflow (and are hidden/cut off by) their containers.

ckrountree:
Let's see if the metrics are important for mobile view...

PR #1373, opened 2019-10-30
mobile-only dashboard styling

exbotanical:
Addresses:

[1333](https://github.com/OpenNeuroOrg/openneuro/issues/1333)
[1334](https://github.com/OpenNeuroOrg/openneuro/issues/1334)
[1335](https://github.com/OpenNeuroOrg/openneuro/issues/1335)

You'll notice much of the metadata/dataset details are gone (e.g. uploader name, upload date, much of the data summary, status icon). In the last meeting regarding the mobile layout, it was decided that we should omit these items. The data summary was abridged to tasks and modalities. 

Notes:

- I'm wondering if we could do without the word "available" on the data summary. This could save space and allow more keywords to appear.
- I tried testing the style changes with a variety of metrics icons and numbers thereof (e.g. 10,000 views) to best simulate the prod version of the site. I think the testing was sufficient, but if I missed an edge case or something, do let me know.


exbotanical:
`/packages/openneuro-app/src/scripts/datalad/fragments/__tests__/__snapshots__/dataset-uploaded.spec.jsx.snap` uses the diction "_almost_ 3 years ago" for the shapshot whereas the correct format is "_about_ 3 years ago". It's causing a test failure, so I've updated this the most recent commit on this PR.

packages/openneuro-app/src/sass/crn_app/_dashboard.datasets.scss
```diff
@@ -50,6 +50,14 @@
     }
 }
 
+@media only screen and (max-width: 765px) {
+    .tabs.nav.nav-pills > li {
+        a { 
+        display: none;
```
david-nishi: This could use an indent

packages/openneuro-app/src/sass/crn_app/_dashboard.datasets.scss
```diff
@@ -356,8 +395,28 @@
...
+                    box-sizing: border-box;
+                    padding: 10px 2px 0px 0px;
+                    align-items: center;
+                }
+            }
+            @media only screen and (max-width: 765px) {
+                .status-container {
+                    display:none;
+                }
             }
```
david-nishi, line 417: Media-queried .metric-container and .status-container styles could be bundled together since they share the same type, feature, and level.

david-nishi (changes requested):
Looks so much better! Vertically stacking the metadata and metrics fits really well and everything that should be display is visible now. 

There are a couple of small fixes that could be made to the panel styles right around the high range of the media queries, including title text with cut off descenders and a slight overlap of the metadata and metrics containers.

Screen Shot 2019-11-07 at 10 34 32

exbotanical:
@david-nishi I'm going to remove the 'available' in the tasks/modalities before you review this.

david-nishi:
Was just about to comment on that. After that's done this is good to go.

packages/openneuro-app/src/scripts/datalad/fragments/dataset-summary.jsx
```diff
@@ -83,7 +83,7 @@ export default class Summary extends React.PureComponent {
     if (items && items.length > 0) {
       return (
         <span>
-          <b>Available</b> {type} : {items.join(', ')}
+          <b className="desktop">Available</b> {type} : {items.join(', ')}
```
david-nishi, line 86: What do you think about making this class "desktop-only"?

david-nishi (approved):

PR #1377, opened 2019-11-04
1374 delete file subscription

david-nishi:
Replaces all old file and directory delete operations with subscription model.
depends on #1376


packages/openneuro-app/src/scripts/datalad/subscriptions/files-subscription.jsx
```diff
@@ -0,0 +1,78 @@
...
+  switch (action) {
+    case 'DELETE':
+      return deleteFilesReducer(payload, draft)
+    default:
+      return { ...draft }
+  }
+}
+
+const FilesSubscription = ({ datasetId }) => (
+  console.log('subscribed to ', datasetId),
```
nellh, line 42: Should take out this extra console.log - though it was helpful for review.

nellh (approved):

PR #1379, opened 2019-11-05
1208 show commit hash

david-nishi:
Show relevant abbreviated git hash on snapshot and draft pages.

Full git hash is displayed on hover and can be copied to clipboard.


nellh:
I think this should be more subtle, it draws a lot of attention to it but it's really only useful information for an advanced user who will likely be familiar with how to find it. Thoughts?

david-nishi:
We talked with Eric about the design and he made a good point about it belonging under the file-tree component, possibly in the same kind of grey outline container as the other two items in that column. I made the choice to add the header with the same styles, but I agree that it could be more subtle. Maybe the whole thing could be styled more like a footnote?

```
      File Tree Container
|______________________________________________________________|
                   Git Hash: 2h8y8uehr2o84uy243uh2o34uh298o34hu


      File Tree Container 
|______________________________|
 Git Hash: 2h8y8uehr2o84uy243uh
                 2o34uh298o34hu
```

nellh:
One more idea towards making it more subtle, GitHub shortens the hashes to the first 7 characters and shows the full hash as a tooltip. 7 characters is usually enough to be unique for a repo.

david-nishi:
That sounds like the way to go!

nellh (approved):

PR #1395, opened 2019-11-20
Update to bids-validator 1.3.10

nellh:
A few fixes were needed to get bids-validator 1.3.10 working for webpack on master.

Test suite is still broken here, making another PR with fixes for tests.


david-nishi (approved):
Works in uploader and from snapshot

PR #1399, opened 2019-11-23
WIP 1282 filetree loading

exbotanical:


exbotanical:
I just noticed the edit to the resolver makes it such that the owner of the dataset can't update/delete from the filetree. Going to keep working on that tonight to fix it.

PR #1405, opened 2019-11-26
1313 dataset error routes

exbotanical:
The error boundary is redirected in the following cases:

1. Dataset does not exist
2. Dataset/draft is private and user lacks permissions
3. Dataset/draft is private and user has permissions, but a pending process is blocking visibility
4. Dataset is public but there is no snapshot 

Update: 
- elif cleanup / refactor checks
- add styles
- discrete FreshDesk component

Pending / Blocked:
- pull gql `getDataset` message to check against for case no. 2


nellh:
FYI the error state is returned to the client already, but discarded by Apollo. Changing the query error policy to pass it to the receiving component will show that it's a permission denied error in the case the user lacks read or write access.

exbotanical:
> FYI the error state is returned to the client already, but discarded by Apollo. Changing the query error policy to pass it to the receiving component will show that it's a permission denied error in the case the user lacks read or write access.

Thanks! This is exactly the issue I was working on :D

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -61,46 +142,32 @@ class ErrorBoundary extends React.Component {
...
-      </>
-    ) : (
-      this.props.children
-    )
+    const { loading } = this.props
+
+    if (error) {
+      if (message) {
+        return <DatasetRedirect message={message} />
+      } else if (loading) {
```
david-nishi: We can remove this conditional. The loading state precedes any response, so if we wanted to include a spinner for it that would be the first condition to handle.

david-nishi (changes requested):

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
exbotanical: Makes sense. Also - on second gander - I think the final else (return children) can be a standalone return statement; I've included this as well.

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -142,16 +143,17 @@ class ErrorBoundary extends React.Component {
   }
 
   render() {
+    const error = this.props.error || this.state.error
```
david-nishi: we should reverse the order of assignment here to `this.state.error || this.props.error`

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -142,16 +143,17 @@ class ErrorBoundary extends React.Component {
...
     const { supportModal, message } = this.state
     const { subject, description } = this.props
-    const error = this.props.error || this.state.error
     const { loading } = this.props
 
+    if (loading) {
+      return <Spinner text="Dataset is currently being updated." active />
```
david-nishi: Since loading occurs when the dataset is simply being fetched, the text might be changed to something like, "Dataset is currently being fetched." or "Fetching dataset..."

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -165,9 +167,8 @@ class ErrorBoundary extends React.Component {
           />
         )
       }
-    } else {
-      return this.props.children
     }
+    return this.props.children
```
david-nishi, line 175: nice!

david-nishi (changes requested):

david-nishi (approved):

nellh:
It looks like this branch no longer shares history with upstream for some duplicate commits. That should be resolved before merge in order to prevent adding duplicate commits to master.

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -28,79 +101,61 @@ class ErrorBoundary extends React.Component {
...
-      () => true,
-    )
+    return getDerivedStateFromErrorOnCondition(error, () => true)
   }
-
   componentDidCatch(error) {
+    let message = this.state.message
+    if (!this.props.dataset) {
+      message = 'This dataset does not exist.'
+    } else if (this.props.dataset && this.props.dataset.snapshots.length < 1) {
```
david-nishi, line 129: Message text is good, but I think users should be pointed to the freshdesk modal on this condition.

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
```diff
@@ -28,79 +101,61 @@ class ErrorBoundary extends React.Component {
...
-    ) : (
-      this.props.children
-    )
+    if (error) {
+      if (message) {
+        return <DatasetRedirect message={message} />
+      } else {
+        return (
+          <FreshdeskModal
+            error={message}
```
david-nishi: message here is always falsy. error should just be the original error object (to be used by FreshdeskWidget).

packages/openneuro-app/src/scripts/errors/errorBoundary.jsx
david-nishi: 'This dataset has no associated snapshots.'

david-nishi (changes requested):

david-nishi (approved):

Issue #1375, opened 2019-11-04
Convert file upload & update operations to subscription model


david-nishi:
current status:
- front end components and reducers complete
- need to file hashes from datalad-service

ckrountree:
Release v3.9

PR #1406, opened 2019-12-03
1375 update file subscription

david-nishi:
Uploads and updates are now handled by the Apollo subscription model.


nellh (changes requested):
The first file update works fine but a second file update of the same file or another file in the same tree breaks the file tree display and shows 'sub-01:anat:sub-01_T1w.nii.gz' and multiple copies.

david-nishi:
Thanks Nell, should be fixed now

david-nishi:
Thanks @nellh, should be fixed now

nellh (approved):
Looks good now.

effigies:
Linking #1375

PR #1415, opened 2019-12-09
Dataset page progressive file loading

### LINKED ISSUE #1282: Reduce unused file tree content loading on dataset and snapshot pages
The dataset file tree loads a lot more data than is initially displayed. While @defer would help here, we should try to avoid loading the extra data at all if the user doesn't interact with it.

ds001734 is a good example of where this is a problem.

nellh:
This fixes #1282 by dividing datasets up into top level files and then a load operation for each branch off the top level tree.

There's a few small things left before this is ready for review:
- [x] The loading state styles are bad
- [x] Needs another pass to find anything which depends on more than top level files in the default query results


david-nishi (approved):
All looks good to me with the file tree loading changes. It's not really a part of this issue, but I also hit that front-page bug and i think an easy fix would be switching the order of the conditionals here:

https://github.com/OpenNeuroOrg/openneuro/blob/8c196116121d88224ff4146af2d5d8dae6acab9b/packages/openneuro-app/src/scripts/front-page/front-page-top-datasets.jsx#L148-L153

nellh:
Directory sort bug found during review is now fixed and the fix is covered by a new unit test.

david-nishi (approved):

Issue #983, opened 2018-10-25
Display accession number on dataset/snapshot landing page


chrisgorgo:
We refer to it in the CLI documentation

ckrountree:
We can code review this but it may benefit from some design discussion before merging.  Another potential approach would be to make the component that currently displays the git hash also capable of displaying the accession number...if we decide to have them displayed the same way (design decision).

ckrountree:
@MatthewZito let's follow this design for the placement of this accession number:
<img width="580" alt="Screen Shot 2020-01-06 at 3 31 54 PM" src="https://user-images.githubusercontent.com/27043528/71857221-91a81f80-309b-11ea-84a9-3ce98253f45b.png">

PR #1424, opened 2019-12-19
add accession number to dataset and snapshot landing pages

exbotanical:


packages/openneuro-app/src/scripts/datalad/fragments/accession-number.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+const Container = styled.div({
+  textAlign: 'right',
+  padding: '5px 10px',
+  color: '#565656',
+  fontSize: '12px',
+})
+
+const AccessionNumber = ({ datasetId }) => (
+  <div className="dataset-git-hash">
```
david-nishi: Since the git hash and accession number components do the same thing, I think it'd be better to generalize one of them so that it can take whatever text and tip is necessary as props (and maybe the padding style too). Otherwise, we should at least change this classname.

david-nishi (changes requested):

exbotanical:
@david-nishi  Just a heads up - we had a quick design meeting on this; feel free to check out the [mockup in Kate's comment](https://github.com/openneuroorg/openneuro/issues/983#issuecomment-571365181).

packages/openneuro-app/src/scripts/datalad/fragments/dataset-summary.jsx
```diff
@@ -68,7 +74,10 @@ export default class Summary extends React.PureComponent {
             <hr />
             <h5>
               <b>
-                {files}, {size}, {subjects}, {sessions}
+                {accessionNumber}
+                <h5>
```
david-nishi: now that files, size, subjects, and sessions are separated out into their own h5 element, lets remove the h5 and b wrappers (lines 75 & 76)

david-nishi (changes requested):
looks great!

david-nishi:
I'm also wondering if it should be included in the minimal summary as well. Was that discussed during the design talks?

packages/openneuro-app/src/scripts/datalad/fragments/dataset-summary.jsx
exbotanical: The others are `span` tags, but I removed the redundant bold.

exbotanical:
> I'm also wondering if it should be included in the minimal summary as well. Was that discussed during the design talks?

Just spoke with Kate - we're not including it in the minimal (I think we're actually removing some things from it soon). Thanks

david-nishi (approved):

PR #1427, opened 2019-12-20
Disable file tree expanding animation due to layout thrashing

### LINKED ISSUE #1425: Disable file tree animations for large file load operations
An opening animation was added in #1282 to make newly loaded file trees indicate the change in state. A few datasets with particularly large single directories now have extremely poor performance until the animation completes.

nellh:
Fixes #1425 by just disabling it entirely. Profiling this shows it results in some deep layout thrashing for very large trees which is a significant regression.


david-nishi (approved):
Hooray for deleted code!

Issue #1332, opened 2019-10-16
Clean up dataset sort button


ckrountree:
Perhaps sorting could be a dropdown menu? The items are stacking and occupying valuable view-space on mobile.
![Screen Shot 2019-10-16 at 9.47.44 AM.png](https://images.zenhubusercontent.com/5bdb67b0afbb1d3bccc8e657/8f1ad72c-c278-4b5c-812d-c5f2d9d2430b)

ckrountree:
Here is the idea that came out of the meeting with Eric:
<img width="417" alt="MOBILE idea for ds listing" src="https://user-images.githubusercontent.com/27043528/67521135-1f5c9d00-f65f-11e9-91e0-3175539500cf.png">

I will look into Google Analytics for the metadata listings for mobile.

ckrountree:
The overall design here has changed since our Search design has been discussed.

PR #1428, opened 2019-12-20
1332 responsive sort dataset

exbotanical:


packages/openneuro-app/src/sass/crn_app/_dashboard.datasets.scss
```diff
@@ -187,6 +188,26 @@
     }
 }
 
+@media only screen and (max-width: 765px) {
```
david-nishi: We use 765px as the breakpoint in many places in the code. I think it would be valuable to pull this out into a variable with a descriptive name like like $tablet-width or something.

packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-query.jsx
```diff
@@ -161,6 +162,7 @@ const datasetQueryDisplay = isPublic => ({
       refetch={refetch}
       queryVariables={variables}
       publicDashboard={isPublic}
+      isMobile={useMedia('(max-width: 700px) ')}
```
david-nishi: The mobile breakpoint is 765 elsewhere so this creates an odd looking intermediate state.

packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-tab.jsx
```diff
@@ -71,7 +72,7 @@ const DatasetTab = ({
       </div>
       <div className="filters-sort-wrap clearfix">
```
david-nishi: The "Filter by:" label overlaps with the sort-by input element when the page width is 690-700 px.

packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-sorter.jsx
```diff
@@ -50,22 +51,50 @@ SortField.propTypes = {
...
-    ))}
-  </>
-)
+const DatasetSorter = ({ queryVariables, refetch }) => {
+  const isMobile = useMedia('(max-width: 700px) ')
+  const onChange = event => {
+    const newQueryVariables = { ...queryVariables }
+    newQueryVariables.orderBy = {}
+    newQueryVariables.orderBy[event.target.value] =
+      queryVariables.orderBy[event.target.value] === 'descending'
```
david-nishi: This will always evaluate to false because the event only fires on change. Reversing the sort order will need some additional button/options or some fancy logic (design Q).

david-nishi (changes requested):
I think we should try to get a little closer to [this comp](https://github.com/openneuroorg/openneuro/issues/1332#issuecomment-546081792). With infinite scrolling on the dashboard, the "More" button and "Showing 1-n of N datasets" doesn't really apply, but we should have the the "Public/My Datasets" heading and more alignment between the styles of the search and sort-by inputs (at least left-right alignment and width). 
localhost_9876_dashboard_datasets

packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-sorter.jsx
exbotanical: Just adding a note for documentation: the now revised solution entails being able to include ascending / descending sort on mobile. Originally, the toggling caret wasn't going to work for a dropdown, but given we will imminently be consolidating / removing some of the sort options, we can add both asc/desc options for each sort item, respectively.

david-nishi (changes requested):
Code looks good, but content is not responsive under 765 px width breakpoint. A min-width for the page makes sense, but above that and especially closer to 765 it looks very cluttered.

david-nishi (approved):

PR #1434, opened 2020-01-02
1337 mobile dataset page

exbotanical:


david-nishi (approved):
Looks good to me! Didn't know you could use mixins like that, very cool.

Might be a good idea to pass this by design before final review.

PR #1438, opened 2020-01-06
1330 collapse mobile menu

exbotanical:


packages/openneuro-app/src/scripts/nav/navbar.jsx
```diff
@@ -60,11 +67,21 @@ class BSNavbar extends React.Component {
...
   }
 
+  setNavExpanded(expanded = true) {
+    this.setState({ navExpanded: expanded })
+  }
+
+  closeNav() {
+    this.setState({ navExpanded: false })
+  }
+
```
nellh, line 78: You can simplify this to not need the closeNav function at all and make it more robust to future additions of new menu items by listening for the route change instead of using a callback on each click event.

```javascript
  componentDidUpdate(prevProps) {
    if (this.props.location !== prevProps.location) {
      this.setNavExpanded(false)
    }
  }
```

nellh (changes requested):
One change to simplify this but otherwise it's working well.

nellh (approved):

Issue #1439, opened 2020-01-07
Update Search field and Results on Public Datasets page


ckrountree:
Based on the Search design for desktop, please make the following changes for mobile:

- [x] Change title from _Public Datasets_ to (-->) _**Public Dataset Results**_
- [x] Move the Search bar above the title
- [x] Add the number of results shown, under the title but above the sort field

![Screen Shot 2020-01-06 at 3.48.19 PM.png](https://images.zenhubusercontent.com/5bdb67b0afbb1d3bccc8e657/eb0a1a5c-0f44-42e4-83ee-e62e373eee65)

Issue #1401, opened 2019-11-25
Design wireframe prototype for Search pages


ckrountree:
We need wireframes:
- [ ] Design search input page; BASIC and ADVANCED modes
- [x] Design RESULTS page

Architecture design in ticket #1387 

Refer to #1314 for criteria (facets)

ckrountree:
**IDEAS:**

### Homepage
- "A" - The user can enter a search parameter on this front-facing page (in this example - _EEG_).  Once triggered, the user will be taken to a _PUBLIC DATASET RESULTS_ page. (see the following for more information from here)
![ON search front](https://user-images.githubusercontent.com/27043528/70576588-2e8ba180-1b5e-11ea-9082-f3d4c5f64c8c.png)

### Public Dataset Results
- "A" - If the user arrived at this page via an initial search query from the homepage (in this example - _EEG_), the search box will show the original search parameter (ie: _EEG_). The user will also see all (only) associated facets will appear in 'buttons' below the search box ("B").  
- "A" - If the user arrived at this page via the PUBLIC DASHBOARD link, this search field will show the default _Search Datasets_.  Once User enters the desired search parameter(s) and clicks SEARCH, all (only) associated facets will appear in 'buttons' below the search box ("B").
- "B" - Hovering or clicking on an associated search facet will produce a drop-down of listed facet parameters (in this example, _Depression_ is marked).  When User selects a parameter, the facet 'button' will change color to denote the additional filter facet to search (in this example - _Diagnosis_ / _Depression_).
- "C" - Public Dataset header should be changed to _PUBLIC DATASET RESULTS_. If no search is performed, then all available/public datasets will be listed as the result.
- "D" - The number of search results served/listed should be noted (in this example - _2_).
- "E" - As currently seen, the user can sort the search results by _Created_ (date), _Name_ (dataset name, alphabetically), and _Acitivity_ (an algorithmic combination of Stars, Downloads, & Subscriptions).  Note: Uploader should be removed, as it took up real estate and did not add any benefit.  These sort labels should be given a bit more thought to be truly descriptive (ie: _Name_ could be construed as author, ds title, etc.)  We could be more declarative.
![ON search dashboard](https://user-images.githubusercontent.com/27043528/70578420-82988500-1b62-11ea-8c26-9fad3f14629d.png)

- "F" - This tab is redundant.  Remove.
[...]

PR #1445, opened 2020-01-07
1439 dataset results

exbotanical:
Just a few things to note:

These design changes are per [this mockup](https://github.com/OpenNeuroOrg/openneuro/issues/1439#issue-546003747). Note that the `Sort By:` that prepends the dropdown was captured in the screenshot used; it was not intended to be included in the design.

The ‘uploader’ field has been removed from sort, as we’re removing / consolidating many of these options [(wireframe)](https://github.com/openneuroorg/openneuro/issues/1401). Beyond 'uploader', the other ones are going to comprise their own ticket (#1446).

The title ‘Public Dataset Results’ is intended for both desktop and mobile views. 

The search results / number of datasets displayed is supposed to be displayed on both the private ‘My Datasets’ and public dashboard components with the diction “Results n”, without a colon. 

The caret in the dropdown was to be added as a visual cue so as to let users know this is indeed a dropdown / functional element. 

The sort dropdown has been widened to align with the dataset cards, as decided in a design meeting and amendment to the mock-up.


exbotanical:
Note: When selected, some of the options in the dropdown menu are too long to fit in the input field. They will either be overlapped by the caret, or clipped by the caret if its white background is given a higher z-index. 

This happens due to the (ascending) / (descending) appended to each option. One alternative here is to display a shortened "(asc)" / "(desc)" instead. This can be accomplished by making the option tags' inner text equal `` {`${i}:${sortField.field}`>{`sortField.field} (${sortField.order === ASC ? 'asc' : 'desc'})`} ``

Screen Shot 2020-01-07 at 2 27 36 PM

@ckrountree

david-nishi (approved):
Looks good!

PR #1456, opened 2020-01-13
886 out of date warning

david-nishi:
OpenNeuro App and CLI tool will log a warning to the user when their major/minor versions are out of date with the server.

### DIFF
--- a/packages/openneuro-cli/src/actions.js
+++ b/packages/openneuro-cli/src/actions.js
@@ -8,8 +8,10 @@ import { getDatasetFiles, createDataset } from './datasets'
 import { getSnapshots } from './snapshots.js'
 import { getDownload } from './download.js'
 
+const { version } = require('../package.json')
+
 export const configuredClient = () =>
-  createClient(`${getUrl()}crn/graphql`, getToken)
+  createClient(`${getUrl()}crn/graphql`, getToken, null, version)
 
 /**
  * Login action to save an auth key locally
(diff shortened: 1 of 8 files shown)


packages/openneuro-server/app.js
```diff
@@ -19,6 +19,8 @@ import * as auth from './libs/authentication/states.js'
 import { sitemapHandler } from './handlers/sitemap.js'
 import { setupPassportAuth } from './libs/authentication/passport.js'
 
+const version = require(`${__dirname}/package.json`).version
```
nellh: I think `import { version } from './package.json'` would be better here.

packages/openneuro-cli/src/actions.js
```diff
@@ -8,8 +8,10 @@ import { getDatasetFiles, createDataset } from './datasets'
 import { getSnapshots } from './snapshots.js'
 import { getDownload } from './download.js'
 
+const { version } = require('../package.json')
```
nellh: Same comment about import vs require.

packages/openneuro-client/src/client.js
```diff
@@ -69,7 +69,43 @@ const middlewareAuthLink = (uri, getAuthorization, fetch) => {
...
+  } else if (
+    serverMajor < clientMajor ||
+    (serverMajor === clientMajor && serverMinor < clientMinor)
+  ) {
+    // panic, then
+    console.warn(
+      'Your openNeuro client is out of date. We strongly recommend you update to the most recent version for an optimal experience.',
+    )
+  }
+}
```
nellh, line 93: For consistency with other version comparisons this should use the semver library we're already using elsewhere.

nellh (changes requested):

nellh (approved):

PR #1472, opened 2020-01-16
Add test for loading state behavior on admin users route

nellh:
This previously seems to have been working due to the fallback on rendering zero users. This adds a test for the loading state and displays a spinner while data is loading.


exbotanical (approved):
Looks great.

PR #1474, opened 2020-01-16
1473 prepopulate tasks field 

exbotanical:
This change pre-populates the metadata field for tasks completed.


david-nishi (approved):
Looks great!

PR #1480, opened 2020-01-22
fix sort and filter offsets

exbotanical:


nellh (changes requested):
This pushes "Filter By:" over to the left and results in some extra spacing below the tabs.

Screenshot from 2020-01-23 15-23-28

exbotanical:
> This pushes "Filter By:" over to the left and results in some extra spacing below the tabs.
> 
> Screenshot from 2020-01-23 15-23-28

Good catch, thanks!

nellh (approved):

PR #1498, opened 2020-01-30
Apply myDatasets filter only if neither public or saved is set

nellh:
Fixes a logic error where the saved dashboard also applies the myDatasets filter, which hides any starred datasets that are public but which you do not have permissions added for.


exbotanical (approved):
Thanks for the fix : )

PR #1500, opened 2020-01-31
1002 subscribers

david-nishi:
Switches from ApolloGraphQL mutation update to subscription model for delete dataset, dataset description updates, snapshot updates, and permission updates. Fixes edge-case bug in file updates.


nellh (approved):
I've been testing this and not run into issues, I think this is good to go :tada:

PR #1504, opened 2020-02-03
Implement search API and initial indexer

nellh:
This adds the GraphQL API for ElasticSearch under the namespace '/search'.

1. New openneuro-indexer package that can be run periodically - it runs once on startup with `docker-compose up` but can be rerun as needed
1. One "datasets" index covering the most recent snapshot values (#1416, #1417, #1418). The indexer will throw an error and skip datasets without any snapshots.
1. Limited portions of the ElasticSearch API are available under /search (to allow only the APIs we whitelist, right now this only makes 'indices' and 'search' available)
1. Typescript support is improved across the project but autogenerated types for the GraphQL schema were removed to simplify this PR.
1. openneuro-client createClient has a consistent name and new options argument, to allow for packages which insert their own ApolloLink objects
1. Added a generator to abstract pagination details from any Relay query style APIs.

The indexer outputs the accession number for each dataset as it is crawled:

```
indexer_1        | Indexing "ds001132"
indexer_1        | Indexing "ds001131"
indexer_1        | Indexing "ds001130"
indexer_1        | Indexing "ds001129"
```

A couple example queries to view indexes and perform a basic search:

```graphql
query {
  search {
    indices {
      get(index: "datasets")
    }
  }
}
```

```graphql
query {
  search {
    search(q: "dataset name")
  }
}
```


david-nishi (commented):
Really cool! Especially like the way the datasetGenerator is handled. All looks good, but I had to run the indexer manually to get search working the first time. Tried restarting everything a few times and encountered that on maybe a third of them. Might be worth bumping up the retryLink max attempts?

nellh:
> Really cool! Especially like the way the datasetGenerator is handled. All looks good, but I had to run the indexer manually to get search working the first time. Tried restarting everything a few times and encountered that on maybe a third of them. Might be worth bumping up the retryLink max attempts?

I've added two fixes that may help with this. The main one is a health check which should let docker-compose wait for the server to actually be ready to handle requests rather than a timeout on the GraphQL endpoint. The other is a minor permissions problem this branch introduced with the node_modules directory in the indexer, where the app container would overwrite files with root permissions. Let me know if this makes the initial start reliably index for you.

nellh:
One more fix, added a dependency on the ElasticSearch container for the indexer. That should also help with ordering when ElasticSearch starts up (especially on first run while it bootstraps).

david-nishi (approved):
Works 5/5

PR #1507, opened 2020-02-04
Search indexing improvements and basic dashboard integration

nellh:
This replaces the Google Custom search components with a new component for basic queries, along with integrating the dashboard with search results.


packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-tab.jsx
```diff
@@ -120,7 +115,7 @@ const DatasetTab = ({
     ) : (
       <ErrorBoundary error={error} subject={'error in dashboard dataset tab'}>
         <DatasetTabLoaded
-          data={data}
+          datasets={data.datasets || data.search}
```
nellh, line 118: I think this fork is not the best way to handle this but it keeps this reusable for now.

packages/openneuro-app/src/scripts/search/search-input.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useEffect, useRef } from 'react'
+import { useHistory } from 'react-router-dom'
+
+const SearchInput = () => {
+  const inputRef = useRef(null)
```
nellh, line 5: This is a ref because otherwise autocomplete causes a lot of DOM churn. Autocomplete isn't implemented here but that's why I ended up with a seemingly useless ref.

packages/openneuro-indexer/src/createIndices.ts
```diff
@@ -0,0 +1,19 @@
...
+  })
+  if (exists.statusCode !== 200) {
+    await elasticClient.indices.create({
+      index: 'datasets',
+      body: {
+        mappings: datasetsMapping,
+      },
+    })
+  }
+}
```
nellh, line 19: Moving this into the indexer keeps it responsible for any ElasticSearch configuration other than the connection.

packages/openneuro-indexer/src/createIndices.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const createIndices = async elasticClient => {
+  const exists = await elasticClient.indices.exists({
+    index: 'datasets',
+  })
+  if (exists.statusCode !== 200) {
+    await elasticClient.indices.create({
+      index: 'datasets',
+      body: {
+        mappings: datasetsMapping,
```
nellh, line 15: The mapping improves the results by typing fields for ElasticSearch. This is how we can tune indexing.

packages/openneuro-server/graphql/elastic-field.js
```diff
@@ -1,17 +0,0 @@
...
-})
-
-// Remove any paths we don't want to expose
-const clientTC = schemaComposer.createObjectTC(fieldConfig.type)
-clientTC.removeOtherFields(['search', 'indices'])
-
-export default {
-  ...fieldConfig,
-  type: clientTC.getType(),
-}
```
nellh, line 17: Dropped directly exposing the ElasticSearch API so that we could use the newer '@elastic/elasticsearch' client and avoid the extra dependency.

david-nishi:
One edge case I believe is that if is search run with an empty field the loading throbber runs indefinitely.

david-nishi (approved):

PR #1509, opened 2020-02-05
update api endpoint

exbotanical:
Added array type check on returned JSON. Root structure of any given document being returned via the regex on the endpoint is type array, so it shouldn't vary...

All documents: https://brainlife.io/api/warehouse/datalad/datasets?find=

### DIFF
--- a/packages/openneuro-server/graphql/resolvers/dataset.js
+++ b/packages/openneuro-server/graphql/resolvers/dataset.js
@@ -300,10 +300,10 @@ export const starred = (obj, _, { user }) =>
  */
 export const onBrainlife = async dataset => {
   try {
-    const url = `https://brainlife.io/api/warehouse/project?find={"openneuro.dataset_id":"${dataset.id}"}`
+    const url = `https://brainlife.io/api/warehouse/datalad/datasets?find={"path":{"$regex":"${dataset.id}$"}}`
     const res = await fetch(url)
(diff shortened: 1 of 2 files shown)


packages/openneuro-server/graphql/resolvers/dataset.js
```diff
@@ -299,12 +299,12 @@ export const starred = (obj, _, { user }) =>
...
  */
 export const onBrainlife = async dataset => {
-  try {
-    const url = `https://brainlife.io/api/warehouse/datalad/datasets?find={"path":{"$regex":"${dataset.id}$"}}`
-    const res = await fetch(url)
-    const body = await res.json()
+  const url = `https://brainlife.io/api/warehouse/datalad/datasets?find={"path":{"$regex":"${dataset.id}$"}}`
+  const res = await fetch(url)
+  const body = await res.json()
+  if (Array.isArray(body)) {
```
david-nishi: I think we should add the condition `&& body.length`

david-nishi (changes requested):

packages/openneuro-server/graphql/resolvers/dataset.js
exbotanical: Agreed; thanks

david-nishi (approved):

PR #1512, opened 2020-02-06
1510 add clientside redirects

exbotanical:
Updates OpenNeuro-App's React Router from v4 -> v5. 
Redirects by matching the dataset id param to a 'library' of datasets for which the id no longer equals the accession number.

Please let me know if you need me to do anything re: rebasing / merging what with the dependency update.


nellh (commented):
One minor tweak to suggest.

This will only work with datasets that do not exist or otherwise error because it is in the error boundary?

nellh (approved):

PR #1517, opened 2020-02-07
1151 mobile share datasets

exbotanical:
This PR adds a ‘share dataset’ button on mobile, and changes the current “share” to “manage dataset permissions”. This augmented diction is intended to better align these features’ names with what they actually do.

The new share button uses the Web Share API; it enables eligible users to either copy to their device’s clipboard the current dataset url, or share it via a variety of apps (notably the proprietary email and messenger clients on the user’s device). ‘Eligible’ user implies the contrary exists; unfortunately FireFox on Android does not support the API. It is also mobile-specific; it will not work on a desktop client and thus is a mobile-only feature.

Here’s a CodeSandbox with this button (sans the CSS) for testing on a mobile device if unable to via other tools: https://codesandbox.io/s/react-web-share-api-9nonm

Please let me know if this needs rebasing (not sure if we've updated prod to react-router v5 yet).


packages/openneuro-app/src/scripts/datalad/mutations/share.jsx
```diff
@@ -42,7 +42,7 @@ export const mergeNewPermission = (
   }
 }
 
-const ShareDataset = ({ datasetId, userEmail, metadata, done }) => (
+const ShareDatasetPermissions = ({ datasetId, userEmail, metadata, done }) => (
```
david-nishi: I think converting fully to ManageDatasetPermissions (and likewise for SHARE_DATASET) would help future readers of the code.

david-nishi (changes requested):
Looks great! Ready for sprint review after small lexical change.

packages/openneuro-app/src/scripts/datalad/mutations/share.jsx
exbotanical: Agreed; I made a few more changes that will help clarify this as 'managing permissions' and not 'sharing'. Thanks

david-nishi (approved):

PR #1518, opened 2020-02-08
set tasksCompleted type to array - fixes submit metadata

exbotanical:
This fix lets us keep `tasksCompleted` in the schema and eliminates the original error / status 400 returned by the `addMetadata` mutation. I checked mongo - the metadatas collection, specifically - to ensure that the fields were updating when I submitted the form. Last, I checked to ensure metadata fields are returning and updating as they should via the graphql playground. 

Let me know if you needed the `tasksCompleted` field removed from the graphql schema nevertheless and I'll find another solution.


nellh (approved):
This is failing tests due to some mismatched snapshots but is otherwise working.

PR #1522, opened 2020-02-11
Refactor mutation permissions and add dataset admin flag check

nellh:
This generalizes some of the resolver helpers we have for verifying permissions and includes the addition of a dataset admin check.


david-nishi (approved):
All works as expected

PR #1532, opened 2020-02-14
Local development docker compose optimization

nellh:
This should make the initial and repeat startup for development much faster.

Adds a new shared nodejs service the unifies a lot of the shared Node server configuration. The React app, server, and indexer now inherit this configuration. node_modules is now kept around between container restarts and Lerna verifies the correct set of packages is present on start.

The number of dataset workers is reduce to two. This is enough to test parallelism but saves half the memory and coordination time, several extra container starts.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,29 +5,59 @@ version: '2.3'
 volumes:
   project:
   yarn_cache:
+  node_modules:
+  webpack_cache:
 
 services:
-  # web app bundle build
-  app:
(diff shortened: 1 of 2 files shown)


david-nishi (approved):
Works great. Timed ~1 min from container shutdown to full reload, so about 5X faster. Thanks @nellh!

exbotanical (approved):
Easily 5x faster now. Works great on linux!

PR #1534, opened 2020-02-17
Remove redundant mapping fields and index summaries

### LINKED ISSUE #1523: Metadata resolvers should return validator defined fields
Looks like this data is missing at the API level, preventing the extracted fields from being extracted by the search indexer.

nellh:
Fixes #1523 to include the validator summary fields in search results.


nellh:
Testing this requires manually dropping the datasets index. I'm working on another PR to automatically alias the index correctly.

Issue #1528, opened 2020-02-13
Search indexer should exit with non-zero code


nellh:
If it fails to run, it does not get rescheduled because Kubernetes considers the task successful since the exit code is zero.

PR #1535, opened 2020-02-17
Use a 'datasets' index alias to allow for reindexing new mappings

nellh:
This swaps in the latest datasets index after each mapping change.


nellh:
This also includes a fix for #1528

exbotanical (approved):

PR #1538, opened 2020-02-18
Fix incorrect key name for search results pagination

nellh:
This fixes a crash when loading paginated search results due to the different root key name (search vs datasets).


david-nishi (changes requested):
Looks good, but needed to update node-sass to 4.13 to get through compilation (thanks @MatthewZito for figuring that out).

Update: to fix, run `docker-compose down -v` before building the containers. The error occurs because the containers are pointed at old volumes that don't match the updated packages. Once all branches have been updated to the new volumes this should not be an issue.

PR #1572, opened 2020-03-19
Switch to Python 3.7 and pipenv for datalad-service

nellh:
To sanely resolve the package versions, it was easiest to switch to Python 3.7 at the same time.

Based on https://github.com/OpenNeuroOrg/datalad-service/pull/113


nellh:
Looks like this was working locally because I had overridden the Celery version. Will make a few changes to support the newer release of Celery, it is rightly pointing out some poor usage of synchronous tasks in our code.

david-nishi (approved):
Dataset access, upload, download, and snapshots all looking good.

PR #1595, opened 2020-04-17
Update README.md

robertoostenveld:
point to the new location of the bids-validator, at bids-standard rather than INCF


effigies:
LGTM. Test failures seem unrelated.

PR #1603, opened 2020-05-03
Switch to path prefix based bulk deletion

nellh:
The original tree model was more flexible but we only really need to delete entire subtrees currently. This replaces the FileTree object with an array of paths which are recursively removed in each layer (React, server, backend service).


david-nishi (changes requested):
It looks like the cache isn't getting updated. The deletes are definitely working properly in datalad, but the client still sees those files on reload.

nellh:
> It looks like the cache isn't getting updated. The deletes are definitely working properly in datalad, but the client still sees those files on reload.

Can you make sure you're running all server and datalad-service changes? There's a POST request to handle the cache update from the dataset workers to the OpenNeuro server which is new and won't get loaded without the workers being restarted.

david-nishi (approved):
Successfully tested removal of root-level directory, sub-directory, and file.

PR #1613, opened 2020-05-14
Remove deleted datasets from search indexes.

nellh:
This cleans up datasets from search indexes on deletion, preventing search queries from linking into datasets which no longer exist.


david-nishi (approved):
Tested with public dataset removed by graphql mutation.

Issue #1555, opened 2020-02-21
Remove native file API origin trial tokens and provide feedback


nellh:
This is a reminder to clean up the tokens once they have expired and provide trial feedback.

The trial ends May 12th, 2020.

ckrountree:
@nellh can you check on this - the trial has not been extended (still May 12th) but now they have a native file api v2. Second trial for Chrome 83 (due out in May), ending September 29th. Chrome 82 will never be released.  Our "current" token has expired.  

- [x] Renew current token

See issue #1586 
- [x] Make edits before May (Chrome 83) release
- [x] Register for native file api v2.0
- [x] Register for next trial (Chrome 83) soon - before May

ckrountree:
@franklin-feingold do you have any feedback for Chrome downloads from users?

franklin-feingold:
I've started to see a subtly uptick in download support (users on Chrome) but this doesn't mean this product will not help serve downloads. I will take note of user experience downloading on Chrome as I get them

nellh:
@franklin-feingold  Note that the native file API downloader was automatically disabled on April 3rd when our existing token expired, so Chrome users may have seen a regression in the last week. It will be re-enabled with today's bugfix release until May 12th.

Chrome 83 will include some minor API changes based on the trial feedback and some bug fixes. We will aim to re-enable it whenever Chrome 83 is stable (maybe before the current trial ends on May 12th).

franklin-feingold:
Ahh all sounds good thank you! I'll look out for tickets after today's release for feedback

nellh:
I provided Chrome with some feedback, mostly that it would be good to have more indication of what files are being accessed than the permission grant in the navbar. V2 of the API does address some of the other issues we had (we had to use JavaScript heap for chunks of files which is slow, but V2 allows us to pass references instead).

PR #1615, opened 2020-05-15
Enable native file API v2 for downloads

nellh:
This updated the origin trial tokens to enable the v2 native file API with Chrome 83's release next week.

Requires Chrome 83 to test (beta channel currently).


nellh:
Also finalizes #1555 by removing the tokens.

PR #1622, opened 2020-05-28
[WIP] Helm charts for Rook / EdgeFS deployment

nellh:
This adds the initial configuration and documentation for our deployment of Rook and EdgeFS for dataset storage.


nellh:
This is WIP while I sort out the mounting target across charts now that it's been refactored into three separate helm charts. Each chart runs independently but we need to expose the NFS mount targets to dataset workers.

nellh:
Closing this as the approach we're taking has changed to no longer using EdgeFS.

Issue #1316, opened 2019-10-09
Fix deleting comments


franklin-feingold:
The delete button on comments is not working

sample use case - [ds002234](https://openneuro.org/datasets/ds002234)

ckrountree:
ds2234 is currently a private ds (pending publication). When running into this problem in the past, Nell manually deleted the comments. Basically, the button is not responsive upon click - and no error message provided - at both the ds Admin and site level. Franklin (as admin) ran into the same issues as the user reporting the issue.

nellh:
@franklin-feingold Do we want to allow dataset admins to remove comments on their datasets? Even if this issue was fixed, it seems like we may want some users to be able to moderate their dataset comments but the button as implemented is restricted to site admins.

franklin-feingold:
I think that sounds good - giving the dataset owner control over their dataset landing page

ckrountree:
@david-nishi let's make sure to write a test for the functionality on this one.

PR #1637, opened 2020-06-15
1316 delete comments

david-nishi:
#1316 
Major Changes
- DeleteComment component
- deleteComment gql mutation
  - has ability to delete dependencies (do we want this?)
- packages/openneuro-server/src/graphql/__tests__/comment.spec.js

### DIFF
--- a/packages/openneuro-server/src/graphql/__tests__/comment.spec.js
+++ b/packages/openneuro-server/src/graphql/__tests__/comment.spec.js
@@ -1,9 +1,19 @@
 import mockingoose from 'mockingoose'
-import { deleteComment } from '../resolvers/comment'
+import { deleteComment, flatten } from '../resolvers/comment'
 import Comment from '../../models/comment'
 
+const incrementHex = (hex, step = 1) => {
+  hex = hex.toString()
+  const splitLen = 14
+  const start = hex.substr(0, splitLen)
+  const end = hex.substr(splitLen)
+  let dec = parseInt(end, 16)
+  dec++
+  return start + dec.toString(16)
+}
+
 describe('comment resolver helpers', () => {
-  describe('for deleteComment', () => {
+  describe('deleteComment', () => {
     let aId
     let commentA, commentB, commentC
     beforeEach(async done => {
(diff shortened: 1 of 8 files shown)


packages/openneuro-server/src/graphql/resolvers/comment.js
```diff
@@ -18,6 +18,34 @@ const replies = obj => {
...
+ * @param {Comment} obj
+ * @returns {Promise<Comment[]>}
+ */
+const allNestedReplies = async obj => {
+  const replies = await Comment.find({ parentId: obj._id }).exec()
+  if (!replies.length) {
+    return replies
+  } else {
+    let nestedReplies = await Promise.all(replies.map(allNestedReplies))
+    const r = flatten([replies, ...nestedReplies])
```
nellh: nestedReplies can be const here and I'd just return the next line instead of creating a single character variable.

packages/openneuro-server/src/graphql/__tests__/comment.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      done()
+    })
+
+    it('returns an array of the deleted comment ids', async done => {
+      const deletedIds = (await deleteComment({}, { commentId: aId })).map(id =>
+        id.toString(),
+      )
+
+      // mockingoose ids are sequential, and the documents returned from each query get new ids for some reason
+      //   TODO: Replace mockingoose with better library
```
nellh, line 47: Yeah, it doesn't 100% do what we want. I haven't looked recently to see if there's better ways to mock out MongoDB with Mongoose. I think mocking it is preferred to running a real MongoDB for tests but we could consider that as well.

packages/openneuro-server/src/graphql/resolvers/comment.js
```diff
@@ -58,6 +86,23 @@ export const editComment = async (
...
+    targetComments.push(...(await allNestedReplies(existingComment)))
+  }
+  const deletedCommentIds = targetComments.map(c => c._id)
+  return Comment.deleteMany({
+    _id: {
+      $in: deletedCommentIds,
+    },
+  }).then(() => deletedCommentIds)
+}
+
```
nellh, line 108: This lets anyone delete any comment. It should be restricted to admins for now.

nellh (changes requested):
The missing permission check is the main issue here. Otherwise some linting problems to fix up.

packages/openneuro-server/src/graphql/__tests__/comment.spec.js
david-nishi: The odd thing is that we have [mongo-mock](https://www.npmjs.com/package/mongo-mock) in the test code already, but it's a bit sparse on features so mongoose was easier to implement in this case. [mongodb-memory-server](https://github.com/nodkz/mongodb-memory-server) looks promising, and it's already got good jest integration.

PR #1671, opened 2020-06-25
Fix Server Dev Builds

david-nishi:
# Major Changes
- bring back ./packages/openneuro-server as server service volume (not sure if this is the optimal solution with the recent service changes, feedback much appreciated)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     working_directory: ~/OpenNeuroOrg/openneuro
     docker:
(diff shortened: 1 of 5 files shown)


nellh (changes requested):
Looks like the src volumes did get missed in the #1617 

The goal with changing those was to only mount the actual dev environment source files, not the top level configuration like package.json since that's tied to the container build now. To do that this volume needs to not overwrite the entire server directory.

```
./packages/openneuro-server/src:/srv/packages/openneuro-server/src
```

The app container also needs this change, can you make it there as well?

nellh (approved):

PR #1694, opened 2020-06-29
Use git-annex whereis to obtain remote URL paths

nellh:
This replaces the error prone method of crawling the buckets for S3 versions to return special remote paths.

Fallback URLs are still provided but they should only be used for private datasets with this change.


david-nishi (approved):
## Tested
- view/download files (private/public, root/nested dir)
- add file & download
- snapshot dataset & download
- ran test suite

Issue #1692, opened 2020-06-29
`openneuro download` corrupts dataset on HTTP errors


kousu:
**Describe the bug**

Sometimes using the GraphQL API will return 504 Gateway Timeout. I *think* it tends to be more common with /download but all sorts of endpoints have it #1513. When this happens to `openneuro download`, the error pages are saved to disk in lieu of the expected data, and no one is the wiser.

**To Reproduce**
1. `openneuro download -s 57fecb0ccce88d000ac17538 ds000001 ds000001`
2. Wait for the download to complete
3. `grep -ri 504 ds000001/`

**Expected behavior**

The site usually works, but sometimes it doesn't! When it doesn't, `openneuro download` should catch these errors and report them instead of writing them into files.

**Screenshots**

```
$ grep -ri 504 ds000001/
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-03_events.tsv:<head><title>504 Gateway Time-out</title></head>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-03_events.tsv:<center><h1>504 Gateway Time-out</h1></center>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-02_bold.nii.gz:<head><title>504 Gateway Time-out</title></head>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-02_bold.nii.gz:<center><h1>504 Gateway Time-out</h1></center>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-01_events.tsv:<head><title>504 Gateway Time-out</title></head>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-01_events.tsv:<center><h1>504 Gateway Time-out</h1></center>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-01_bold.nii.gz:<head><title>504 Gateway Time-out</title></head>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-01_bold.nii.gz:<center><h1>504 Gateway Time-out</h1></center>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-02_events.tsv:<head><title>504 Gateway Time-out</title></head>
ds000001/sub-14/func/sub-14_task-balloonanalogrisktask_run-02_events.tsv:<center><h1>504 Gateway Time-out</h1></center>
[...]

Issue #1513, opened 2020-02-06
OpenNeuro-CLI: ServerParseError (504 errors)


franklin-feingold:
we are seeing a previously uncommon error becoming more prevalent and coming up across several users. Here is an example of one of the error readouts (they are all very similar if not exactly the same as this). The full thread of this evolving error can be found on [NeuroStars](https://neurostars.org/t/openneuro-cli-problem/5107/28)

```
SyntaxError [ServerParseError]: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at /Users/alfoi/.config/yarn/global/node_modules/apollo-link-http-common/lib/index.js:35:25
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  name: 'ServerParseError',
  response: Response {
    size: 0,
    timeout: 0,
    [Symbol(Body internals)]: { body: [PassThrough], disturbed: true, error: null },
    [Symbol(Response internals)]: {
      url: 'https://openneuro.org/crn/graphql',
      status: 504,
      statusText: 'Gateway Time-out',
      headers: [Headers],
      counter: 0
    }
  },
  statusCode: 504,
  bodyText: '<html>\r\n' +
    '<head><title>504 Gateway Time-out</title></head>\r\n' +
    '<body bgcolor="white">\r\n' +
    '<center><h1>504 Gateway Time-out</h1></center>\r\n' +
    '</body>\r\n' +
    '</html>\r\n'
}
```

kousu:
fwiw, I left more in-depth logs on #1692.

kousu:
I also want to ask that `openneuro-cli` learns to catch HTTP errors. In its current state, `openneuro download` will silently corrupt your datasets if the server is feeling under the weather.

ckrountree:
This should be fixed once we deploy the updated server side & cli version uploaders.

PR #1697, opened 2020-06-29
Prevent openneuro-cli from downloading all files at once

nellh:
Fixes openneuro-cli creating too many download requests at once. The superagent pipe here does not actually return a promise (just a thenable object), we need to wait for the finish event.


nellh:
This isn't the only issue but it should help with #1513 / #1692

nellh:
Closing this in favor of #1704

PR #1704, opened 2020-06-30
CLI: Rework downloads to use node-fetch instead of superagent

nellh:
This replaces #1697 and reworks the CLI download process to add error handling and drop the dependency on superagent.


david-nishi (approved):
## Tests
- download public dataset with cli
- download private dataset with cli
- run test suite

PR #1708, opened 2020-07-02
Handle unversioned S3 objects in parse_rmet_line

nellh:
This affects a few older datasets, a fallback URL is served instead but we should handle the unversioned case.


nellh:
ds002723 is an example of this issue.

david-nishi (approved):

rwblair:
Can/would we want the objects be accessed without the `?versionId={}` ?

PR #1728, opened 2020-07-10
Enable TypeScript support for openneuro-server

nellh:
This enables using TypeScript files within openneuro-server (same as openneuro-indexer already had).

One additional change upgrading ioredis and removing an unused redis package.


david-nishi (approved):
## Tested:
- log out/in
- navigate public/private dashboard
- dataset
  - upload
  - browse
  - comment
  - publish
  - update
  - snapshot
  - download

Issue #1723, opened 2020-07-06
Reduce Redis cache memory requirements


nellh:
**Describe the bug**
There's some pretty inefficient caching present across the API. With better filesystem performance available, we could take several strategies to reduce the cache size significantly. Expiring unreachable or rarely accessed keys, dropping easily recreated portions of the cached values. Most of our cache is JSON serialized backend responses, we could do away with the JSON step to save some encoding overhead as well.

**To Reproduce**
1. Crawl the snapshots for all datasets in production, this produces a total cache size above 4GB.

**Expected behavior**
4GB is probably not too unreasonable to support quick access to this many datasets but do we need to really cache that much? This mainly benefits burst traffic to specific datasets, so caching just those would save hosting resources.

**Additional context**
Recently we've been hitting the memory limit for Redis in production. Optimizations here will allow us to support more datasets in the same memory allocation.

PR #1729, opened 2020-07-10
Use compression for redis keys

nellh:
This should be a major improvement to #1723 

I consolidated the cache behavior into one interface called CacheItem. Part of the idea here is to add a more advanced cache controller that owns these but for now they directly replace a lot of the hand written cache behavior within the handlers.


nellh:
I missed one Redis cache usage that didn't match the pattern for the dataset description, updated that in a286b92.

david-nishi (approved):
## Tested:
- for old and new dataset
  - snapshot
  - browse files

PR #1733, opened 2020-07-14
Handle expiration in withProfile HOC

nellh:
This checks for expired tokens when rendering user profile dependent components, fixing a UI issue if your token expired instead of being removed at logout.


david-nishi (approved):
# Tested:
- refreshed page after jwt expiration; confirmed ui components at coherent login state

PR #1737, opened 2020-07-15
Fix missing readme content for snapshots

nellh:
This is a regression introduced in #1729

The issue is the readme and description resolvers can both be called from either a dataset or snapshot context. There was different logic for handling this, breaking the readme version. This PR consolidates it to a function and adds some unit tests for the shared function.


david-nishi (approved):
## Tested:
- updated README, snapshotted dataset, and confirmed changes in new snapshot

Issue #1443, opened 2020-01-07
Network error throws "This dataset does not exist"


nellh:
`Network error: JSON.parse: unexpected character at line 1 column 1 of the JSON data` should show something like "this request timed out, try again later" instead of the confusing message that your dataset does not exist.

This is a bug from #1405

nellh:
I believe this was specifically caused by a legitimate error thrown by the backend during snapshot creation but the UI reported dataset does not exist.

There's a general issue where errors are caught and reported as the wrong error by the existing handler in the error boundary for dataset pages.

PR #1740, opened 2020-07-20
1443 network error

david-nishi:
## Related Issue
- #1443 

## Major Changes
- DatasetRedirect extracted from ErrorBoundary and set in DatasetQuery
- ErrorBoundary no longer accepts error as a prop. Errors must be thrown internally.
- ErrorBoundary is no longer dependent on dataset state.
- GraphQL errors are relayed by ErrorBoundary 

## Minor Changes
- FreshdeskInterface extracted from ErrorBoundary
- sentry eventId now sent to freshdesk
- operations on non-existant datasets throw error “Dataset <datasetId> does not exist.”
- snapshots updated
- tests involving checkDatasetWrite() updated

## Maintained features
- Freshdesk modal
- Sentry exception captures
- Error handling for datasets with no snapshots
- GraphQL permissions errors
- Assertion failure exception
- Minor ErrorBoundaries


packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-query.jsx
```diff
@@ -64,6 +65,9 @@ export const datasetQueryDisplay = (isPublic, isSaved) => ({
   variables,
   error,
 }) => {
+  useEffect(() => {
+    if (error) throw error
+  }, [error])
```
nellh: This is basically the same as setting errorPolicy = 'none' in the Query component. If we can't render the partial result at all, I'd just change that and avoid useEffect throwing errors.

packages/openneuro-app/src/scripts/datalad/dataset/dataset-query.jsx
```diff
@@ -119,12 +119,19 @@ export const DatasetQueryHook = ({ datasetId, draft, history }) => {
...
   })
 
+  useEffect(() => {
+    if (error) {
+      throw error
+    } else if (data && data.dataset && data.dataset.snapshots.length === 0) {
+      throw Error('This dataset has no associated snapshots.')
+    }
+  }, [error, data])
+
```
nellh: This one is better since it does have some logic at least, but maybe we can move the boundary closer to where this actually causes issues and still allow some data to be shown?

packages/openneuro-app/src/scripts/errors/freshdeskInterface.jsx
```diff
@@ -0,0 +1,57 @@
...
+import {
+  Overlay,
+  ModalContainer,
+  ExitButton,
+} from '../../scripts/styles/support-modal.jsx'
+import PropTypes from 'prop-types'
+
+const FreshdeskInterface = props => {
+  const [showModal, setShowModal] = useState(false)
+  console.log(props)
```
nellh: One extra console.log here.

packages/openneuro-server/src/graphql/permissions.js
```diff
@@ -78,31 +83,39 @@ export const checkDatasetRead = (datasetId, userId, userInfo) => {
  * @param {object} state Level to verify
  */
 
-export const checkDatasetWrite = (
+export const checkDatasetWrite = async (
   datasetId,
   userId,
   userInfo,
   state = states.WRITE,
+  checkExists = true,
```
nellh: Since the signature is changing anyways, maybe we should make this an options object argument? That way the signature can remain the same in the future. Not sure, either way seems pretty good here.

nellh (changes requested):

packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-query.jsx
david-nishi: When I tried setting errorPolicy to 'none' the errors weren't caught in the appropriate error boundary. For example in the dataset/dataset-query, an undefined dataset would cause error when it's id property was accessed, rather than displaying the "dataset does not exist" error message from the server.

I've moved it to DatasetTab to fix the useEffect hook activation (doesn't always work in an HOC) and updated the error handling to display the results if datasets are returned with the query. Error boundary is activated with server's error message when datasets are undefined.

nellh: When it's set to none, all query errors will need to be handled by the child of the query component when it receives the error prop. Network errors we could handle with an error link instead, that might be a better approach than using React error boundaries to handle non-React issues? It's not clear to me if a query error is available in the link too, that could be a nice way to catch both cases generally.

nellh (changes requested):
I'm still running into some issues with this other than the linting issue reported by the last build.

```
Invariant Violation: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

A quick way to reproduce it is to clear the cache for a dataset and stop the datalad service, then load the dataset.

PR #1747, opened 2020-07-22
Refactor to remove Celery

nellh:
This drops the Celery task queue, eliminating a decent amount of complexity from the datalad service component. Two long running tasks, publish and validation are now handled in greenlets which do not block requests but do run under the same uwsgi worker process.

This PR excludes the related changes to the Helm configuration, but includes all local dev changes.


nellh:
We could probably get some better performance making more things into gevent greenlets but I think we will probably want to switch away from gevent and wsgi with Falcon 3.0 and use asyncio instead. That means keeping the explicit gevent usage to the minimum for now.

david-nishi (approved):
## Tested (with node version 12.18.1 in DockerFile):
- upload dataset
- view draft
- update description
- view files
- update file
- view snapshot
- snapshot dataset

Issue #1752, opened 2020-07-29
Public dataset without a snapshot


franklin-feingold:
**Describe the bug**
A dataset owner is able to publish their dataset without a snapshot. This may make the dataset public, but it is not accessible by users. 

**To Reproduce**
1. Go to https://openneuro.org/datasets/ds003037 (8/3 update: another dataset is experiencing this same issue: https://openneuro.org/datasets/ds002936)
2. The error `This dataset has no associated snapshots. Please support us by documenting this issue with FreshDesk.`

**Expected behavior**
The desired behavior is the dataset cannot be made public without a snapshot.

**Desktop (please complete the following information):**
- OS: iOS
 - Browser: Chrome
 - Version: 83

nellh:
This failed because DOI minting failed but looks like we don't have a good error report for why. I filed #1753 to capture the exception so we can debug further.

nellh:
Narrowed this down to HTTP 422 unprocessable entity from DataCite. This means the DOI minting request failed due to the POST body failing to validate against their API schema. This might be a bug in how we're populating these fields but with other datasets working it's not immediately obvious why these two are different (the fields submitted to DataCite look correct).

rwblair:
Fortunately the DataCite schema hasn't changed recently. They aren't overly clear in the schema documenation about limitations of data values, interestingly in this properties table there is caveat about transliterating non roman creator names but I'd be surprised if it threw 422s for that:
http://schema.datacite.org/meta/kernel-4.3/doc/DataCite-MetadataKernel_v4.3.pdf#%5B%7B%22num%22%3A30%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C68%2C655%2C0%5D

nellh:
Narrowed this down more, it's that dataset_description.json is empty in the cache for the commit, despite the actual underlying git object containing data. This is some kind of cache state bug.

Workaround is dropping the cache and then creating a snapshot.

PR #1753, opened 2020-07-30
Capture errors when DOI minting fails

nellh:
This should help catch issues like #1752


david-nishi (approved):
## Tested:
- set up rejection with error after oldDesc declaration
- create snapshot
- confirm error "#1753 test" is thrown on sentry

Issue #1662, opened 2018-12-05
access to data in many datasets via git-annex/datalad doesn't work


yarikoptic:
I don't know what is the reason yet. But, I was trying to get all the data so I could extract metadata, but saw that in many datasets `get` fails for me. So this bash snippet tries to get 2 random files from each dataset.
```shell
$> for ds in ds0*; do  builtin cd $ds; sample_files=$(/usr/bin/find -type l -iname \*nii.gz | shuf | head -n 2); echo ${sample_files} | xargs datalad get > .git/log-sample-get 2>&1 && echo "$ds get2 ok" || echo "$ds get2 failed"; datalad drop --nocheck $sample_files &>/dev/null; builtin cd ..; done
ds000001 get2 ok
ds000002 get2 ok
ds000003 get2 ok
ds000005 get2 ok
ds000008 get2 ok
ds000011 get2 ok
ds000017 get2 ok
ds000051 get2 ok
ds000052 get2 ok
ds000101 get2 ok
ds000102 get2 ok
ds000105 get2 ok
ds000107 get2 ok
ds000109 get2 ok
ds000114 get2 failed
ds000116 get2 failed
ds000117 get2 failed
ds000120 get2 failed
ds000122 get2 ok
ds000138 get2 failed
ds000144 get2 failed
ds000157 get2 failed
ds000164 get2 failed
ds000168 get2 failed
ds000174 get2 failed
ds000200 get2 failed
ds000201 get2 failed
ds000202 get2 failed
ds000205 get2 failed
ds000208 get2 failed
ds000213 get2 failed
ds000214 get2 failed
ds000218 get2 failed
ds000219 get2 failed
ds000222 get2 failed
ds000228 get2 failed
...
```
and so on -- some later seems to work out ok but not the other. git annex is `6.20181011+git124-g94aa0e2f6-1~ndall+1` (the same with `7.20181121+git58-gbc4aa3f0e-1~ndall+1`).
Here is more details on the failing one (hopefully @joeyh would have a chance to look -- http://github.com/OpenNeuroDatasets/ds001545), where they all (?) fail like this (unfortunately --debug is not sufficient):
```bash
(git)smaug:/mnt/btrfs/scrap/datalad/openneuro-slim/ds001545[master]git
$> git annex get --debug --from=s3-PUBLIC sub-01/anat/sub-01_T1w.nii.gz
[...]

PR #1760, opened 2020-08-11
Add script to delete all versioned objects in an S3 bucket prefix

nellh:
This is part of the fix for #1662 

To successfully recreate these exports in the same place we have to fully delete them on the S3 side.

### DIFF
--- a/scripts/s3-delete-all-versions.py
+++ b/scripts/s3-delete-all-versions.py
@@ -1,0 +1,48 @@
+import argparse
+
+import boto3
+
+
+def nuke_prefix():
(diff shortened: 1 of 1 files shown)


david-nishi (approved):
## Tested by:
- upload dataset and make updates & snapshots
- verify in s3 bucket
- run script on uploaded dataset in bucket and confirm deletion on s3

rwblair:
Would an empty string like `''` nuke the entire bucket or is argparse picky enough that if it gets a required argument but its null/empty it will complain. Guess '/' would kill everything as well.

PR #1763, opened 2020-08-11
Add an endpoint field to prepareUpload to coordinate ingress -> worker mapping

nellh:
This simplifies the routing needed to forward file upload requests to the correct backend. See #1762 for the Kubernetes half of this.


david-nishi (approved):
## Tested:
- upload dataset from browser and client
- view dataset on browser

PR #1771, opened 2020-08-14
bug fix for issue 1712 where scroll is reset on FF when typing in com…

thinknoack:


nellh (approved):
Tested this in Firefox on desktop and mobile, working great!

Issue #1358, opened 2019-10-23
Implement background fetch for uploads


nellh:
Rework the current uploader implementation to support using the background fetch API if available.

ckrountree:
Background fetch has been implemented but it is not being used because it didn't work well in Chrome 85 (current release of Chrome).
Background fetch is ~75% of the implementation. So, with the current implementation, it wouldn't show any upload progress but turning it on should successfully upload once Chrome fixes it.

nellh:
No Firefox support still, I think this may not be useful.

PR #1787, opened 2020-08-28
Port React uploader to the new upload API

nellh:
This consolidates a lot of the uploader mechanics to use the same implementation across the frontend and the CLI, built around Fetch instead of Apollo Upload. With HTTP/2 on the server, this is a decent performance boost as well since that is usable with the browser implementation of Fetch transparently.

A very basic version of displaying the state for individual files within the upload is implemented.

Background fetch (#1358) is disabled due to instability but I left some of the hooks for it in place so that we can easily re-enable it when it has better browser support (hopefully soon for Chrome).


david-nishi (approved):
## Tested:
- Upload new dataset from browser
- Upload dataset with modifications from browser
- Refresh partway through dataset upload and resume upload from dataset page

Issue #1665, opened 2019-10-10
Don't depend on DataLad repo class APIs


nellh:
From https://github.com/OpenNeuroOrg/datalad-service/pull/114#issuecomment-540262800

- [ ] repo.get_tags()
- [ ] repo.get_hexsha()
- [ ] repo.init_remote() for S3 remotes
- [ ] repo.dirty
- [x] repo.repo.commit(_tag_).hexsha
- [x] repo.repo.delete_tag(_tag_)
- [x] `git show <tree>:<file>` which we use to read a file in a past revision

ckrountree:
Release 13 came out on June 23, 2020.

nellh:
Closing this as we do still use the repo methods implemented by DataLad but all the GitPy API calls are gone.

PR #1788, opened 2020-08-28
Update to datalad 0.13.3

nellh:
This fixes a few of our dependencies on DataLad 0.12 specific APIs and resolves some use of non-public APIs. Updates to git-annex 8.20200618 as well since this release of DataLad is compatible with it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
           working_directory: ~/project/services/datalad
 
       # install git-annex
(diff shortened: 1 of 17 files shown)


nellh:
This should also fix #1665 - not every API is removed but this limits our use to things that are fairly stable in the AnnexRepo and GitRepo classes within DataLad.

david-nishi (approved):
## Tested:
- upload dataset (from browser)

PR #1809, opened 2020-09-14
Refactor direct file API to share implementation with most other file access

nellh:
This removes the corner case used when a file is directly requested. This should avoid some timeouts for files that are not available on S3 yet by skipping the costly cache miss for some situations with `await getFiles(datasetId, currentRevision)`.


david-nishi (approved):
## Tested:
- view dataset files and contents

ckrountree:
### Sprint review
File contents now exist in ds002290.  Files that weren't annexed would previously hang but not any longer.

PR #1814, opened 2020-09-18
Switch to explicit fetch imports and fetch-h2 for openneuro-cli

nellh:
This fixes an issue where the wrong fetch implementation was used in some Node.js environments, resulting in the ReadableStream not being consumed. Instead of relying on cross-fetch, we import fetch-h2 directly and call that.

Also cleans up a few debugging statements that had slipped through in the previous release and makes sure to consume the body for completed upload requests because fetch-h2 requires this to avoid hanging h1 sockets when an http2 server is unavailable (docker-compose setup).

### DIFF
--- a/packages/openneuro-cli/package.json
+++ b/packages/openneuro-cli/package.json
@@ -22,6 +22,7 @@
     "commander": "^2.15.1",
     "cross-fetch": "^3.0.5",
     "esm": "^3.0.16",
+    "fetch-h2": "^2.5.1",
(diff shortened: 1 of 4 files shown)


david-nishi (approved):
## Tested
- download with cli
- download from browser

ckrountree:
### Sprint review:
Saw a successful upload with the branch tested (this branch) - ds002290.  Files (including nifti) are all full and available.

PR #1815, opened 2020-09-21
1076 refetch state

### LINKED ISSUE #1076: Handle refetch state after dataset mutations
Mutations (mainly dataset edits) currently track only part of the lifecycle for showing progress.

1. Prepare mutation
2. Mutation request
3. Mutation complete
4. Refetch request
5. Refetch complete

The loading indicator goes away after part 3, mutation complete. This is misleading for users as it indicates nothing else will happen but shortly after the page data refreshes. We should show some indication of loading during the refetch if one is required since the refetch could take a significant amount of time (as much as a minute for large datasets on low bandwidth connections) or fail in some rare cases. 

#1072 is an attempt at this but the straightforward networkState approach is currently not implemented in Apollo Client, the subscribed component will never be notified of the refetch state. We need some additional state handling to know when the refetch is in progress at time of render for the datasets page.

david-nishi-dev:
## Major changes:
- resolves #1076

## Minor changes:
- fix delete directory
- update apollo cache on dataset description updates

### DIFF
--- a/packages/openneuro-app/src/scripts/datalad/mutations/__tests__/__snapshots__/delete.spec.jsx.snap
+++ b/packages/openneuro-app/src/scripts/datalad/mutations/__tests__/__snapshots__/delete.spec.jsx.snap
@@ -84,6 +84,7 @@ exports[`DeleteDataset mutation renders with common props 1`] = `
 
 exports[`DeleteDir mutation renders with common props 1`] = `
 <Mutation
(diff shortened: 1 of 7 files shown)


nellh (approved):
This is working well for me. There's still some issues with delete where the cache does not seem to update after delete, but the status is updated correctly for the mutation now which covers the original issue.

PR #1816, opened 2020-09-21
Delete dataset improvements

### LINKED ISSUE #1577: Add to dataset deletion process
An enhancement to our current dataset deletion processes. Documented additional processes from https://github.com/OpenNeuroOrg/openneuro/issues/1574

The additions are to: 
- [x] remove dataset from [GitHub](https://github.com/OpenNeuroDatasets)
- [x] remove dataset from S3
- [ ] document list for Yarik to remove index from datalad (look at #1259 first)

david-nishi:
resolves #1577 and #1259 

# Major Changes
- Users deleting a dataset are prompted for a reason ("duplicate dataset", "abuse of service", or "Other" with custom input) for deletion and "Superseded by (URL)".
- Delete info is stored in mongo DB, including fields: datasetId, reason, redirect, user(_id), and a timestamp.
- When users delete a dataset, that dataset's siblings on github and S3 (Private and Public) are removed as well.

### DIFF
--- a/services/datalad/datalad_service/handlers/dataset.py
+++ b/services/datalad/datalad_service/handlers/dataset.py
@@ -45,7 +45,7 @@ class DatasetResource(object):
 
     def on_delete(self, req, resp, dataset):
         try:
-            delete_siblings(dataset)
+            delete_siblings(self.store, dataset)
             delete_dataset(self.store, dataset)
             resp.media = {}
             resp.status = falcon.HTTP_OK
--- a/services/datalad/datalad_service/handlers/dataset.py
+++ b/services/datalad/datalad_service/handlers/dataset.py
@@ -45,10 +45,16 @@ class DatasetResource(object):
 
     def on_delete(self, req, resp, dataset):
         try:
+            print('DELETE DATASET *****')
             delete_siblings(self.store, dataset)
+            print('SIBLINGS DONE *****')
             delete_dataset(self.store, dataset)
+            print('DATASET DONE *****')
             resp.media = {}
             resp.status = falcon.HTTP_OK
-        except:
+        except Exception as e:
+            print('<<<<<')
+            print(e)
+            print('<<<<<')
             resp.media = {'error': 'dataset not found'}
             resp.status = falcon.HTTP_NOT_FOUND
--- a/services/datalad/datalad_service/handlers/dataset.py
+++ b/services/datalad/datalad_service/handlers/dataset.py
@@ -5,7 +5,7 @@ import falcon
 from datalad_service.common.user import get_user_info
 from datalad_service.tasks.dataset import create_dataset
 from datalad_service.tasks.dataset import delete_dataset
-
+from datalad_service.tasks.publish import delete_siblings
 
 class DatasetResource(object):
 
@@ -45,6 +45,7 @@ class DatasetResource(object):
 
     def on_delete(self, req, resp, dataset):
         try:
+            delete_siblings(dataset)
             delete_dataset(self.store, dataset)
             resp.media = {}
             resp.status = falcon.HTTP_OK
--- a/services/datalad/datalad_service/handlers/dataset.py
+++ b/services/datalad/datalad_service/handlers/dataset.py
@@ -45,16 +45,10 @@ class DatasetResource(object):
 
     def on_delete(self, req, resp, dataset):
         try:
-            print('DELETE DATASET *****')
             delete_siblings(self.store, dataset)
-            print('SIBLINGS DONE *****')
             delete_dataset(self.store, dataset)
-            print('DATASET DONE *****')
             resp.media = {}
             resp.status = falcon.HTTP_OK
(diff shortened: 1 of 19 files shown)


nellh:
One test is currently failing but just a snapshot so I am starting the review of this.

ckrountree:
### Sprint Review
Created a test ds001048. Present on ON public, Github (sibling remote) repo, and S3 private (covers #1577). We have to select reason for deleting and add a redirect url  (covers #1259). Then submitted deletion request. It was missing/deleted from all remotes and also on site (message warning that the ds no longer existing.  MongoDB logged it with all pertinent info (e.g. id), including reasons for deleting and redirect.

services/datalad/datalad_service/handlers/dataset.py
```diff
@@ -45,9 +45,16 @@ def on_post(self, req, resp, dataset):
...
+            print('SIBLINGS DONE *****')
             delete_dataset(self.store, dataset)
+            print('DATASET DONE *****')
             resp.media = {}
             resp.status = falcon.HTTP_OK
-        except:
+        except Exception as e:
+            print('<<<<<')
+            print(e)
+            print('<<<<<')
```
nellh: Some debug code to clean up here.

services/datalad/datalad_service/tasks/publish.py
```diff
@@ -145,6 +149,43 @@ def publish_github_async(store, dataset, snapshot, github_remote):
...
     publish_target(ds, github_remote, snapshot)
 
+def delete_s3_sibling(dataset_id, siblings, realm):
+    sibling = get_sibling_by_name(realm.s3_remote, siblings)
+    if sibling:
+        try:
+            session = boto3.Session(
+                aws_access_key_id=AWS_ACCESS_KEY_ID,
+                aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
+            )
```
nellh: We'll need to make sure OpenNeuro's service accounts can do this, at the moment they cannot.

nellh: Actually I'm not sure that we can grant the credentials to the same account, this seems to be an S3 limitation. We probably need to have a new set of environment variables for a second set of credentials.

services/datalad/datalad_service/tasks/publish.py
```diff
@@ -145,6 +149,43 @@ def publish_github_async(store, dataset, snapshot, github_remote):
...
+            s3 = session.resource('s3')
+            bucket = s3.Bucket(realm.s3_bucket)
+            bucket.objects.filter(Prefix=f'{dataset_id}/').delete()
```
nellh: It looks like this does not delete the versions, only adds a delete marker. That wouldn't make old versions unavailable, I think this delete step needs to be replaced with one which crawls all versions of objects within the prefix and issues a delete for each version.

services/datalad/datalad_service/tasks/publish.py
```diff
@@ -145,6 +149,43 @@ def publish_github_async(store, dataset, snapshot, github_remote):
...
+        r.delete()
+    except StopIteration:
+        raise Exception(f'Attempt to delete dataset {dataset_id} from GitHub has failed, because the dataset does not exist.')
+
+def delete_siblings(store, dataset_id):
+    ds = store.get_dataset(dataset_id)
+
+    siblings = ds.siblings()
+    delete_s3_sibling(dataset_id, siblings, DatasetRealm(1))
+    delete_s3_sibling(dataset_id, siblings, DatasetRealm(2))
```
nellh: These should use the enum names in case they are reordered or additional remote realms are defined later.

nellh (changes requested):

services/datalad/datalad_service/tasks/publish.py
david-nishi: Went with client instead

services/datalad/datalad_service/tasks/publish.py
```diff
@@ -145,6 +149,55 @@ def publish_github_async(store, dataset, snapshot, github_remote):
...
+    siblings = ds.siblings()
+    delete_s3_sibling(dataset_id, siblings, DatasetRealm.PRIVATE)
+    delete_s3_sibling(dataset_id, siblings, DatasetRealm.PUBLIC)
+
+    remotes = ds.repo.get_remotes()
+    if DatasetRealm.PUBLIC.github_remote in remotes:
+        delete_github_sibling(dataset_id)
+
+    for remote in remotes:
+        ds.siblings('remove', remote)
```
nellh, line 200: I think we should also run this with gevent, that will prevent this call from timing out for larger datasets.

You can see an [example of that here](https://github.com/OpenNeuroOrg/openneuro/blob/d2ba4d39b5ea5417a28f23b024a467d9d2955602/services/datalad/datalad_service/handlers/publish.py#L18).

nellh (changes requested):

PR #1824, opened 2020-09-24
Avoid passing entire file list for prepareUpload to server

nellh:
Instead we hash it on the client or allow API clients to provide their own id to identify a particular upload.


david-nishi (approved):
## Tested:
- uploaded dataset from browser and cli
- verified contents

## Notes:
- Separate bug: dataset.draft.partial is sometimes not set to false when dataset is finished uploading

PR #1826, opened 2020-09-28
Chrome 86 native file API updates

nellh:
This makes the updates required to enable native file downloads for Chrome 86 or later now that the origin trial has concluded.


david-nishi (approved):
## Tested:
- uploaded datasets & verified contents
- failing test is due to 1827; is resolved there

PR #1827, opened 2020-09-28
Namespace uploadId hashes to the dataset to prevent duplicate uploadIds.

nellh:
Bug fix for missing content on the second upload of the same set of files. Regression introduced in #1824


david-nishi (approved):
## Tested:
- uploaded datasets & confirmed data

PR #1831, opened 2020-10-01
openneuro-cli: Raise minimum Node.js version to 12.9.0 account for current API usage

nellh:
When we updated our Node.js target to the current LTS release the engine field in openneuro-cli was missed. 12.9.0 is the minimum version with all of the required APIs.


david-nishi (approved):
## Tested
- rebuild containers

PR #1832, opened 2020-10-05
Apollo upload cleanup

nellh:
This removes the "partial" upload state from datasets and cleans up all APIs for the Apollo upload libraries in favor of the new uploader. Additionally it ties add file or add directory into the React uploader state, showing a progress bar for large uploads and preventing conflicts where multiple additions are in progress at once.

One known limitation is this prevents the final files from appearing at the moment since this behaviour is now shared with the regular uploader. I think we should address this for both cases but it is not handled here.


david-nishi (approved):
## Tested:
- in browser and cli
  - upload dataset
  - check dataset contents
- in browser
  - add & remove file(s)
  - update README

Issue #1834, opened 2020-10-05
Search results missing


franklin-feingold:
**Describe the bug**
It appears our indexer may not be uptaking new information. This appears to be affecting datasets made public in the past month. 

**To Reproduce**
1. Go to https://openneuro.org/public/datasets
2. Search a recent dataset by name
3. Dataset is not returned

**Expected behavior**
Dataset is returned

nellh:
A fix for this was released in 3.22.1

PR #1835, opened 2020-10-06
Fix various typescript compile issues in openneuro-client

nellh:
This is the underlying cause of #1834 - since we don't build this until runtime for openneuro-indexer it doesn't throw an error until it executes. We should do the build ahead of time to catch this.


david-nishi (approved):
## Tested:
- confirm new dataset is searchable after running indexer

PR #1839, opened 2020-10-09
openneuro-cli: Avoid untracked flag while preparing uploads.

nellh:
This drops untracked: true from the request used to diff the remote content with the local content. This prevents an issue with the size of this request getting dropped for very large datasets. This is no longer necessary with the current uploader, it was a requirement previously to resume in-progress uploads.


david-nishi (approved):
## Tested:
- view dataset in browser
- download dataset

PR #1840, opened 2020-10-09
Fix deprecated partial flag in search results query

nellh:
This fixes a 400 error for search results introduced in 3.22.0


david-nishi (approved):
## Tested:
- searched datasets

Issue #1619, opened 2020-05-19
Automate docker-compose node_modules updates


nellh:
The fix in #1617 for macOS means you need to remember to run docker-compose build after each node_modules change. We should catch that case and automatically rebuild it.

PR #1844, opened 2020-10-16
Add a root level script to rebuild dependencies

nellh:
Looking for ways to really solve #1619 doesn't seem possible given the docker-compose tooling available. Instead I think we should update the recommended way to the start docker-compose to add the `--build -V` flags which will always result in consistent dependencies on start. This makes the `docker-compose up` method of starting more of an advanced use case.

This PR adds `yarn start` and documents it in the README.


david-nishi (approved):
## Tested:
- ran `yarn start`
- checked basic site functions (search, login)
- installed package & ran `yarn start` again

Issue #1797, opened 2020-09-04
openneuro-cli allowing upload of disallowed files


nellh:
**Describe the bug**
.DS_Store is an example, previously this was removed by the client but I think we should apply the same list of ignored paths to the solve this at the API level.

**To Reproduce**
1. Add .DS_Store to your dataset
2. Upload the dataset with openneuro-cli

**Expected behavior**
This file and a few others should always be removed from the resulting dataset.

PR #1847, opened 2020-10-19
Skip invalid filenames during upload

### LINKED ISSUE #1656: Filter all dot files except .bidsignore from datasets on upload
**Describe the bug**
[ds002898](https://openneuro.org/datasets/ds002898/versions/1.0.0) does not have the imaging files. They are not present on S3. There may be a bug with git-annex.

**To Reproduce**
1. Go to [ds002898](https://openneuro.org/datasets/ds002898/versions/1.0.0)
2. View the imaging files - only the sidecars and metadata (non-annexed files) are present

**Expected behavior**
The imaging files being available. When the user tries to reupload the imaging files, the cli will skip those files.

nellh:
This ignores dotfiles, .git, etc at the API level. For API clients, we return a flag "skipped" in the JSON response for any files that were skipped to allow the client to report this back at the end of the upload.

Fixes #1656 and #1797

### DIFF
--- a/services/datalad/datalad_service/handlers/upload_file.py
+++ b/services/datalad/datalad_service/handlers/upload_file.py
@@ -1,10 +1,19 @@
-import os
 import logging
+import os
+import re
 
 import falcon
 
 from datalad_service.common.user import get_user_info
 from datalad_service.handlers.upload import UploadResource
+
+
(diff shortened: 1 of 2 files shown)


david-nishi (approved):
## Tested:
- ran test suite

Issue #1610, opened 2020-05-11
Add radio button for user verification of defacing


franklin-feingold:
The button will be verifying that the user has permission to share images with faces or they are verifying the images are defaced. 

This can be added to step 4 of the upload process - terms of use page

effigies:
Here is a quick draft. Feel free to adjust.

# Current text

#### By uploading this dataset to OpenNeuro I agree to the following conditions:

I am the owner of this dataset and have any necessary ethics permissions to share the data publicly. This dataset does not include any identifiable personal health information as defined by the Health Insurance Portability and Accountability Act of 1996 (including names, zip codes, dates of birth, acquisition dates, facial features on structural scans etc.). I agree to destroy any key linking the personal identity of research participants to the subject codes used in the dataset.

I agree that this dataset and results of all analyses performed on it using the OpenNeuro platform will become publicly available under a Creative Commons CC0 or Creative Commons CC-BY license (depending on the dataset metadata) after a grace period of 36 months counted from first successful analysis of data from more than one participant. You will be able to apply for up to two 6 month extensions to increase the grace period in case the publication of a corresponding paper takes longer than expected. See FAQ for details.

Please refrain from uploading datasets already publicly available in other repositories.

This dataset is not subject to GDPR protections.

# Suggested text

#### By uploading this dataset to OpenNeuro I agree to the following conditions:

I am the owner of this dataset and have any necessary ethics permissions to share the data publicly. This dataset does not include any identifiable personal health information as defined by the Health Insurance Portability and Accountability Act of 1996 (including names, zip codes, dates of birth, acquisition dates, etc.). I agree to destroy any key linking the personal identity of research participants to the subject codes used in the dataset.

I agree that this dataset and results of all analyses performed on it using the OpenNeuro platform will become publicly available under a Creative Commons CC0 or Creative Commons CC-BY license (depending on the dataset metadata) after a grace period of 36 months counted from first successful analysis of data from more than one participant. You will be able to apply for up to two 6 month extensions to increase the grace period in case the publication of a corresponding paper takes longer than expected. See FAQ for details.

Please refrain from uploading datasets already publicly available in other repositories.

This dataset is not subject to GDPR protections.

Please affirm one of the following:

* [ ] All structural scans have been defaced, obscuring any tissue on or near the face that could potentially be used to reconstruct the facial structure.
* [ ] I have explicit participant consent and ethical authorization to publish structural scans without defacing.

# Diff

```Diff
[...]

PR #1854, opened 2020-10-20
Add checkboxes for disclaimer step during uploads

nellh:
Implements #1610 with an integration test for the behavior requiring one box to be checked to proceed.

Screenshot from 2020-10-19 21-55-27

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
   },
   "devDependencies": {
     "@sentry/cli": "1.37.4",
-    "@types/jasmine": "^2.8.8",
     "@typescript-eslint/eslint-plugin": "^2.20.0",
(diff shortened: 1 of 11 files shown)


nellh:
Looking into the test failure here, seems to be a TypeScript setup issue rather than an issue with this PR.

david-nishi (approved):
## Tested:
- confirmed dataset upload prevented until a checkbox is clicked

nellh:
@david-nishi I've added some additional changes to this to record the state for the checkboxes as dataset metadata fields during upload.

david-nishi (approved):
## Tested:
- submitted datasets with one and both checkboxes selected; confirmed expected results in database

PR #1856, opened 2020-10-20
openneuro-cli: Correctly handle the exit path if the user aborts the upload

nellh:
If you start an upload with `openneuro upload ...` but choose to not upload by answering no to the prompt, it would ignore you and continue uploading anyways. This fixes that bug so aborting an upload has the correct behavior and sensible messaging.


david-nishi (approved):
## Tested:
- upload large dataset with cli and kill process partway through
- resume upload

PR #1858, opened 2020-10-20
Improve finishUpload API performance with rename over copy operations

nellh:
Improve performance for finishUpload API calls. This should help with datasets that contain larger files by using renames over copies.

This also fixes a bug where dot prefixed files were being always skipped (only affects .bidsignore but this is why bidsignore might be missing in some uploads).


david-nishi (approved):
## Tested:
- upload dataset with cli and browser

Issue #1646, opened 2020-06-18
Admin status for individual datasets


nellh:
**Is your feature request related to a problem? Please describe.**
There isn't much transparency in if your dataset is done exporting to S3, is stuck for some reason, etc.

**Describe the solution you'd like**
We could improve that by just adding some status info that site or dataset admins could see on the dataset page. For dataset admins, showing the export status and some repo metadata would help. For site admins, showing the storage pools assignments and real disk utilization would be my request.

PR #1865, opened 2020-10-30
Add datalad-service API for dataset history

nellh:
Adds an internal API that returns a JSON log with one line per commit for the dataset draft HEAD.

This is a dependency for #1646


ckrountree:
### SPRINT REVIEW
added a real dataset query and saw a git log of history.

david-nishi (approved):
## Tested:
- w/ test suite

PR #1867, opened 2020-10-30
Add cache clear mutation for admin dashboards

nellh:
Adds a mutation to allow admin users to clear our cache for a given dataset.

This is a dependency for #1646


ckrountree:
### sprint review
Checked caching = true, and then false, and then true again.

david-nishi (approved):
## Tested:
- ran mutation
- checked key absence in container from redis-cli

PR #1868, opened 2020-11-02
Add download progress toast to the native file web downloader

nellh:
This is an alternative solution to the issue in #1356 due to the disk space limitations with actually using background fetch for these operations. Instead this implements a download toast that shows progress per file and allows you to cancel downloads in progress, giving most of the same benefits (no downloading without the main thread though, so the page must stay open).

### DIFF
--- a/packages/openneuro-app/package.json
+++ b/packages/openneuro-app/package.json
@@ -55,7 +55,7 @@
     "react-router": "^5.1.2",
     "react-router-dom": "^5.1.2",
     "react-spring": "^8.0.27",
-    "react-toastify": "^4.4.0",
+    "react-toastify": "6.0.9",
     "react-virtualized": "^9.21.0",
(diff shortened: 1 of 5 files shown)


ckrountree:
### SPRINT REVIEW
Could see the download in progress, complete, and cancel:
Screen Shot 2020-11-03 at 10 06 09 AM


Can even download two at a time:
Screen Shot 2020-11-03 at 10 07 02 AM

david-nishi (changes requested):
On ds*/download page, "Download Error" toast is displayed if download is canceled at directory selection step.

For a future issue, I think it would improve UX if we added a step between hitting "DOWNLOAD" and choosing a target directory that prompts the user with a message like "Please choose a folder to download this dataset to."

## Tested: 
- downloaded dataset from browser

david-nishi (approved):

PR #1873, opened 2020-11-03
Fix a regression in resumeDataset with an incorrect query name

nellh:
Fixes a regression with "add files" or "add directory" in the dataset file tree throwing invariant violation.


david-nishi (approved):
## Tested:
- upload dataset & abort partway, then resume upload
- confirm uploaded dataset files

PR #1877, opened 2020-11-04
Add cacheClear event after publish to S3 to expose new URLs

nellh:
This adds an extra cache drop after each S3 publish to make sure we always have the new URLs for a given snapshot.


david-nishi (approved):
## Tested:
- snapshot and publish dataset and view in public dashboard
- navigate to dataset page from url bar

PR #1880, opened 2020-11-07
Prevent erroneous nesting when adding files to existing directories

nellh:
This fixes an issue with the finishUpload logic when adding data to an existing directory that would cause the directory to be incorrectly nested.

Adds two test cases for this:
1. The working behavior where the directory does not exist
2. The failing behavior where the directory does exist

### DIFF
--- a/services/datalad/datalad_service/handlers/upload.py
+++ b/services/datalad/datalad_service/handlers/upload.py
@@ -13,7 +13,6 @@ from datalad_service.common.annex import CommitInfo
 
 
 def move_files(upload_path, dataset_path):
-    print(dataset_path)
     for filename in pathlib.Path(upload_path).glob('**/*'):
         if os.path.isfile(filename):
             target = os.path.join(dataset_path, os.path.relpath(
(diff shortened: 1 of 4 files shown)


services/datalad/datalad_service/handlers/upload.py
```diff
@@ -12,6 +12,16 @@
 from datalad_service.common.annex import CommitInfo
 
 
+def move_files(upload_path, dataset_path):
+    print(dataset_path)
+    for filename in pathlib.Path(upload_path).glob('**/*'):
+        if os.path.isfile(filename):
+            target = os.path.join(dataset_path, os.path.relpath(
+                filename, start=upload_path))
+            os.makedirs(target, exist_ok=True)
```
effigies: I think you want to create the parent dir?

```suggestion
            pathlib.Path(target).parent.mkdir(parents=True, exist_ok=True)
```

nellh: Yep, thanks. I fixed the tests to check that these are files and not just directories with the right names.

david-nishi (changes requested):
## Tested:
- confirmed test suite passing
- upload file to dataset root dir
- upload file to dataset nested dir
- upload directory to dataset root dir
  - *unexpected behavior:* directory name loses first character
- upload directory to dataset nested dir

david-nishi (approved):

PR #1881, opened 2020-11-07
CLI: Add early exit for zero length files

nellh:
This will halt the upload if a file is zero bytes long. The one legitimate way you can end up triggering this check is a .bidsignore file that is supposed to be empty, otherwise the validator check for zero length files is considered an error. This should prevent overwriting files with zero length files in the case where the issue is the file became unreadable after validation but before the upload is complete.

### DIFF
--- a/packages/openneuro-cli/src/upload.js
+++ b/packages/openneuro-cli/src/upload.js
@@ -154,6 +154,15 @@ export const uploadFiles = async ({
       console.error(err)
       controller.abort()
     })
+    fileStream.on('close', () => {
+      if (fileStream.bytesRead === 0) {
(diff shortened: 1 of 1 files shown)


david-nishi (approved):
## Tested:
- uploaded a dataset with an empty .bidsignore file
- confirmed output 
```
? Begin upload? Yes
=======================================================================
Starting a new upload (e8f2123) to dataset: 'ds001070'
ds001070 [----------------------------------------] 0% | ETA: 0s | 0/9
Warning: ".bidsignore" read zero bytes - check that this file is readable and try again
=======================================================================
Upload Complete
```

PR #1883, opened 2020-11-09
App: Catch errors parsing snapshot dates and show 'N/A' if date is invalid

nellh:
This prevents any datasets with invalid snapshot dates from failing to render (showing N/A for the date instead).


david-nishi (approved):
## Tested:
- extracted and ran lines 23-28 and confirmed expected behavior

PR #1887, opened 2020-11-16
Refactor downloads to show CLI option first for browsers without native file API support

nellh:
This hides the browser based download option for browsers that don't work well with it and cleans up some related code for generating zip files in the service worker.


david-nishi (approved):
## Tested:
- download button is not available on Firefox 78.0.2 and Safari 13.1.2, but is available on Chrome 86.0.4240.198 and Edge 86.0.622.69

PR #1891, opened 2020-11-17
Add --ignoreSubjectConsistency to datalad-service

nellh:
This brings the datalad-service validation step into alignment with the uploader checks by disabling the subject consistency checks that run into stack limits.


david-nishi (approved):
## Tested:
- uploaded a dataset with 100 subjects

PR #1892, opened 2020-11-19
Coealesce invalid EthicsApprovals values to a single element array

nellh:
This adds support for automatically fixing non-array EthicsApprovals values, like the other DatasetDescription array fields.


david-nishi (approved):
## Tested:
- upload a dataset with a string value for EthicsApprovals

PR #1898, opened 2020-12-07
Display the affirmed defaced and affirmed consent fields on dataset page

nellh:
This adds these as read only values on the metadata page, making them visible somewhere that doesn't require writing a query to see the values chosen during upload.


david-nishi (approved):
## Tested:
- checked dataset metadata form
- successfully failed to update affirmedDefaced and affirmedConsent

PR #1899, opened 2020-12-07
Fix extra files being returned when multiple directories match prefix

nellh:
Resolves an issue with loading directory trees where a prefix might match multiple subdirs (sub-30 and sub-3 for example).


david-nishi (approved):
## Tested:
- add dataset with subdirectories sub-001, sub-0010, and sub-0011
- confirm Dataset File Tree is correct

PR #1900, opened 2020-12-09
Update to Apollo 3.0 - drop unused deps and rebase on master.

nellh:
This is mostly import fixes thanks to compatibility improvements in more recent releases of Apollo 3, a few extra libraries are dropped and an extraneous HOC in DatasetContent are also removed.


david-nishi (approved):
## Tested
- test suite passing
- sign in/out
- up/download
- edit readme & description
- download, view, and delete for directory and file

PR #1901, opened 2020-12-10
Drop dataset.revision field from MongoDB model

nellh:
This replaces the draft revision pointer with a dynamic resolver which always points at the HEAD from the dataset. This adds an extra request to many draft operations but simplifies complex updates like git pushes required for git-annex push support.

Secondary change here is that since the draft pointer no longer exists, I've implemented a revalidation component to replace the reset draft pointer component used on the admin page.


david-nishi (approved):
## Tested:
- test suite passing
- dataset snapshotted, checked Git Hash on dataset draft and snapshot pages
- checked admin dataset history draft head and latest commits
- re-validated commits and observed network responses

PR #1910, opened 2020-12-18
Fix missing revision resolver call in draft lookups from aggregates

nellh:
This issue is caused by the simple draft resolver here failing to return anything for the revision field after #1901 and this workaround will successfully resolve the most recent revision in aggregates.


david-nishi (approved):
## Tested:
- test suite passing
- view draft
- create snapshot
- confirm new draft ref is HEAD
- make changes
- confirm changes on reload.

PR #1913, opened 2020-12-21
Only show admin tools on drafts (not valid on snapshots)

nellh:
This hides the non-functional admin tools button on snapshots.


david-nishi (approved):
## Tested
- test suite passing
- confirm admin dataset tools are only visible on Draft page

PR #1914, opened 2020-12-22
Automatically rerun validation when it is accessed and not complete

nellh:
* draft -> issues queries will try to rerun validation is no validation data is found.
* Locks the validation step for 5 minutes to prevent duplicate validation runs.


david-nishi (approved):
## Tested:
- test suites passing (jest & pytest)
- uploaded dataset with no issues/warnings
- updated dataset to have only one author and pushed change with git
- viewed 'only one author' warning in browser dataset draft issues

PR #1924, opened 2021-01-12
Add pre-receive hook to run validation on git push

nellh:
Depends on https://github.com/bids-standard/bids-validator/pull/1143

This will run validation on the file tree during a git push. This is limited to checks that do not examine file contents or sizes as only newly pushed objects have their size examined. Newly pushed git objects are examined and limited to 100MB.

The hook is installed for each repo when a push containing new refs is received.

### DIFF
--- a/services/datalad/Dockerfile
+++ b/services/datalad/Dockerfile
@@ -1,5 +1,6 @@
 FROM python:3.8-slim as base
 
+COPY hooks /hooks
 COPY package.json /package.json
 COPY Pipfile /Pipfile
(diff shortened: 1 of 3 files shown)


david-nishi (approved):
## Tested
- test suites passing (pytest & jest)
- pointed bids-validator dependency at `https://github.com/nellh/bids-validator/tree/git-hook-support` and pushed changes with git
  - confirmed validation from browser

nellh:
This is done but won't be merged until the bids-validator release including the required flag.

PR #1931, opened 2021-01-13
1676 snapshot unsaved changes

### LINKED ISSUE #1676: Add banner to drafts when changes have not been snapshotted
I have found the process for adding data into `drafts` and creating a snapshot to share the updates can be enhanced. Changes were made on `drafts` but the owners may not have been aware they needed to generate a snapshot to make them public.  

Adding a banner near the top of the page - perhaps below the banner notifying the owner to make the dataset public. 

There could be a more optimal approach. I think the key is signaling to the dataset owner their changes are not public.

david-nishi:
resolves #1676 

## Major Changes
- added 'published with unsaved changes' banner
Screen Shot 2021-01-13 at 00 30 21

## Minor Changes
- fixed draft subscriber

### DIFF
--- a/packages/openneuro-app/src/sass/bootstrap/_alerts.scss
+++ b/packages/openneuro-app/src/sass/bootstrap/_alerts.scss
@@ -8,8 +8,6 @@
 
 .alert {
   padding: $alert-padding;
-  border: 1px solid transparent;
-  border-radius: $alert-border-radius;
 
   // Headings for larger alerts
   h4 {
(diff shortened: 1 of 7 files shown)


nellh (changes requested):
#1676 has some design updates to this banner than need to be implemented.

david-nishi:
Thanks @nellh, styles are updated

nellh (changes requested):
This is still missing a style fix in #1676, the removal of the border radius and borders on the banner. The primary text is also bolder but might just be font rendering differences.

@thinknoack Is the green border intentional? To me it looks like removing the banner border was the idea.

thinknoack:
> This is still missing a style fix in #1676, the removal of the border radius and borders on the banner. The primary text is also bolder but might just be font rendering differences.
> 
> @thinknoack Is the green border intentional? To me it looks like removing the banner border was the idea.

Yeah. removing the border and radius, while making the alert 100% with no margin-top.

nellh (approved):

PR #1933, opened 2021-01-14
minor update to make logs more useful

thinknoack:


nellh (approved):
Realistically this only fails in unknown ways, so we want the traceback caught by logging, not eaten by the try/except handler.

PR #1934, opened 2021-01-14
Simplify github export and prevent gevent blocking timeouts for new snapshot exports

nellh:
This prevents an issue with timeouts during snapshot exports by applying the same fix from migrate_to_bucket to publish_target.


david-nishi (approved):
## Tested:
- test suites passing
- successfully created snapshot from browser

PR #1944, opened 2021-01-23
Validate branch and tag names on pre-receive

nellh:
Disallows tags not matching basic semver (only numeric versions) or branches except for the standard branch names and the [prefix synced/...](https://git-annex.branchable.com/sync/).

### DIFF
--- a/services/datalad/hooks/pre-receive
+++ b/services/datalad/hooks/pre-receive
@@ -28,6 +28,29 @@ function main() {
       continue
     fi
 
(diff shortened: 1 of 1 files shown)


david-nishi (approved):
## Tested:
- test suites passing (pytest, jest)
- push dataset branch `test` to openneuro with git & confirm error message
- tag dataset w/ "test.tag.1" and push to openneuro with git; confirm error message

PR #1946, opened 2021-01-25
Implement git-annex external remote for pushing annexed objects.

nellh:
This allows git-annex to export these objects directly to OpenNeuro.

Includes documentation additions, the worker endpoints, and openneuro-cli support for a custom remote.

One minor fix resolves a race condition in datalad_service.common.stream.update_file when updating files, this is needed to prevent issues with receiving git-annex objects rapidly but it may affect other kinds of uploads as well.

### DIFF
--- a/docs/git.md
+++ b/docs/git.md
@@ -71,9 +71,31 @@ path=/git/0/ds000001
 EOF
 ```
 
-
 ## Snapshots
 
-Snapshots are represented as git tags. To use all OpenNeuro features, you must have at least one tag with a valid semantic versioning name. This is typically `1.0.0` but can be any valid semantic version.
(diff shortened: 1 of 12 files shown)


nellh:
There's one known issue here, it sometimes hangs after all the work is completed. After about a minute the process will still exit correctly in this case but I haven't been able to narrow down what's causing the delay here.

nellh:
The background process is just the open HTTP/2 socket, I've added an explicit exit after git-annex closes the socket to solve the delay.

david-nishi (approved):
## Tested:
- test suites passing (jest, pytest)
- set openneuro remote on dataset w/ annexed files
- modify and upload annexed files
- pull dataset
- connect to and get annexed files from openneuro remote
- confirm updated files.

PR #1947, opened 2021-01-25
Reexport buttons

### LINKED ISSUE #1896: Add a re-export button to the admin page
It would be nice for the OpenNeuro admin team to be able rerun exports. (related to #1895)

Adding a button to rerun exports from the admin page allows OpenNeuro admin to rerun exports when necessary. Although, adding the status display feature first would be a good idea.

**Additional context**

### LINKED ISSUE #1930: Add export status indicators
For the re-export display, along with the button, add some indication if the export is processing and a log of the latest export status.

### LINKED ISSUE #1943: Add documentation for export status via admin-Kibana dashboard

david-nishi:
resolves #1896, resolves #1930, resolves #1943

## Major Changes:
- Button triggering a reexport to remotes exposed as admin-level tool (`/datasets/ds000000/admin-exports`).
- Button linking to kibana records for `logs-reexporter` added to dataset admin-exports page.
- Add reexportRemotes mutation to GraphQL API (requires admin credentials).
- Add local kibana instance to docker-compose (disabled by default).
- Datalad-service exports process logs from reexporter to elasticsearch, with fields `dataset_id`, `text`, `timestamp`, and `process_start`.
- reexport script will do nothing if latest snapshot exists in remote
- snapshots trigger a reexport

## Minor Changes:
- fix hexsha bug on draft page right after upload
- fix snapshot tool icon getting stuck in loading when clicked
- rename `packages/openneuro-app/src/scripts/admin/admin.users.jsx` to `packages/openneuro-app/src/scripts/admin/users.jsx`

## Notes to Reviewer:
- the link to the reexporter logs in kibana won't work locally unless the kibana service is enabled (docker-compose: comment out line 138)


packages/openneuro-app/src/scripts/resources/kibana.js
```diff
@@ -0,0 +1,22 @@
...
+ *   or use localhost url.
+ * For localhost url to be available, the kibana service must be enabled (docker-compose).
+ */
+const getKibanaURI = () => {
+  if (process.env.ELASTICSEARCH_CLOUD_ID) {
+    const ELASTICSEARCH_CLOUD_ID = process.env.ELASTICSEARCH_CLOUD_ID
+    const base64 = /:(.+?==)$/.exec(ELASTICSEARCH_CLOUD_ID)[1]
+    const decoded = atob(base64)
+    const deploymentId = /\$.+?\$(.+?)$/.exec(decoded)[1]
+    return `${deploymentId}.us-east-1.aws.found.io:9243/app/discover#/`
```
nellh: The hostname here is also in the cloud ID, to make this more robust to deployment (I could see it breaking if the ElasticSearch setup was rebuilt) it should use that.

services/datalad/datalad_service/handlers/reexporter.py
```diff
@@ -0,0 +1,23 @@
+import falcon
+import os.path
+import logging
+import subprocess
+import pty
+from datalad_service.common.elasticsearch import log_reexporter
+
+class ReexporterResource(object):
+    """/info/refs returns current state for either git-receive-pack or git-upload-pack"""
```
nellh: This comment doesn't make sense.

nellh (changes requested):
This issue isn't clear about this but I believe #1896 was aimed at doing this on a per dataset basis.

The script executed by the ReexporterResource is not safe to run generally in production, it will result in the bare S3 export containing the wrong snapshot if it was previously exported and an older snapshot was not. This is something we can check before running it on an individual dataset - ideally this re-export would abort if this case was found. It is also unlikely to complete for all datasets on one worker.

I think a better approach is to call the single dataset export and improve it so that it can handle this case and catch the S3 out of order export situation.

david-nishi:
rebased on master and removed draft hexsha fix

services/datalad/datalad_service/tasks/publish.py
```diff
@@ -104,48 +108,75 @@ def get_dataset_realm(ds, siblings, realm=None):
...
+    client = boto3.client(
+        's3',
+        aws_access_key_id=AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
+    )
+    response = client.get_object(Bucket=realm.s3_bucket, Key=f'{dataset_id}/CHANGES')
+    text = response['Body'].read().decode('utf-8')
+    tag_pattern = re.compile(f'^{tag} ', re.MULTILINE)
+    match = tag_pattern.match(text)
+    return bool(match)
```
nellh: This seems like it will be too fragile now that we are accepting git datasets created outside of the OpenNeuro. It's pretty likely that CHANGES will not always be 1:1 with snapshots if the dataset was not entirely generated on OpenNeuro.

Maybe a better approach is to check if the tree object is available with `git-annex info --json <tag>`?

nellh (changes requested):
A few comments and there are still some test failures.

services/datalad/datalad_service/tasks/publish.py
```diff
@@ -104,48 +109,73 @@ def get_dataset_realm(ds, siblings, realm=None):
...
+def check_s3_has_version(dataset, tag):
+    response = dataset.repo._run_annex_command(
+        'info',
+        annex_options=[
+            '--json',
+            tag,
+        ]
+    )
+    info = json.loads(response[0])
+    return info['success']
```
nellh: This is just checking if the tag exists locally. It needs to verify the expected remote is listed in the output as well.

docker-compose.yml
```diff
@@ -80,6 +80,7 @@ services:
     volumes:
       - ${PERSISTENT_DIR}/datalad:/datalad
       - ./services/datalad/datalad_service:/datalad_service
+      - ./scripts/repair-git-annex-branch-all-tags.sh:/scripts/repair-git-annex-branch-all-tags.sh
```
nellh: I'd remove this line.

nellh (changes requested):

nellh (approved):

PR #1950, opened 2021-01-25
Deny force push requests

nellh:
This requires that within a given branch any new updates include the existing history.


nellh:
This might not be a good idea actually, there's a very legitimate use case for a force push where you create an empty dataset but then don't want to use the settings from OpenNeuro or mess with your existing history at all, so you force push on the initial upload.

Issue #1937, opened 2021-01-14
Regression: Draft is unavailable on first load after upload until initial snapshot is complete


nellh:
**Describe the bug**

If you upload a complex dataset and the initial snapshot is still being created, the draft page throws a new error due to #1931

```javascript
const hasDraftChanges =
  dataset.draft.head !==
  dataset.snapshots[dataset.snapshots.length - 1].hexsha
```

**To Reproduce**
1. Upload dataset
2. After complete but before the snapshot is generated, visit the draft

**Expected behavior**
The draft page should load without snapshots without errors (some data is expected to be unavailable without snapshots).

**Screenshots**
![Screenshot from 2021-01-14 15-19-56](https://user-images.githubusercontent.com/11369795/104660952-6972d980-567c-11eb-8104-2777218b1b4a.png)

ckrountree:
Would be good to include writing a test that renders the draft without any snapshot data.

PR #1953, opened 2021-01-26
Add support for create-dataset command line action

nellh:
This lets you create an empty dataset to make the initial push of an existing DataLad/git-annex dataset.

`openneuro create-dataset`

It outputs a couple lines of helper text:
```
Dataset https://openneuro.org/datasets/ds002414 created.
Git remote: https://openneuro.org/git/0/ds002414
```


nellh:
Note that the resulting dataset is supposed to be accessible but isn't due to #1937 which is fixed by the changes in #1957

tranttommy (approved):

PR #1957, opened 2021-01-26
Fix crash on draft page when no snapshots are available

nellh:
* Adds a jest mock for window.matchMedia
* Adds a test to render the dataset page with no snapshots (minus file tree) to verify this case automatically
* If no snapshots exist, assume there are edits that need a snapshot created


david-nishi (approved):
## Tested:
- test suites passing
- uploaded dataset to browser and viewed draft immediately after
  - confirm draft page loads without errors and snapshot loads when available

PR #1959, opened 2021-01-28
Upgrade Python locked dependencies for datalad-service

nellh:
This is improves cross-architecture development environment compatibility and adds some logic for installing the arm64 builds of git-annex if required.


david-nishi (approved):
## Tested:
- test suites passing (jest, pytest)
- confirmed dataset features from web client
  - uploads
  - downloads
  - snapshots

PR #1965, opened 2021-02-06
Add custom APM transactions with datasetId/snapshotTag for downloads

nellh:
Tracks client being used and matches up the transactions to track download issues and compare the CLI/browser clients directly.


david-nishi (approved):
## Tested:
- test suites working
- confirm log sent to elastic (viewed response)

Issue #1722, opened 2020-07-06
updatePermissions mutation does not return an error when the user does not exist


nellh:
**Describe the bug**
Sharing with a user not present on OpenNeuro should warn you. At the moment, this is silently skipped.

**To Reproduce**
```graphql
mutation updatePermissions(..., userEmail: "doesnothaveanaccount@example.com", level: "read") {
  id
}
```

**Expected behavior**
There should be some indication that adding permissions didn't work.

ckrountree:
### SPRINT REVIEW:
![1721, 1722 and test](https://user-images.githubusercontent.com/27043528/107282291-a1690480-6a0f-11eb-9f91-24fac15f5759.png)

There are unit tests, as well.

PR #1966, opened 2021-02-06
Error handling

### LINKED ISSUE #1721: ManagePermissions component does not handle errors
**Describe the bug**
Neither the ManagePermissions page or UpdateDatasetPermissions child component handles errors received from the server.

**To Reproduce**
1. Share a dataset with an email address not related to an OpenNeuro user
2. You will not get a visible error explaining why the server 

**Expected behavior**
You should get feedback that the user does not have an account on OpenNeuro.

tranttommy:
Resolves #1721, #1722 

## Major Changes:
- Made the `updatePermissions` resolver throw an error when a user isn't found
- Added error handling and feedback when a user tries to give permissions to an email address that doesn't have an OpenNeuro account

### DIFF
--- a/packages/openneuro-app/package.json
+++ b/packages/openneuro-app/package.json
@@ -30,6 +30,7 @@
     "core-js": "^3.3.2",
     "date-fns": "^2.16.1",
     "draft-js": "^0.10.5",
+    "email-validator": "^2.0.4",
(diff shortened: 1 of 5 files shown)


nellh (approved):
Tested these cases and the toasts work well.  :+1:  to the integration test checking for the new case as well.

PR #1974, opened 2021-02-09
Enable HSTS for staging and production

nellh:
We've always enforced TLS with redirects but this sets the HSTS header to be sure.


david-nishi (approved):
## Tested:
confirmed header at localhost

PR #1977, opened 2021-02-19
Improve reliability for long running uploads in openneuro-cli

nellh:
Implements a basic exponential backoff strategy which often allows network issues to recover and continue the upload. The max attempts per file is raised from 3 to 4. Fixes missing retry handler for network aborts on the fetch call for individual files as well.


david-nishi (approved):
## Tested:
- test suites passing
- added infinite sleep to post file resource; confirm "Failed to upload..." message on upload attempt

PR #1978, opened 2021-02-19
Fix missing ETA for some uploads

nellh:
Fixes a minor issue where uploads would sometimes round down to zero for the ETA buffer and not report an ETA.


david-nishi (approved):

ckrountree:
### SPRINT REVIEW
Now ETA is reading properly, rather than `NA`.
Screen Shot 2021-02-22 at 11 29 48 AM

PR #1982, opened 2021-02-23
Re-enable reset draft head button for dataset admin tools

nellh:
Brings back the component for this and re-enables it on the admin page with a new icon to differentiate it from the validation button.


david-nishi (approved):
## Tested:
- test suites passing
- draft ref is updated on reload

## Note:
- I think a good future ticket would be to get the draftUpdated subscriber working, so that the page doesn't need to be refreshed to view changes

PR #1983, opened 2021-02-23
Reduce git push limit to 10MB for non-annexed objects

nellh:
I think we should lower this since it really only affects pushes. Some existing datasets break this rule but you can still push to them if you annex the large files in a new commit.


david-nishi (approved):
## Tested:
- test suite passing

Issue #1970, opened 2021-02-09
Add deface verification to CLI upload 


franklin-feingold:
**Is your feature request related to a problem? Please describe.**
This feature would mimic the defacing verification completed when uploading via the web. 

**Describe the solution you'd like**
A user response to confirm the dataset is defaced if they want to create a new dataset

nellh:
To implement this we want to update the createDataset mutation to require a response for this and update the clients which use it to ask the question in a consistent manner (new dataset upload on web and two CLI commands, upload and create). The mutation should drop the label argument as this breaking change is a good time to fix the signature.

ckrountree:
**Sprint review:** rebuilt the node component to test. Updated prompt call type: `checkbox` to `list` (in action.js and createDataset.js).

PR #1995, opened 2021-03-05
1970 add deface verification

tranttommy:
Addresses #1970 .

- Added `affirmedDefaced` and `affirmedConsent` to `createDataset` mutation
- Updated `createDataset` resolver
- Removed `label` parameter
- Updated tests
- Added verification to CLI

### DIFF
--- a/packages/openneuro-cli/src/createDataset.js
+++ b/packages/openneuro-cli/src/createDataset.js
@@ -1,23 +1,30 @@
 import gql from 'graphql-tag'
+import inquirer from 'inquirer'
 import { getUrl } from './config.js'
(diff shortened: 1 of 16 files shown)


packages/openneuro-cli/src/createDataset.js
```diff
@@ -5,7 +5,7 @@ import { apm } from './apm.js'
 
 const CREATE_DATASET = gql`
   mutation createDataset {
-    createDataset(label: "") {
+    createDataset {
```
nellh: The new arguments should be used in this mutation as well, right?

nellh (approved):
Tested each permutation of this change and everything is working well :+1:

PR #1997, opened 2021-03-08
Enable TypeScript builds for openneuro-app

nellh:
This sets up the necessary changes to include openneuro-app in openneuro-server to enable server side rendering. Most of the TypeScript issues are fixed but this includes some overrides, mostly in the metadata form which is probably getting dropped or majorly refactored by upcoming work.

Major changes:
* Many more TypeScript files in openneuro-app
* Enable typechecking with only specific files excluded (previously the compiler was used but strict typechecking was off)
* Update to React 17 and latest releases of Apollo + GraphQL for TypeScript fixes
* Linting for common mistakes with TypeScript and react-hooks
* Changes TypeScript build root for openneuro-server to root of the project to allow for openneuro-app to be import


david-nishi (changes requested):
## Issues:
- Navbar gear icon does not open logout menu (appears to be an issue with the `DropdownButton` from react-bootstrap).
- Search by text in name, author, readme content returns no results (confirmed running containers and expected clean exits).
- The metadata form on upload doesn’t add it’s fields to Metadata collection (though computed metadata fields are captured). The Add Metadata form under dataset tools works fine.

## Tested:
 - Yarn tests passing
- Front page:
    - Queried data
    - Login
    - Logout (issue)
    - Admin nav item
    - Support nav item and its function
    - Responsive layout
- Admin page:
    - Has expected users
    - Filters
    - Make admin action
    - Block action
- My Dashboard (my and saved datasets) & Public Dashboard
    - Has expected datasets
    - Search (issue)
    - Filters
    - Sorters
    - Responsive layout
- Upload & Download
    - Validates pre-upload
    - Ingests metadata (issue)
    - Prevents upload without defacing affirmations
    - Validates and snapshots once uploaded
    - Browser download
- Dataset Page
    - Has expected:
        - data
        - snapshots
        - git hash
        - validator warnings
    - Readme and description edits
    - File and directory deletions
    - File and directory uploads
    - File viewers (text, json, nifti)
    - File updates
    - Edits to draft are validated immediately
    - Snapshots
    - Draft head reset
    - Publish dataset
    - Remote reexports
    - Adds metadata
    - Comments (add, reply, delete)

nellh:
* Fixed the DropdownButton by replacing it with an nearly identical but much simpler custom component.
* Switched to esbuild for openneuro-app - this makes it much easier to coordinate the build in a way that lets you import it openneuro-server without additional shimming (it compiles things like the image references and workers so that Node.js doesn't fail to parse those)
* Added a root level `yarn build` command which tests builds every package and can watch them

Still in progress:
- [ ] Indexer doesn't run because it had some bad assumptions about how cross package imports work
- [ ] bids-validator crashes because of node only package references
- [ ] Haven't looked into the metadata form issue still - may or may not be an issue now

nellh:
Closing in favor of #2017

Issue #1956, opened 2021-01-26
Keeping validation logs in elastic search


franklin-feingold:
**Is your feature request related to a problem? Please describe.**
Keeping our validation logs for longer than we currently store them

**Describe the solution you'd like**
Pushed to elastic search

nellh:
This should collect the command output (stderr, stdout) when bids-validator fails to exit normally and annotate it with some metadata for what dataset it is, what user triggered the validation, and the exit status. This is written to ElasticSearch as a logs index for analyzing failures.

nellh:
Adding some more specificity to this issue, it refers to failures that occur when bids-validator is run by the OpenNeuro backend service. Validator output is already captured in a Mongo collection but sometimes there's things like tracebacks or unhandled promise exceptions because of a bids-validator bug that we want to track to be able to look for any silent failures.

ckrountree:
### sprint review
Nell will check in pre-release review.

PR #1998, opened 2021-03-09
added apm span to validate function

tranttommy:
Addresses ticket #1956 

- Added apm span for bids validation in openneuro cli

### DIFF
--- a/services/datalad/datalad_service/tasks/validator.py
+++ b/services/datalad/datalad_service/tasks/validator.py
@@ -7,6 +7,7 @@ from gevent import subprocess
 import sentry_sdk
 
 from datalad_service.config import GRAPHQL_ENDPOINT
+from datalad_service.common.elasticsearch import ValidationLogger
 
 
 def setup_validator():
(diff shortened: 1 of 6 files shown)


nellh (changes requested):
This looks fine for collecting a transaction in openneuro-cli so we may want to merge this but the original issue refers to validation when run by OpenNeuro via datalad-service. The idea is that sometimes validation will exit unexpectedly or produce some messages indicating a bug that aren't recorded by the bids-validator API results, so we want to capture the command output and keep that.

services/datalad/datalad_service/tasks/validator.py
```diff
@@ -26,10 +27,10 @@ def validate_dataset_sync(dataset_path, ref):
         process = gevent.subprocess.run(
             ['./node_modules/.bin/bids-validator', '--json', '--ignoreSubjectConsistency', dataset_path], stdout=subprocess.PIPE, timeout=300)
         return json.loads(process.stdout)
-    except subprocess.TimeoutExpired:
+    except subprocess.TimeoutExpired as err:
+        esLogger.log(err)
```
nellh, line 31: This should also handle the case where the error returns quickly. That's anytime the exit code isn't 0. Adding check=True to the subprocess.run arguments will have it throw CalledProcessError that could be handled with the same implementation.

nellh (changes requested):

nellh (approved):

PR #2004, opened 2021-03-10
Fix TypeError for property startTransaction on download without APM

nellh:
This fixes an issue where openneuro download would fail if error reporting is disabled.


nellh:
> This issue also happens when you try to upload a dataset with error reporting disabled but with the "addLabels" method

Yeah, it should fix both cases though. Are you still running into it with an upload somehow?

tranttommy:
Oh never mind, it's all good

tranttommy (approved):

Issue #2012, opened 2021-03-17
Disable publication without snapshots


ckrountree:
This is part of #1973

ckrountree:
### SPRINT REVIEW
Checked this on web and in cli.  Saw both the banner and lack of publication button when no snapshot was available and then added a snapshot and saw the publication button appear and banner message change.

Issue #2011, opened 2021-03-17
Remove snapshot on upload


ckrountree:
This is part of #1973

ckrountree:
### SPRINT REVIEW
No snapshot was created upon upload.

Issue #1973, opened 2021-02-09
Unify datalad/web/cli experience for initial upload, snapshot and publication


effigies:
**Is your feature request related to a problem? Please describe.**

Web and CLI uploads create a snapshot on first upload. This is enabled by a full client-side validation prior to upload. Datalad pushes require server-side validation, and a snapshot cannot be guaranteed in principle.

Currently, the first web edit of the dataset would trigger an automatic snapshot, as well.

**Describe the solution you'd like**

As common an experience as possible, regardless of upload path.

* Stop snapshotting on upload
* Disable publication without snapshots (helpful error message, link to snapshot page)
* Push draft notifications to git clients, e.g. "This dataset is not yet public", "This dataset has no snapshots"
* Email reminder if dataset has no snapshot

**Describe alternatives you've considered**

* None

PR #2015, opened 2021-03-22
1973 unify snapshot ux

tranttommy:
Addresses tickets: #1973, #2011, #2012

- Removed dataset snapshots on upload.
- Disabled dataset publication without a snapshot
- Added email notification for new dataset

### DIFF
--- a/services/datalad/datalad_service/handlers/upload.py
+++ b/services/datalad/datalad_service/handlers/upload.py
@@ -8,6 +8,7 @@ import gevent
 import sentry_sdk
 
 from datalad_service.common.user import get_user_info
+from datalad_service.common.draft import update_head
 from datalad_service.common.annex import CommitInfo
 
 
@@ -37,6 +38,7 @@ class UploadResource(object):
                 gevent.sleep()
                 move_files(upload_path, ds.path)
                 ds.save(unlock_files)
+                update_head(ds, dataset_id, cookies)
                 gevent.sleep()
                 shutil.rmtree(upload_path)
         except:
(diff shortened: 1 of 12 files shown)


services/datalad/datalad_service/handlers/upload.py
```diff
@@ -38,7 +37,6 @@ def _finish_upload(self, dataset_id, upload, name, email, cookies):
                 gevent.sleep()
                 move_files(upload_path, ds.path)
                 ds.save(unlock_files)
-                update_head(ds, dataset_id, cookies)
```
nellh: This is not the right approach for preventing snapshots on upload, this will still create a snapshot on the next commit instead of only when the user creates one.

Here's where that actually happens:
https://github.com/OpenNeuroOrg/openneuro/blob/0fff635c53753cd1c49c4eafc8c30f4ebdcb13fe/packages/openneuro-server/src/graphql/resolvers/draft.js#L24-L36

It might be possible to drop updateRef entirely as this is the only thing it is responsible for at the moment.

david-nishi (changes requested):
## Tested:
- test suite passing
- add dataset via browser
  - confirm snapshot not created
  - confirm publish tool not available

services/datalad/datalad_service/handlers/upload.py
tranttommy: Opted to just remove the section where a new snapshot is created instead of dropping updateRef entirely.

david-nishi (approved):
## Tested:
- test suite passing
- add dataset, delete file, confirm no snapshot

Issue #1971, opened 2021-02-09
Validate .gitattributes on datalad pushes


effigies:
**Is your feature request related to a problem? Please describe.**

Datalad pushes may include `.gitattributes` files that diverge from what we need or would specify ourselves.

**Describe the solution you'd like**

Checkable constraints on the `.gitattributes` file. This could be as simple as "These lines must be present", or do some kind of analysis. I'm not sure if there are any existing tools to help us verify that constraints are satisfied.

The easiest thing would be to have a very strict check and wait for someone to have a valid use-case for deviating before we relax.

**Describe alternatives you've considered**

* None as yet

ckrountree:
### sprint review
![Screen Shot 2021-04-06 at 4 15 41 PM](https://user-images.githubusercontent.com/27043528/113789089-5ad41580-96f3-11eb-8a50-03c27ff64ad9.png)

PR #2016, opened 2021-03-25
made function to check gitattributes file in pre-receive hook

tranttommy:
Addresses ticket: #1971


nellh (commented):
The extra submodule here should be removed.

tranttommy:
Ok it's removed

nellh (approved):

PR #2017, opened 2021-03-27
Enable SSR

nellh:
This supersedes #1997 and includes some fixes from the review of that.

New major changes:
* Switch to Vite for all builds, dropping Webpack
* Additional TypeScript and eslint fixes to allow for less overrides
* Replace useMedia with a CSS approach that's compatible with SSR called [@artsy/fresnel](https://www.npmjs.com/package/@artsy/fresnel)
* Extends openneuro-app with the SSR server (it should be its own package but it complicates things further so this is a temporary home)
* Fixes various TypeScript project reference issues in #1997 
* Replaces all client side cookie handling with SSR compatible (useCookies in [react-cookie](https://www.npmjs.com/package/react-cookie)
* More types as they were needed to keep TSC happy.
* Switch to tsc-watch from ts-node in most places. ts-node/ts-node-dev seems to have some issues with project references and this is only a little slower.
* Disables code splitting because @loadable/component is only working with Webpack and React.lazy does not work in server mode without ugly hacks. Vite is improving this so we can likely turn it back on once upstream fixes arrive.
* Warming the Apollo Client cache on SSR page loads to get rendered queries without JavaScript without the extra re-render after load with JavaScript.

This requires using the dev branch of bids-validator and merging https://github.com/bids-standard/bids-validator/pull/1238

It also depends on a change in Vite to support classic web workers. We need classic workers for Firefox support for bids-validator. This requires some workarounds that I'll break out from this PR

One final note, for testing with SSR try disabling JavaScript. You should be able to click anything which is a real href and have the routes load. Some are visually compromised without JavaScript (the dashboard page) but should work well enough for search indexing.


nellh:
Looks like CI just needs to be updated for tests, they should pass locally. I'm working on the CI issue.

nellh:
This is blocked on the next bids-validator release for merge.

david-nishi (changes requested):
It looks like the metadata from the upload form isn't being saved, but otherwise this looks great.

## Tested:
 - Yarn tests passing
- Front page:
    - Queried data
    - Login
    - Logout 
    - Admin nav item
    - Support nav item and its function
    - Responsive layout
- Admin page
- My Dashboard (my and saved datasets) & Public Dashboard
    - Has expected datasets
    - Search 
    - Filters
    - Sorters
    - Responsive layout
- Upload & Download
    - validator runs properly
    - metadata is saved (issue)
    - defacing/authorization is required
- Dataset Page
    - Has expected:
        - data
        - snapshots
        - git hash
        - validator warnings
    - Readme and description edits
    - File and directory deletions
    - File and directory uploads
    - File viewers (text, json, nifti)
    - File updates
    - Edits to draft are validated immediately
    - Snapshots
    - Draft head reset
    - Publish dataset
    - Remote reexports
    - Adds metadata
    - Comments (add, reply, delete)

nellh:
Looks like the metadata issue is older than this branch but I went ahead and fixed it.

david-nishi (approved):

david-nishi:
Just a snapshot needs updating, but otherwise good to go.

PR #2021, opened 2021-03-29
1574 Delete Updates

david-nishi:
## Major Changes
- datasetChanges query, which returns up to 1000 of the most recent dataset changes (creations, modifications, or deletions)
- datasetChanged subscription, which pushes a notification whenever a new dataset change is made

### Note:
This is in lieu of sending a notification to Yarik on dataset deletions.

### DIFF
--- a/packages/openneuro-server/src/models/dataset.js
+++ b/packages/openneuro-server/src/models/dataset.js
@@ -1,4 +1,5 @@
 import mongoose from 'mongoose'
+import Changes from './changes.js'
 
 const datasetSchema = new mongoose.Schema(
   {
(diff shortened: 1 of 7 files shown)


packages/openneuro-server/src/models/dataset.js
```diff
@@ -44,6 +45,29 @@ datasetSchema.virtual('subscriptions', {
...
+})
+
+datasetSchema.post('deleteOne', function() {
+  const datasetId = this._conditions ? this._conditions.id : null
+  new DatasetChange({
+    datasetId,
+    deleted: true,
+  }).save()
+})
+
```
nellh, line 70: I like this approach to tracking these.

nellh (approved):

Issue #2013, opened 2021-03-18
Update react-ga to address Analytics upgrade


ckrountree:
Google is telling us to upgrade to gtag.js from analytics.js.

When we deploy this update we’ll want to schedule a cutover at an off-hour time (at midnight 4/1, for example), so that we’re not changing any analytics collection mechanism mid-month.

nellh:
Looks like we should actually drop react-ga and use our own component if we want to do this.

nellh:
Adding gtag.js globally (index.html) and just providing a wrapping function for pageview and event should be fine. The wrapping function is to keep the window call required in one place (window is not available in isomorphic rendering).

PR #2022, opened 2021-03-29
removed react-ga and replaced with gtag module

tranttommy:
Addresses: #2013

### DIFF
--- a/packages/openneuro-app/src/index.html
+++ b/packages/openneuro-app/src/index.html
@@ -2,6 +2,14 @@
 <html>
 
 <head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=GA_MEASUREMENT_ID"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date())
+      gtag('config', 'GA_MEASUREMENT_ID');
+    </script>
     <base href="/"></base>
     <meta http-equiv="expires" content="0" />
     <meta charset="utf-8" />
--- a/packages/openneuro-app/src/index.html
+++ b/packages/openneuro-app/src/index.html
@@ -4,11 +4,6 @@
 <head>
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <script async src="https://www.googletagmanager.com/gtag/js"></script>
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-    </script>
     <base href="/"></base>
     <meta http-equiv="expires" content="0" />
     <meta charset="utf-8" />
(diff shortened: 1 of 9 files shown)


packages/openneuro-app/src/scripts/client.jsx
```diff
@@ -31,8 +30,6 @@ loadConfig().then(config => {
     })
   }
 
-  GoogleAnalytics.initialize(config.analytics.trackingId)
```
nellh, line 34: This part still needs to happen so it can get the GA_MEASUREMENT_ID from configuration.

```ts
gtag('config', config.analytics.trackingId);
```

packages/openneuro-app/src/scripts/utils/gtag.js
```diff
@@ -0,0 +1,3 @@
+export const pageview = (...args) => window && window.gtag(...args)
```
nellh: The call for [page_view now looks like this](https://developers.google.com/analytics/devguides/collection/gtagjs/pages#page_view_event):
```ts
gtag('event', 'page_view', {
  page_title: '<Page Title>',
  page_location: '<Page Location>',
  page_path: '<Page Path>'
  send_to: '<GA_MEASUREMENT_ID>'
})```

File: packages/openneuro-app/src/scripts/utils/gtag.js [hunk 3]
```diff
@@ -0,0 +1,3 @@
...
+
+export const event = (...args) => window && window.gtag(...args)
```
nellh: These will need to be `typeof window !== undefined` to avoid throwing an error when it isn't. Or better `'gtag' in globalThis` to check global in Node and window in the browser main thread.

nellh:
```ts
gtag('event', <action>, {
  'event_category': <category>,
  'event_label': <label>,
  'value': <value>
})
```

nellh (changes requested):

packages/openneuro-app/src/index.html
```diff
@@ -2,6 +2,13 @@
 <html>
 
 <head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=GA_MEASUREMENT_ID"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date())
+    </script>
```
nellh: Looks like this won't work because of the id defined here. Can we just include the script tag here with src="https://www.googletagmanager.com/gtag/js" and do any initialization in client.jsx?

nellh (changes requested):
One more change we should make but this looks good otherwise.

packages/openneuro-app/src/scripts/client.jsx
```diff
@@ -31,7 +30,7 @@ loadConfig().then(config => {
...
+  'gtag' in globalThis && gtag('config', config.analytics.trackingId)
```
nellh: dataLayer initialization is missing here, so gtag is never defined.

nellh (changes requested):

nellh (approved):

Issue #1884, opened 2020-11-09
[CLI] Error Stream Premature close


franklin-feingold:
**Describe the bug**
The cli outputs an `ERR_STREAM_PREMATURE_CLOSE` error. The dataset owners started the uploader and at a point in the process, this error is output for each remaining file. It appears the connection is killed to the server. it is unclear why it was killed. This error was replicated by a few dataset owners with a stable connection. 

**To Reproduce**
1. Error was produced when uploading to ds003243 and ds002241
2. Dataset owners have up to date versions of node and npm. The dataset owners have found this in v3.23.0 (it didn't appear the new version would induce this error. 

**Expected behavior**
Dataset uploads

alexlicohen:
I am also getting this error when trying to create a new data set, or update an existing one.

(I tried the suggestion to install openneuro into a separate dir from here: https://github.com/OpenNeuroOrg/openneuro/issues/1842#issuecomment-709491276 but still no luck)

Error message:
```
../software/node_modules/.bin/openneuro upload --dataset ds003363 GSP1000/
Adding files to "ds003363"
bids-validator@1.5.7

        Summary:                  Available Tasks:        Available Modalities:
        1572 Files, 7.56GB                                T1w
        1568 - Subjects
        1 - Session


	If you have any questions, please post on https://neurostars.org/tags/bids.

=======================================================================
2750 files to be uploaded with a total size of 219.6 GB
? Begin upload? Yes
=======================================================================
Starting a new upload (3eb55d62) to dataset: 'ds003363'
ds003363 [=---------------------------------------] 1% | ETA: 95945s | 38/2750
Error [ERR_STREAM_PREMATURE_CLOSE] [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at ClientHttp2Stream.onclose (internal/streams/end-of-stream.js:121:38)
    at ClientHttp2Stream.elasticAPMCallbackWrapper (/Volumes/Drobo/Research/software/node_modules/elastic-apm-node/lib/instrumentation/index.js:309:27)
    at ClientHttp2Stream.emit (events.js:327:22)
    at ClientHttp2Stream.EventEmitter.emit (domain.js:486:12)
    at closeStream (internal/http2/core.js:1766:14)
    at ClientHttp2Stream._destroy (internal/http2/core.js:2171:7)
    at ClientHttp2Stream.destroy (internal/streams/destroy.js:39:8)
    at internal/http2/core.js:1063:46
[...]

Issue #2001, opened 2021-03-09
Build static openneuro-cli as alternative to package manager installation (npm or yarn)


ckrountree:
Consider the cli making a static build.

ckrountree:
This will be better addressed after releasing SSR.

nellh:
#2717 is the modern version of this issue.

PR #2032, opened 2021-04-01
caught error if startTransaction throws, updated apm calls to be conditional

tranttommy:
Addresses: #1884


nellh (commented):
To maintain Node.js 12 compatibility we can't use optional chaining in openneuro-cli because the package is run directly by end users. #2001 could let us target an older release of Node.js for the built package. How about we just defer this fix until #2001 is done?

Issue #2029, opened 2021-03-31
Add second analytics property id for GA4


nellh:
**Is your feature request related to a problem? Please describe.**
OpenNeuro only lets you configure one Google Analytics token, if we have two we can maintain universal analytics and GA4 simultaneously.

**Describe the solution you'd like**
Make the environment variable into an array? It's only consumed in one place so parsing out multiple values and configuring them should work.

**Additional context**
Need to verify that all events go to both properties.

PR #2033, opened 2021-04-01
refactored gtag initialization to take an array of tracking IDs

tranttommy:
Addresses: #2029

### DIFF
--- a/packages/openneuro-app/src/scripts/utils/gtag.js
+++ b/packages/openneuro-app/src/scripts/utils/gtag.js
@@ -6,7 +6,8 @@ function gtag() {
 
 gtag('js', new Date())
 
-export const initialize = trackingId => gtag('config', trackingId)
+export const initialize = trackingIds =>
+  trackingIds.forEach(trackingId => gtag('config', trackingId))
 
 export const pageview = path =>
   gtag('event', 'page_view', {
--- a/packages/openneuro-server/src/handlers/config.js
+++ b/packages/openneuro-server/src/handlers/config.js
@@ -35,7 +35,9 @@ const config = {
   },
 
   analytics: {
-    trackingId: process.env.GOOGLE_TRACKING_ID,
+    trackingIds: process.env.GOOGLE_TRACKING_IDS.split(',').map(id =>
+      id.trim(),
+    ),
   },
 
   sentry: {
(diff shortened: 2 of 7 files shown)


packages/openneuro-app/src/scripts/utils/gtag.js
```diff
@@ -6,7 +6,8 @@ function gtag() {
 
 gtag('js', new Date())
 
-export const initialize = trackingId => gtag('config', trackingId)
+export const initialize = trackingIds =>
+  trackingIds.forEach(trackingId => gtag('config', trackingId))
```
nellh, line 10: This doesn't look like it will send all events to all tracking ids?

packages/openneuro-server/src/handlers/config.js
```diff
@@ -35,7 +35,9 @@ const config = {
   },
 
   analytics: {
-    trackingId: process.env.GOOGLE_TRACKING_ID,
+    trackingIds: process.env.GOOGLE_TRACKING_IDS.split(',').map(id =>
+      id.trim(),
+    ),
```
nellh, line 40: This should handle the case where the value is not provided or only one value is provided.

nellh (changes requested):
Can we handle the case where this is undefined or there's only one value?

nellh:
One more piece of feedback, this should also update the Helm chart since it defines these environment variables and the variable is renamed.

packages/openneuro-app/src/scripts/utils/gtag.js
tranttommy: `gtag('config', trackingId)` just needs to be called for each tracking ID, and subsequent calls will go to both unless specified with a `send_to` property.

https://www.optimizesmart.com/using-multiple-trackers-for-cross-domain-tracking-in-universal-analytics/#b8

packages/openneuro-server/src/handlers/config.js
tranttommy: One value will just result in an array containing that one value. No value will result in an array with an empty string as the first value

packages/openneuro-app/src/scripts/utils/gtag.js
nellh: Okay, great. I found the current documentation for that.

https://developers.google.com/analytics/devguides/collection/gtagjs/#configure_additional_google_analytics_properties

packages/openneuro-server/src/handlers/config.js
nellh: Okay, the issue I was having with this is a setup problem with my config for this, not an issue with this section. Thanks!

nellh (approved):

PR #2034, opened 2021-04-02
2003 annex files

david-nishi:
## Major Changes:
- annexed files have a Remove Annexed Object(s) button on Draft and Snapshots for admins, and a Flag Sensitive Data button for dataset owners.
- annexed files that have been flagged or removed are tracked in mongodb collection BadAnnexObject
- annexed files that have been removed are not sent to the client


services/datalad/datalad_service/common/annex.py
```diff
@@ -196,7 +196,7 @@ def get_repo_files(dataset, branch='HEAD'):
             filename = symlinkFilenames[(index - 1) // 2]
             file_id = compute_file_hash(key, filename)
             files.append({'filename': filename, 'size': int(
-                size), 'id': file_id, 'key': key, 'urls': []})
+                size), 'id': file_id, 'key': key, 'urls': [], 'annexed': True})
```
nellh, line 199: Tried seeing if we could do this without passing the key value back to the client because it increases the size of the files response but it is required to keep the Apollo cache id unique for the same content in different trees. We might want to add some additional pagination support to improve the performance for larger datasets but I think this works for now.

services/datalad/datalad_service/handlers/annex_objects.py
```diff
@@ -0,0 +1,31 @@
...
+                file = next(f for f in files if annex_key == f.get('key'))
+            except StopIteration:
+                resp.media = {'error': 'file does not exist'}
+                resp.status = falcon.HTTP_BAD_REQUEST
+            urls = file.get('urls')
+
+            gevent.spawn(remove_file_remotes, self.store, urls)
+        else:
+            resp.media = {'error': 'annex-key is missing'}
+            resp.status = falcon.HTTP_BAD_REQUEST
```
nellh: Should probably just be HTTP_NOT_FOUND

services/datalad/tests/test_annex.py
```diff
@@ -35,13 +35,15 @@ def test_get_ls_tree_line():
     symlinkObjects = []
     read_ls_tree_line("""100644 blob a786c385bd1812410d01177affb6ce834d85facd     459	dataset_description.json""",
                       files, symlinkFilenames, symlinkObjects)
+    print(files)
```
nellh: Extra print to remove.

services/datalad/datalad_service/handlers/annex_objects.py
```diff
@@ -0,0 +1,31 @@
...
+            ds = self.store.get_dataset(dataset)
+            files = get_repo_files(ds, snapshot)
+            try:
+                file = next(f for f in files if annex_key == f.get('key'))
+            except StopIteration:
+                resp.media = {'error': 'file does not exist'}
+                resp.status = falcon.HTTP_BAD_REQUEST
+            urls = file.get('urls')
+
+            gevent.spawn(remove_file_remotes, self.store, urls)
```
nellh, line 29: The local copy needs to be removed as well.

packages/openneuro-server/src/models/badAnnexObject.js
```diff
@@ -0,0 +1,18 @@
...
+    remover: { _id: String },
+    flagged: { type: Boolean, default: false },
+    flagger: { _id: String },
+  },
+  { timestamps: { createdAt: 'created_at' } },
+)
+
+const BadAnnexObject = mongoose.model('BadAnnexObject', badAnnexObject)
+
+export default BadAnnexObject
```
nellh: Before merging this, the new collection should have a TypeScript interface to avoid conflicts with #2054

This model could use an index on datasetId, annexKey as well.

nellh (changes requested):
Have a few changes, the one other comment here is that there doesn't seem to be a way to find flagged files for admins. Maybe we should have an admin tab that shows a list of files for each dataset with flagged but not-deleted files? Would be a pretty simple aggregate query to just list them.

packages/openneuro-server/src/models/badAnnexObject.js
ckrountree: @david-nishi please add a tooltip explaining what this does.

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
nellh, line 1: Is this because of mixing Husky versions? There should just be .husky/pre-commit so I'm wondering how this directory got created for you?

nellh (changes requested):
The local annex file issue still needs to be addressed.

services/datalad/datalad_service/handlers/annex_objects.py
```diff
@@ -26,6 +28,7 @@ def on_delete(self, req, resp, dataset, snapshot, annex_key):
             urls = file.get('urls')
 
             gevent.spawn(remove_file_remotes, self.store, urls)
+            FilesResource.on_delete(self, req, resp, dataset, filename)
```
nellh: This generates a new commit, which is not the intention of removing the annexed content.

You don't need the filename at all to accomplish this, instead just remove the annex object file. It's probably best to just call `git-annex drop --force --key=` to do this.

nellh (changes requested):

nellh (approved):

PR #2043, opened 2021-04-08
Enable optimized SSR production build for Helm deployment

nellh:
This optimizes #2017 when running in production.

* Build optimized bundles in production mode
* SSR server now uses the optimized bundle
* express.static to serve static assets
* Reworks configuration to load at runtime (instead of import.meta.env from Vite)


nellh:
Ran into some configuration issues deploying to staging, added on some configuration changes here moving it to runtime for SSR mode as well. Also found a merge issue from #2017 and #2033 where the client was still referencing the old variable name.

david-nishi (approved):
## Tested:
- ran build

PR #2046, opened 2021-04-09
Fix admin page regression from #2017

nellh:
This fixes a crash on the admin page when rendering the user list.


david-nishi (approved):
## Tested:
- viewed admin dashboard
- made/unmade user admin
- test suite passing

PR #2051, opened 2021-04-13
Replace react-bootstrap Panel and PanelGroup with simplified versions

nellh:
Drops Panel and PanelGroup components from react-bootstrap.

This fixes a production build issue by upgrading react-bootstrap past 0.31.2 which crashes at import once built by Vite because of invalid imports via babel-runtime -> corejs. 0.33 refactored how panel components work, so this replicates the old behavior with simplified components that accept the same signature.


david-nishi (approved):
## Tested:
- tests passing
- confirmed working modal

PR #2052, opened 2021-04-13
Switch to Yarn 2

nellh:
Switches to Yarn 2 in PnP mode which eliminates node_modules and the need for `docker-compose build node` to update dependencies.

Needs testing on macOS because I believe this results in the ability to share the cache with the host without conflicts but I've only tested Linux/Linux.

One breaking change, to run any node scripts they must be started with yarn to result in correct dependency resolution. `docker-compose down --remove-orphans` is a good idea too.


david-nishi (approved):
## Tested:
- test suites passing
- searched datasets
- uploaded & downloaded dataset

PR #2053, opened 2021-04-13
Increase cache duration for SSR resources

nellh:
This restores some missing cache-control headers removed in #2017


david-nishi (approved):
## Tested
- test suite passing
- Container builds and serve connects to db

Issue #2044, opened 2021-04-08
Update to Mongoose 5.12 (or later)


nellh:
**Describe the bug**
Going from Mongoose 5.10 -> 5.12 causes a number of TypeScript build issues. We will need to fix these to get things working with the upcoming Mongoose 6 release.

**To Reproduce**

1. cd packages/openneuro-server && yarn install mongoose@5.12.2
2. yarn tsc -b

**Expected behavior**
Successful build.

PR #2054, opened 2021-04-14
Mongoose upgrade

tranttommy:
Addresses: #2044 
- `yarn tsc -b` builds successfully
- Upgraded Mongoose to 5.12.3.
- Updated models to be in TypeScript
- Updated .js model imports to .ts


packages/openneuro-app/src/scripts/datalad/mutations/add-metadata.jsx
```diff
@@ -82,7 +82,7 @@ const runValidations = values =>
     .map(validation => {
       const relevantValues = validation.fields.map(key => values[key])
       // TODO - This doesn't seem necessary?
-      // @ts-expect-error
+      // @ts-ignore
```
nellh: Curious what requires changing these to @ts-ignore?

tranttommy: When I ran `yarn tsc -b`, I kept getting an `Unused '@ts-expect-error' directive` message yesterday. When I took out the line, then the compiler failed. It seems to be working now, so I'll be change it back.

packages/openneuro-server/src/models/permission.ts
```diff
@@ -0,0 +1,18 @@
+import mongoose, { Document } from 'mongoose'
+const { Schema, model } = mongoose
+
+export interface PermissionDocument extends Document {
+  datasetId: string
+  userId: string
+  level: string
```
nellh: This can be typed as the string enumeration:

```ts
level: 'ro' | 'rw' | 'admin'
```

packages/openneuro-server/src/models/summary.ts
```diff
@@ -0,0 +1,34 @@
...
+const { Schema, model } = mongoose
+
+export interface SummaryDocument extends Document {
+  id: string
+  datasetId: string
+  sessions: any[]
+  subjects: any[]
+  subjectMetadata: object
+  tasks: any[]
+  modalities: any[]
```
nellh: Subjects, sessions, tasks, and modalities can be string arrays.

nellh (changes requested):
Rebased and tested, working great but I have a few suggestions for the types.

nellh (approved):

PR #2059, opened 2021-04-14
Speed up front page participants aggregate with query cache

nellh:
This causes a lot of load at startup as our cache has been protecting us from how long this query really takes. This quick fix just caches the query result for 24 hours to avoid stacking queries on MongoDB.


nellh:
Closing this in favor of #2060 which makes a similar but more general improvement.

PR #2060, opened 2021-04-15
feat(server): Enable distributed Apollo cache

nellh:
This reduces some query duplication by sharing the API cache between instances (via Redis).


david-nishi (approved):
## Tested:
- builds
- serves datasets on dashboard and dataset pages

PR #2061, opened 2021-04-15
Optimize participiantCount resolver execution time

nellh:
This reduces the database time to resolve participantCount from 5000ms per call to 90ms.

1. Adds a sparse index on dataset.public to locate public datasets in the collection faster
1. The query is inverted to scan starting from the datasets collection allowing an index scan using that index
1. $lookup call to summaries is optimized with a new index on summaries.id
1. $lookup call to snapshots gets a narrower index on only the datasetId field
1. Consolidates some extra steps for a shorter aggregate pipeline

### DIFF
--- a/packages/openneuro-server/src/graphql/resolvers/snapshots.js
+++ b/packages/openneuro-server/src/graphql/resolvers/snapshots.js
@@ -6,7 +6,7 @@ import { description } from './description.js'
 import { summary } from './summary.js'
 import { snapshotIssues } from './issues.js'
 import { getFiles, filterFiles } from '../../datalad/files.js'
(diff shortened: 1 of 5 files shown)


david-nishi (changes requested):
Very cool! I think this would let us remove cacheControl from the participantCount endpoint too.

## Tested:
- test suite passing
- ran participantCount query w/o cache

nellh:
> Very cool! I think this would let us remove cacheControl from the participantCount endpoint too.

It's still a potentially expensive query, I agree we can reduce the cache duration though. I'll lower it to one hour.

david-nishi (approved):

Issue #2062, opened 2021-04-15
Use GraphQL DatasetFiles for all download methods


nellh:
**Is your feature request related to a problem? Please describe.**
An endpoint for `/crn/datasets/{datasetId}/download` and  `/crn/datasets/${datasetId}/snapshots/${snapshotTag}/download` was created to support streaming browser downloads, which is no longer implemented. This API is still used for file access API downloads and the CLI. The behavior of this older API is different to support the use case of bundling in the browser even though this no longer applies. We should consolidate on one implementation for providing file objects.

**Describe the solution you'd like**
The GraphQL APIs for this already exist, the CLI and React app just need to be updated to use them.

**Additional context**
Some related discussion in #1905

PR #2067, opened 2021-04-19
2062 graphql download

tranttommy:
Addresses: #2062


nellh (changes requested):
This is missing the CLI implementation.

nellh (approved):
Tested this with several datasets and it's working great.

PR #2068, opened 2021-04-19
fix(datalad-service): Avoid losing bids-validator output on expected errors

nellh:
Fix a regression introduced in #1998, the validation output is useful on non-zero output unless it can't be parsed (which is usually due to bids-validator bugs). Instead of checking for non-zero exit to log failures, we should just catch the JSON parsing exception and log at that point.

This extends the error logging with the full stderr and stdout, which would have helped with tracking this down.


david-nishi (approved):
## Tested:
- test suites passing (yarn & pytest)

PR #2069, opened 2021-04-19
fix(datalad-service): Use yarn v2 in node_modules mode for dev install

nellh:
This fixes the missing bids-validator install in development due to #2052


david-nishi (approved):
## Tested:
- installed local copy of bids-validator

PR #2071, opened 2021-04-19
Storybook setup and components

thinknoack:
Initial setup for storybook as well as a number of components.


packages/openneuro-components/package.json
```diff
@@ -0,0 +1,126 @@
...
+    "style-loader": "^2.0.0",
+    "ts-node-dev": "^1.1.6",
+    "url-loader": "^1.0.1",
+    "vite": "2.1.5",
+    "webpack": "^4.20.2",
+    "webpack-cli": "^3.1.2",
+    "webpack-dev-server": "^3.1.9",
+    "webpack-merge": "^4.1.4",
+    "webpack-visualizer-plugin": "^0.1.11",
+    "worker-plugin": "^3.2.0"
```
nellh: This looks like it's just a copy of the app deps? I think we should try to reduce this to packages used in this one.

packages/openneuro-components/package.json
```diff
@@ -0,0 +1,126 @@
+{
+  "name": "openneuro-components",
```
nellh: This should be "@openneuro/components" to be compatible with #2079

packages/openneuro-components/stories/icon/Icon.tsx
```diff
@@ -0,0 +1,49 @@
...
+export interface IconProps {
+  primary?: boolean
+  secondary?: boolean
+  backgroundColor?: string
+  label?: string
+  disabled?: boolean
+  icon?: string
+  color?: string
+  imgSrc?: string
+  iconSize?: string
```
nellh: Looks like some of these are required and should be marked as such for types.

packages/openneuro-components/stories/icon/Icon.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  iconSize: '23px',
+}
+
+export const IconWithBackgroundColor = Template.bind({})
+IconWithBackgroundColor.args = {
+  icon: 'fab fa-google',
+  iconSize: '12px',
+  color: '#fff',
+  backgroundColor: '#555',
+}
```
nellh: I think this could be organized better, either just removing the stories part of the file paths or moving the components to the top level and keeping stories in a stories directory. Making it a separate directory makes it easier to not include the stories in the final bundle but as long as the pattern to identify stories is simple, they be mixed in which might be preferred just to keep you from digging when working on a component and the related stories.

nellh (changes requested):
Made some comments about organization. I think we should at least shorten the paths, the stories part of openneuro-components/stories/accordion/AccordionTab.tsx seems unnecessary. I like the idea of keeping the stories with the components and just following .stories.tsx suffix convention, so they can be excluded from bundling.

This should use the Vite support in Storybook 6.2, it seems to be using Webpack at the moment.

I also want to bring up scss vs emotion, is there an advantage to using scss for component styles over emotion? I see the reasons for global styles but hopefully we can reduce our global style usage significantly.

packages/openneuro-components/stories/icon/Icon.stories.tsx
nellh: To be consistent, how about this? openneuro-components/src/icon/Icon.tsx and openneuro-components/src/icon/Icon.stories.tsx

thinknoack: Should this be ?

openneuro-components/src/icon/Icon.tsx and openneuro-components/**stories**/icon/Icon.stories.tsx

nellh: It could be but I don't think it needs to be. We can add an exclude pattern for .stories.* files when compiling it. I like having tests next to the real component, just generally makes it easier to work on both at the same time with less hunting around the repo. What do you think?

thinknoack:
> I also want to bring up scss vs emotion, is there an advantage to using scss for component styles over emotion? I see the reasons for global styles but hopefully we can reduce our global style usage significantly.

No reason besides personal comfort. I am including the scss in the component which I think is only loaded when that component exists(reducing the global styles), correct? 

I am also open to using styled components using emotion but I like having the styles outside of the component (if there are a lot).

nellh (changes requested):
Looks like the content files are missing and this breaks tests. It also needs to be rebased or merged, just lockfiles so should be straightforward (just start the merge, regenerate those, and commit the new versions).

nellh:
> Looks like the content files are missing and this breaks tests. It also needs to be rebased or merged, just lockfiles so should be straightforward (just start the merge, regenerate those, and commit the new versions).

Thinking about this, maybe the best structure is the components package just mocks the content and openneuro-app contains the real content?

thinknoack (commented):
content dir was being ignored, changed the dir name and updated the refs

packages/openneuro-components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,49 @@
...
+  toggle,
+  closeText,
+  ...props
+}) => {
+  const showModal = isOpen ? 'show-modal' : 'hide-modal'
+  return (
+    <div className={'modal-wrapper ' + showModal} {...props}>
+      <div className="overlay" onClick={toggle}></div>
+
+      <div className="grid modal" isOpen={isOpen} toggle={toggle}>
```
nellh: @thinknoack Looks like isOpen and toggle can just be removed here. They are not valid div properties.

nellh (changes requested):

packages/openneuro-components/src/modal/Modal.tsx
thinknoack: *removed

PR #2072, opened 2021-04-19
fix(cli): Handle import resolution difference with yarn pnp

nellh:
This should be a singleton but isn't when pnp and esm (the npm package) are combined.

Ideally we'll drop esm and use pnp with native esm here when Node 12 is EOL and Yarn supports it (https://github.com/yarnpkg/berry/issues/638)


tranttommy (approved):
CLI works!

PR #2079, opened 2021-04-20
refactor: Rename npm packages to use @openneuro namespace

nellh:
This renames the npm packages to use a namespace. This largely a dev focused change but we'll want to mark openneuro-cli as deprecated and point users to @openneuro/cli once a release has been published.

This will avoid conflicts with any public packages named openneuro-* as we add a few new packages to the project (components, ssr, etc).


david-nishi (approved):
## Tested:
- test suites passing
- containers built

PR #2080, opened 2021-04-21
fix(cli): Incorrect auth headers and stream type changes from #2017

nellh:
Two fixes for CLI download issues:
1. This does use cookie rather than bearer token auth because it is a REST endpoint.
2. The actual file transfer was reverted to cross-fetch but was still using the fetch-h2 API.


tranttommy (approved):
Download works!

Issue #1985, opened 2021-02-23
Bulk delete via the web UI


franklin-feingold:
**Is your feature request related to a problem? Please describe.**
There is not a clear way to bulk remove objects from the file tree. This can be problematic as our datasets continue to grow in size.

**Describe the solution you'd like**
Ability to select multiple files and remove them at one time, rather than the current implementation of removing them one by one (or one directory at a time)

ckrountree:
Server side of this has been implemented.  We still need to address the design/frontend.  Also, add an integration test that tests the delete functionality.

ckrountree:
~~Add a tooltip explaining what this does.~~

nellh:
> Add a tooltip explaining what this does.

I believe this was meant for #2034

ckrountree:
The multi-delete functionality is not part of the app yet. However, we can add it now because the api is available. @tranttommy, can you add the basics backend api calls, etc.? Then @thinknoack can start building out the UI.

PR #2085, opened 2021-04-24
1985 bulk delete

### LINKED ISSUE #2081: Create bulk delete functionality in web front end
This is a necessary step before continuing on #1985

### LINKED ISSUE #2082: Create bulk delete resolver
This is necessary for completing #1985 and relates to #2081

### LINKED ISSUE #2083: Create bulk delete function on datalad-server interface
Maybe will be directly in datalad, this area seems to be the source of the bugs.

This is necessary for completing #1985 and relates to #2081 and #2082

tranttommy:
Tickets addressed:
resolves #2081, resolves #2082, resolves #2083 as part of #1985

### DIFF
--- a/packages/openneuro-server/src/graphql/resolvers/dataset.js
+++ b/packages/openneuro-server/src/graphql/resolvers/dataset.js
@@ -128,34 +128,6 @@ export const deleteDataset = async (
 /**
  * Delete files from a draft
  */
-export const deletePath = async (
-  obj,
-  { datasetId, path },
-  { user, userInfo },
-) => {
-  try {
-    await checkDatasetWrite(datasetId, user, userInfo)
-    await datalad.deletePath(datasetId, path, userInfo)
-    pubsub.publish('filesUpdated', {
-      datasetId,
-      filesUpdated: {
-        action: 'DELETE',
-        payload: [
-          {
-            id: `${path}:0`,
-            filename: path,
-          },
-        ],
-      },
-    })
-    return true
-  } catch (err) {
-    console.error(err)
-    Sentry.captureException(err)
-    return false
-  }
-}
-
 export const deleteFiles = async (
   obj,
   { datasetId, files },
--- a/packages/openneuro-server/src/graphql/resolvers/dataset.js
+++ b/packages/openneuro-server/src/graphql/resolvers/dataset.js
@@ -181,6 +181,44 @@ export const deleteFile = async (
   }
 }
 
+export const deleteBulk = async (
+  obj,
+  { datasetId, files },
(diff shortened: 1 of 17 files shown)


services/datalad/datalad_service/handlers/files.py
```diff
@@ -131,6 +132,24 @@ def on_delete(self, req, resp, dataset, filename):
...
+            name, email = get_user_info(req)
+            if name and email:
+                media_dict['name'] = name
+                media_dict['email'] = email
+            try:
+                remove_files(self.store, dataset, files=deleted_files, name=name, email=email, cookies=req.cookies)
+                resp.media = media_dict
+                resp.status = falcon.HTTP_OK
+            except:
+                resp.status = falcon.HTTP_INTERNAL_SERVER_ERROR
```
nellh: The implementation here looks fine but I wouldn't overload the same API now three paths to doing the same thing. If this is its own resource handler it wouldn't need the req.bounded_stream hack. req.media does this internally when accessed, including the JSON parsing.

This should also have test coverage.

packages/openneuro-server/src/graphql/schema.js
```diff
@@ -161,6 +161,13 @@ export const typeDefs = `
     reexportRemotes(datasetId: ID!): Boolean
     # Reset draft commit
     resetDraft(datasetId: ID!, ref: String!): Boolean
+    # Bulk delete individual files
+    deleteBulk(datasetId: ID!, files: [DeleteFile]): Boolean
```
nellh: Can we consolidate the delete APIs down to one deleteFiles API? The other cases map pretty well to this implementation.

packages/openneuro-server/src/graphql/resolvers/dataset.js
```diff
@@ -181,6 +181,35 @@ export const deleteFile = async (
...
+  { datasetId, files },
+  { user, userInfo },
+) => {
+  try {
+    await checkDatasetWrite(datasetId, user, userInfo)
+    const deletedFiles = await datalad
+      .deleteBulk(datasetId, files, userInfo)
+      .then(filenames =>
+        Promise.all(filenames.map(filename => new UpdatedFile(filename))),
+      )
```
nellh, line 142: Looks like this can fail silently if you delete files that don't exist?

nellh:
This looks good except the styling needs some tweaks. Maybe we could just put the bulk delete button in the top level actions and replace the checkbox with a toggle kind of like the star/follow buttons as a quick improvement?

image

tranttommy:
I believe @thinknoack will be taking care of the styling

Issue #2087, opened 2021-04-26
Combine delete APIs into one


ckrountree:
Related to master ticket: #1985
Related additional tickets (2081-2083)

PR #2091, opened 2021-04-28
Delete file api refactor

tranttommy:
Addresses: #2087 

Notes: No more recursive query in the delete request, directory deletes are assumed to be recursive


nellh (approved):
Tested single file and directory delete, working well.

PR #2092, opened 2021-04-29
rename some files and add the radio component

thinknoack:
default radio component using scss styling


nellh (changes requested):
Couple simple type fixes but maybe the radio components deserve an integration test with react-testing-library to automatically verify the radio behavior since there was some discussion about how it should work?

packages/openneuro-components/src/form/Radio.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+        radioGroupName="example"
+        defaultChecked
+        checked={active === 0}
+        onClick={() => setActive(0)}
+      />
+      <Radio
+        elementId="radio3"
+        label="Bannanas"
+        radioGroupName="example"
+        radioNumber="1"
```
tranttommy: Remove `radioNumber`, reasons below

packages/openneuro-components/src/form/Radio.tsx
```diff
@@ -0,0 +1,35 @@
...
+  radioNumber,
+  onClick,
+  checked,
+  ...props
+}) => {
+  return (
+    <span className="custom-radio">
+      <input
+        type="radio"
+        id={elementId}
```
tranttommy: Since this `Radio` component mostly functions as a wrapper for the radio input type, I think that the prop names should match the input attribute names. This way, someone coming in that knows how radio buttons work can just assume that the props are like a regular input type radio, and can avoid checking which props go to which attributes. (ie `radioGroupName` -> `name`). I also think `label` and `value` can both get their own props with `label` being used for the label and `value` being used for the button value

packages/openneuro-components/src/form/Radio.tsx
```diff
@@ -0,0 +1,35 @@
...
+        name={radioGroupName}
+        checked={checked}
+        onClick={onClick}
+        value={label}
+      />
+      <label htmlFor={elementId}>{label}</label>
```
tranttommy: It looks like the `elementId` prop is just used for the `htmlFor` attribute. If that's the case, I don't think we need an `id` prop, and can make a unique id with something like `const id = name + label` and then use that

packages/openneuro-components/src/form/RadioWrap.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export interface RadioWrapProps {
+  children: object
+  radioWrapID: string
+  layout: string
+}
+
+export const RadioWrap: React.FC<RadioWrapProps> = ({
+  children,
+  radioWrapID,
```
tranttommy: I think just calling the prop `id` would be fine

packages/openneuro-components/src/form/RadioWrap.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import './forms.scss'
+
+export interface RadioWrapProps {
+  children: object
+  radioWrapID: string
+  layout: string
+}
+
+export const RadioWrap: React.FC<RadioWrapProps> = ({
```
tranttommy: So this is an idea on another way we can implement this. We could make this into a `RadioGroup` component that takes an array of `Radio` props (ie `[ { label: 'Apple' value: 'apple' } ]` and then have this component loop through it and make the radio buttons there. We can talk about more specifics if it's something you think we can do/explore.

tranttommy (changes requested):

packages/openneuro-components/src/form/Radio.tsx
```diff
@@ -0,0 +1,35 @@
...
+export interface RadioProps {
+  elementId?: string
+  label: string
+  radioGroupName: string
+  radioNumber: string
+  onClick?: () => void
+  checked: boolean
+}
+
+export const Radio: React.FC<RadioProps> = ({
```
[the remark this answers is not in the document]
nellh: @tranttommy Thanks for finding that, that's a good breakdown of the tradeoffs.

packages/openneuro-components/src/form/RadioGroup.tsx
```diff
@@ -0,0 +1,32 @@
...
+import './forms.scss'
+
+export interface RadioGroupProps {
+  id: string
+  layout: string
+}
+
+export const RadioGroup: React.FC<RadioGroupProps> = ({
+  radioArr,
+  id,
```
tranttommy: I think we can remove `id` usage until we run into a situation where we need to bring it back

packages/openneuro-components/src/form/RadioGroup.tsx
```diff
@@ -0,0 +1,32 @@
...
+  layout,
+  ...props
+}) => {
+  const [active, setActive] = React.useState(0)
+
+  return (
+    <div className={'on-radio-wrapper' + ' ' + layout} {...props} id={id}>
+      {radioArr.map((item, index) => (
+        <Radio
+          name={item.name}
```
tranttommy: Since all the radio buttons in a `RadioGroup` should always have the same name, I think the `name` can be a prop for `RadioGroup` and then the `radioArr` doesn't need a `name` property

tranttommy (changes requested):

tranttommy (approved):

PR #2093, opened 2021-04-30
Pet public dataset count

tranttommy:
Made AggregateCount component that can be used to create the PublicDatasetCount and ParticipantCount components. 

Now needs styling.

Address GitLab tickets: 73, 74


thinknoack (approved):
approved upon successful merge of #2094

PR #2094, opened 2021-04-30
adding agg count styles and placements

thinknoack:
updating styling for agg count in the header and modality-cubes

### DIFF
--- a/packages/openneuro-components/src/header/LandingExpandedHeader.tsx
+++ b/packages/openneuro-components/src/header/LandingExpandedHeader.tsx
@@ -41,7 +41,7 @@ export const LandingExpandedHeader: React.FC<LandingExpandedHeaderProps> = ({
         <div className="grid grid-between">
           <div className="col col-6 expaned-h-left">
             {frontPage.pageDescription}
-            <div className="header-aggriate">
+            <div className="header-aggregate">
               <AggregateCount type="publicDataset" count={202} />
               <AggregateCount type="publicDataset" count={22} />
             </div>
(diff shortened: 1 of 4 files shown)


packages/openneuro-components/src/header/header.scss
```diff
@@ -125,6 +125,18 @@
       margin-top: auto;
       margin-bottom: 30px;
     }
+    .header-aggriate {
+      display: flex;
+      justify-content: space-between;
+      max-width: 90%;
+      .aggregate-count {
```
tranttommy, line 132: Having the AggregateCount's styles come from the header component's styles makes it less modular. This also makes its individual Storybook entry less useful.

I think that there should be separate scss file for AggregateCount. And then we can pass in something like a `placement` prop whose values can be `'header'` or `'cube'` and then have a class for each placement. Something like `.header-aggregate-count` and `.cube-aggregate-count`

packages/openneuro-components/src/header/header.scss
```diff
@@ -125,6 +125,18 @@
       margin-top: auto;
       margin-bottom: 30px;
     }
+    .header-aggriate {
```
tranttommy: Typo

packages/openneuro-components/src/header/LandingExpandedHeader.tsx
```diff
@@ -35,6 +41,10 @@ export const LandingExpandedHeader: React.FC<LandingExpandedHeaderProps> = ({
         <div className="grid grid-between">
           <div className="col col-6 expaned-h-left">
             {frontPage.pageDescription}
+            <div className="header-aggriate">
```
tranttommy: Typo

tranttommy (changes requested):
Ignore this

PR #2095, opened 2021-04-30
fix(client): Fix runtime issues with @openneuro/indexer

nellh:
This cleans up the exports from @openneuro/client which was causing a reference error running the production build of @openneuro/indexer.


david-nishi (approved):
## Tested:
- ran build
- ran a dataset search

PR #2096, opened 2021-05-01
created tooltip

thinknoack:


packages/openneuro-components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,27 @@
...
+  flow: 'up' | 'down' | 'left' | 'right'
+}
+
+export const Tooltip = ({
+  children,
+  tooltip,
+  flow,
+  ...props
+}: TooltipProps) => {
+  const myRef = React.createRef<HTMLSpanElement>()
```
tranttommy: You should use the `useRef` hook instead for this

packages/openneuro-components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,27 @@
...
+export interface TooltipProps {
+  tooltip: string
+  flow: 'up' | 'down' | 'left' | 'right'
+}
+
+export const Tooltip = ({
+  children,
+  tooltip,
+  flow,
+  ...props
```
tranttommy: What might the `...props` be used for?

packages/openneuro-components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect } from 'react'
+import './tooltip.scss'
+
+export interface TooltipProps {
+  tooltip: string
+  flow: 'up' | 'down' | 'left' | 'right'
+}
+
+export const Tooltip = ({
+  children,
```
tranttommy: Need to include children in the `interface`. The type for `children` would be `React.ReactNode`

packages/openneuro-components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const Template: Story<TooltipProps> = args => <Tooltip {...args} />
+
+export const Up = Template.bind({})
+Up.args = {
+  children: <>Lorem ipsum</>,
+  tooltip: 'this is the tooltip using the data attr',
+  flow: 'up',
+}
+Up.parameters = {
```
tranttommy, line 23: What does this `parameters` property do?

tranttommy (changes requested):

tranttommy (approved):
Tooltips look good, but we may want to merge in the Radio buttons branch before this. I would suggest not branching off another component branch next time to make things easier to review and merge.

thinknoack (commented):
> Tooltips look good, but we may want to merge in the Radio buttons branch before this. I would suggest not branching off another component branch next time to make things easier to review and merge.

yep, thats how my other branches are. There was a reason I branched off of radio but I am not remembering why at the moment.

PR #2097, opened 2021-05-03
Dropdown

thinknoack:
adding sortby dropdown that shows on search results

### DIFF
--- a/packages/openneuro-components/src/dropdown/Dropdown.tsx
+++ b/packages/openneuro-components/src/dropdown/Dropdown.tsx
@@ -8,7 +8,6 @@ export interface DropdownProps {
 
 export const Dropdown = ({ children, label }: DropdownProps) => {
   const [isOpen, setIsOpen] = React.useState(false)
-  console.log(isOpen)
 
   function removeIsOpen(ref) {
     useEffect(() => {
@@ -25,11 +24,13 @@ export const Dropdown = ({ children, label }: DropdownProps) => {
   }
   const wrapperRef = useRef(null)
   removeIsOpen(wrapperRef)
+
   return (
-    <div className="dropdown-wrapper" ref={wrapperRef}>
-      <div
-        className={`toggle ${isOpen ? 'active' : ''}`}
-        onClick={() => setIsOpen(prevIsOpen => !prevIsOpen)}>
+    <div
+      className="dropdown-wrapper"
+      ref={wrapperRef}
+      onClick={() => setIsOpen(prevIsOpen => !prevIsOpen)}>
+      <div className={`toggle ${isOpen ? 'active' : ''}`}>
         <span>{label}</span>
       </div>
       <div className={`menu ${isOpen ? 'expanded' : 'collapsed'}`}>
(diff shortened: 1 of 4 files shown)


packages/openneuro-components/src/dropdown/DropdownList.tsx
```diff
@@ -0,0 +1,37 @@
...
+    },
+  ]
+}
+
+export const DropdownList = ({ items }: DropdownListProps) => {
+  return (
+    <Dropdown
+      label={
+        <div className="list-label">
+          <b>SORT BY:</b> Newest{' '}
```
tranttommy: We need a way to set "SORT BY" to the item that's been selected.

packages/openneuro-components/src/dropdown/DropdownList.tsx
```diff
@@ -0,0 +1,37 @@
...
+          <i className="fas fa-exchange-alt fa-rotate-90" />{' '}
+        </div>
+      }>
+      <div className="dropdown-list">
+        <ul>
```
tranttommy, line 33: Is there a reason why we're using `ul` and `li` elements instead of `select` and `option` elements?

packages/openneuro-components/src/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import './dropdown.scss'
+
+export interface DropdownProps {
+  label: Record<string, any>
```
tranttommy, line 5: What is the `Record` type and where is it coming from?

packages/openneuro-components/src/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,23 @@
...
+  children
+}
+
+export const Dropdown = ({ children, label }: DropdownProps) => {
+  const [isOpen, setOpen] = React.useState(false)
+  return (
+    <div className="dropdown-wrapper" onClick={() => setOpen(!isOpen)}>
```
tranttommy: Since the new value of `isOpen` is being set based on what its previous value is, it's better practice to pass in a callback that uses the previous `isOpen` value instead of using `isOpen` directly.

ie `onClick={() => setOpen(prevIsOpen => !prevIsOpen)}`

Also the setter function should be called `setIsOpen`

tranttommy (changes requested):

packages/openneuro-components/src/dropdown/DropdownList.tsx
thinknoack: Select and option are real limited when styling. This is a pretty standard alternative. This dropdown is more like how a menu dropdown works. If you think that this works better as a form element we can change it to select and option.

packages/openneuro-components/src/dropdown/Dropdown.tsx
thinknoack: Hmm. not sure what you mean. Nell mentioned using Record over object when a prop could be a string or an element.

nellh: It's a utility type for general key/value type mappings. https://www.typescriptlang.org/docs/handbook/utility-types.html

ckrountree:
This addresses [GitLab #91](https://gitlab.com/squishymedia/nimh/openneuropet/-/issues/91)

PR #2098, opened 2021-05-03
2077 delete dataset docs

david-nishi:
## Major Changes
- add deleting a dataset to API and User Guide docs


nellh (changes requested):
This doesn't really cover the programmatic update changes in #2021 that prompted this issue.

PR #2099, opened 2021-05-03
fix(docker): Don't depend on bash for arch check

thinknoack:


tranttommy:
`docker-compose up` works on my end!

tranttommy (approved):

PR #2100, opened 2021-05-03
fix(docker): Run node containers as non-root

nellh:
This helps avoid permission conflicts on files that are shared by the host and these containers do not depend on root access.


tranttommy (approved):
Running `docker-compose up` works!

PR #2101, opened 2021-05-04
fix: Use init container for yarn install

nellh:
Requires docker-compose 1.29.0

This avoids multiple yarn installs and potential conflicts on start (particularly common on macOS).


tranttommy (commented):
This is working well on my end!

david-nishi:
Smooth start here as well.

PR #2104, opened 2021-05-05
fix(ssr): Disable SSR for authenticated users

nellh:
Increases cache timeout for SSR pages to improve cache hits. This does result in some extra fade-ins in various places on the first load but we'll be improving that in the new components.


david-nishi (approved):
## Tested:
- test suite passing
- build runs
- server handles logged-in & logged-out states

PR #2105, opened 2021-05-06
fix(dev): Workaround Vite bug with * exports

nellh:
After debugging this for a while, this seems like it's just a bug in path handling in how Vite parses this file in SSR mode. This workaround seems to resolve the issue.


tranttommy (approved):
Making changes no longer breaks imports from `@openneuro/client`!

PR #2110, opened 2021-05-07
refactor(graphql): Generate types for GraphQL schemas

nellh:
Adds automatic type generation using [graphql-codegen](https://www.graphql-code-generator.com/).

This is a work in progress because it requires giving queries and mutations unique names. I fixed openneuro-client but not the other packages to try things out.

TODO:
- [ ] Fix all the overlapping/anonymous queries/mutations in the CLI and React packages.
- [ ] Have CI verify the generated types match the committed version
- [ ] Decide if we want to generate separate types for each package or just stick them all in @openneuro/client or a new utility package.
- [ ] Document how to update types (pretty much just `yarn codegen`)


packages/openneuro-client/README.md
```diff
@@ -0,0 +1,6 @@
+# Generated schema files
+
+```sh
+# To regenerate graphql-schema.json
+yarn apollo service:download --endpoint=http://localhost:9876/crn/graphql graphql-schema.json
+```
```
nellh: This is from apollo:codegen and can just be removed.

.yarnrc.yml
```diff
@@ -3,3 +3,10 @@ plugins:
     spec: "@yarnpkg/plugin-typescript"
 
 yarnPath: .yarn/releases/yarn-berry.js
+
+packageExtensions:
+  "docz-core@*":
+    dependencies:
+      cross-spawn: "*"
+      tslib: "*"
+      opn: "*"
```
nellh: This is not needed either.

nellh:
Closing this in favor of a new approach.

PR #2130, opened 2021-05-13
Facets

thinknoack:


tranttommy (approved):
I think this may need to be refactored in the future as the use case becomes more clear, but it looks good for now.

PR #2132, opened 2021-05-14
fix(ssr): Use cache-and-network for drafts

nellh:
Prevents SSR cache age from displaying incorrect drafts after load.

### DIFF
--- a/packages/openneuro-app/src/scripts/datalad/dataset/dataset-query.jsx
+++ b/packages/openneuro-app/src/scripts/datalad/dataset/dataset-query.jsx
@@ -111,6 +111,7 @@ export const DatasetQueryHook = ({ datasetId, draft, history }) => {
     {
       variables: { datasetId },
       errorPolicy: 'all',
+      fetchPolicy: draft ? 'cache-and-network' : 'cache-first',
     },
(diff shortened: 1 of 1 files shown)


tranttommy:
I don't know if it's related, but when I upload a dataset and then try to view it, I see the dataset page for a brief moment but then it automatically refreshes a couple of times and then it turns into this:
Screen Shot 2021-05-14 at 11 57 21

When I go to "My Datasets", this happens: 
Screen Shot 2021-05-14 at 11 56 47

nellh:
> I don't know if it's related, but when I upload a dataset and then try to view it, I see the dataset page for a brief moment but then it automatically refreshes a couple of times and then it turns into this:
> Screen Shot 2021-05-14 at 11 57 21
> 
> When I go to "My Datasets", this happens:
> Screen Shot 2021-05-14 at 11 56 47

Not related, we should have an issue for this though. It's caused by bids-validator@1.7.1 + the Vite dev server.

tranttommy:
This is working with the merge of #2133

ckrountree:
@franklin-feingold  This should address the issue with ds001705.

PR #2133, opened 2021-05-14
fix(dev): Enable production build pluralize package fix on vite dev server

nellh:
Fix for https://github.com/OpenNeuroOrg/openneuro/pull/2132#issuecomment-841440157


tranttommy (approved):
Dataset upload and viewing dashing works!

Issue #101, opened 2017-09-24
Job polling sometimes passes null to mongo query $addToSet argument


nellh:
> server_1  | (node:1) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 9120): MongoError: The argument to $each in $addToSet must be an array but it was of type null

This might be a duplicate of #68

nellh:
Not related to #68. Getting a fix ready for this.

nellh:
Fixed in #115

PR #2139, opened 2021-05-22
Warnbutton

thinknoack:
clean up and simplification.


thinknoack:
gitlab issues #100 #101

packages/openneuro-components/src/warn-button/WarnButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+          size="xsmall"
+        />
+
+        <Button
+          className="btn-warn-component success"
+          iconSize="12px"
+          icon="fa fa-check"
+          color="#fff"
+          backgroundColor="#00b489"
+          onClick={onClick}
```
david-nishi: this should also toggle displayOptions when clicked

david-nishi: This isn't in the current WarnButton, but i think it'd be good to add a loading state at some point.

david-nishi (changes requested):

david-nishi:
It looks like the ci build is failing. Looks like the signOutAndRedirect prop in UserMenu.tsx needs to be updated to `(history: any) => void` and the Tooltip component directory will need to be made lowercase.

packages/openneuro-components/src/warn-button/WarnButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const viewAction = (
+    <div className="warn-btn-group " role="group">
+      <div className="slide-in">
+        <Button
+          className="btn-warn-component cancel"
+          iconSize="12px"
+          icon="fa fa-times"
+          color="#fff"
+          backgroundColor="#c00342"
+          onClick={() => setDisplayOptions(!displayOptions)}
```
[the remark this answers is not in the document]
thinknoack: the parameter type throws an error. keeping boolean works tho. thoughts?

packages/openneuro-components/src/warn-button/WarnButton.tsx
thinknoack: created a confirmClick function, let me know if this is a good approach

david-nishi (approved):

PR #2140, opened 2021-05-24
update to some scss vars and created new toggle counter button

thinknoack:


thinknoack:
github issues #100 #101

packages/openneuro-components/src/count-toggle/CountToggle.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { Story, Meta } from '@storybook/react'
+import { useQuery, useMutation } from 'graphql-hooks'
```
nellh: These imports are available in @apollo/client and mixing them from another package will cause some subtle type mismatches.

```ts
import { useQuery, useMutation } from '@apollo/client'
```

packages/openneuro-components/src/button/Button.tsx
```diff
@@ -14,7 +14,8 @@ export interface ButtonProps {
   color?: string
   imgSrc?: string
   iconSize?: string
-  buttonClass?: string
+  className?: string
+  children
```
nellh: The type for children is `React.ReactNode` if it can take any arbitrary child component tree.

packages/openneuro-components/src/count-toggle/CountToggle.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+  disabled,
+  tooltip,
+}) => {
+  const [displayOptions, setDisplayOptions] = React.useState(false)
+  const [count, setCount] = React.useState(1)
+
+  const onClick = () => {
+    let newCount
+    count === 1 ? (newCount = 2) : (newCount = 1)
+    setCount(newCount)
```
nellh:
```suggestion
    setCount(count === 1 ? 2 : 1)
```

packages/openneuro-components/package.json
```diff
@@ -7,8 +7,10 @@
   "author": "Squishymedia",
   "license": "MIT",
   "dependencies": {
+    "@apollo/client": "^3.3.19",
     "@mdx-js/react": "^1.6.22",
     "date-fns": "^2.21.1",
+    "graphql-hooks": "^5.1.1",
```
nellh:
```suggestion
```

nellh (changes requested):

packages/openneuro-components/src/count-toggle/CountToggle.stories.tsx
thinknoack: this was an unused import so I removed it.

packages/openneuro-components/src/button/Button.tsx
thinknoack: I believe I found all the type declarations and changed them to children: React.ReactNode

david-nishi:
This looks good, but to fix the ci issue it will probably be easiest to rebase onto the warnbutton branch after Tooltip is updated.

PR #2142, opened 2021-05-25
Landingheader

thinknoack:


thinknoack:
dependent on PR https://github.com/OpenNeuroOrg/openneuro/pull/2140

packages/openneuro-components/src/count-toggle/CountToggle.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <div className="toggle-counter-wrap">
+        {tooltip ? (
+          <Tooltip flow="up" tooltip={tooltip}>
+            {disabled ? disabledToggleButton : toggleButton}
+          </Tooltip>
+        ) : disabled ? (
+          disabledToggleButton
+        ) : (
+          toggleButton
+        )}
```
david-nishi: For readability, i think it'd be good to extract `disabled ? disabledToggleButton : toggleButton` to a const.

packages/openneuro-components/src/count-toggle/CountToggle.tsx
```diff
@@ -0,0 +1,76 @@
...
+  tooltip,
+  onClick,
+  displayOptions,
+  setDisplayOptions,
+  count,
+}: CountToggleProps) => {
+  const toggleClick = () => {
+    onClick()
+    setDisplayOptions(!displayOptions)
+  }
```
david-nishi: This can belong to the parent component, so that only toggleClick and displayOptions need to be passed in.

I think for this application, it would also be good to rename displayOptions to `clicked` or `showClicked`.

packages/openneuro-components/src/facets/FacetListWrap.tsx
```diff
@@ -23,49 +29,61 @@ export const FacetListWrap = ({
   items,
   selected,
   setSelected,
+  startOpen,
+  label,
+  accordionStyle,
+  dropdown,
 }: FacetListWrapProps) => {
   const setSetter = (e, item) => {
```
david-nishi: I think a longer name like `setSelectorNoPropagation` would help with readability

packages/openneuro-components/src/facets/FacetListWrap.tsx
```diff
@@ -23,49 +29,61 @@ export const FacetListWrap = ({
...
+        dropdown={dropdown}>
+        <div className="facet-list">
+          <ul className="level-1">
+            {items.map((item, index) => (
+              <li
+                key={index}
+                onClick={e => setSetter(e, item)}
+                className={
+                  selected && selected.value == item.value
+                    ? 'selected-facet facet'
```
david-nishi, line 56: I'm not seeing the color change, but the class is getting updated. Is that expected for the storybook component?

packages/openneuro-components/src/header/LandingExpandedHeader.tsx
```diff
@@ -1,12 +1,14 @@
...
 
 import { Button } from '../button/Button'
 import { ModalityCube } from '../modality-cube/ModalityCube'
+import { Input } from '../input/Input'
 import { cubeData } from '../mock-content/modality-cube-content.jsx'
 import orcidIcon from '../assets/orcid_24x24.png'
 import { AggregateCount } from '../aggregate-count/AggregateCount'
 
 import { frontPage } from '../mock-content/front-page-content.jsx'
+import { FrontFacetExample } from '../facets/Facet.stories'
```
david-nishi, line 12: The cubeWrap mostly looks great, but gets distorted at some view widths.

david-nishi (changes requested):

packages/openneuro-components/src/header/LandingExpandedHeader.tsx
david-nishi: Maybe a container of set width? i'm not sure how to fix this

packages/openneuro-components/src/facets/FacetListWrap.tsx
thinknoack: it was subtle because it was using the primary color. I adjusted it to use the header primary color.

packages/openneuro-components/src/header/LandingExpandedHeader.tsx
thinknoack: We have to do some additional responsive work. I think at smaller viewports we will stack them and get rid of the cube. undecided at the moment.

david-nishi (approved):

PR #2144, opened 2021-05-27
107 feature flag

david-nishi:
## Major Changes:
- Add endpoints /crn/feature-flag/redesign-2021/enable and /crn/feature-flag/redesign-2021/disable.
- Create FeatureToggle component.
- Add FeatureToggle for redesign to site.

## Notes & Considerations:
- Enable and disable endpoints are minimal. Do we want design on these?


nellh (changes requested):
The endpoints should redirect to the home page, that way no design for them needed and they just magically take you to the right version of the site. Otherwise, working great.

nellh (approved):

PR #2147, opened 2021-06-01
adding facet date picker

thinknoack:


packages/openneuro-components/src/facets/FacetDatePicker.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <AccordionWrap className="facet-accordion facet-date-picker">
+      <AccordionTab
+        accordionStyle={accordionStyle}
+        label={label}
+        startOpen={startOpen}
+        dropdown={dropdown}>
+        <div className="range-input">
+          <DateRangePicker
+            onChange={setSelected}
```
david-nishi, line 35: `onChange` passes `null` to the callback when the clear button is pressed, so where `setSelected` is defined we should capture that case and set the value to `[null, null]`.

packages/openneuro-components/src/search-page/FiltersBlock.tsx
```diff
@@ -30,37 +31,73 @@ export const FiltersBlock = ({
...
+            <span>
+              {type === 'Age' || type === 'Subjects'
+                ? item[0] + ' - ' + item[1]
+                : item}
+              <span>&times;</span>
+            </span>
+          </li>
+        </>
+      )
   }
```
david-nishi, line 53: If we do end up using the chiclets for the text list inputs, then i think we should pull this out into its own component. That way we can pass in things like `type` and `item`, as well as a callback for removing that filter from the searchParams.

packages/openneuro-components/src/search-page/FiltersBlock.tsx
```diff
@@ -30,37 +31,73 @@ export const FiltersBlock = ({
...
+          _listItem('Subjects', subjectRange)}
         {author_pi && _listItem('Author/PI', author_pi)}
         {gender && _listItem('Gender', gender)}
         {task && _listItem('Task', task)}
         {diagnosis && _listItem('Diagnosis', diagnosis)}
         {section && _listItem('Section', section)}
         {species && _listItem('Species', species)}
         {domain && _listItem('Domain', domain)}
+        {subjectRange && !dateIsNull && (
+          <li className="Date-Range">
```
david-nishi, line 87: This would also be its own component.

david-nishi (changes requested):

packages/openneuro-components/src/search-page/FiltersBlock.tsx
thinknoack: I have another branch where I have updated this so I will do this there, that work?

packages/openneuro-components/src/search-page/FiltersBlock.tsx
thinknoack: same as above.

david-nishi (approved):

PR #2153, opened 2021-06-08
Search params

david-nishi:
187, 188

## Major changes
- Implement Search param term inputs (`keywords`, `authors`, and `tasks`).
- Make term chiclets individually removable.

## Minor changes
- Move global stylesheet behind feature toggle (causes first paint by SSR to be unstyled, might want to implement with a wrapper class instead).
- Implement add on `enter` for term input.
- Prevent term input when value is empty or already exists in list.

### DIFF
--- a/packages/openneuro-components/src/search-page/FilterDateItem.tsx
+++ b/packages/openneuro-components/src/search-page/FilterDateItem.tsx
@@ -1,12 +1,21 @@
 import React from 'react'
 
 type DateRange = [Date | null, Date | null]
+type Item = {
+  param: string
+  value: DateRange
+}
 
 export interface FilterDateItemProps {
-  item: DateRange
+  item: Item
   type?: string
+  removeFilterItem?(param, value): void
 }
-export const FilterDateItem = ({ item, type }: FilterDateItemProps) => {
+export const FilterDateItem = ({
+  item,
+  type,
+  removeFilterItem = () => {},
+}: FilterDateItemProps) => {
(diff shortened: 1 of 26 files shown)


packages/openneuro-components/src/search-page/FilterDateItem.tsx
```diff
@@ -24,7 +33,9 @@ export const FilterDateItem = ({ item, type }: FilterDateItemProps) => {
           <strong>{type}:</strong>
           <span>
             {startMonth}/{startDay}/{startYear} - {endMonth}/{endDay}/{endYear}
-            <span>&times;</span>
+            <span onClick={() => removeFilterItem(item.param, item.value)}>
```
thinknoack, line 36: same question as the other removeFilterItem being on the parent

packages/openneuro-components/src/search-page/FilterListItem.tsx
```diff
@@ -19,9 +31,11 @@ export const FilterListItem = ({ item, type }: FilterListItemProps) => {
           <strong>{type}:</strong>
           <span>
             {type === 'Age' || type === 'Subjects'
-              ? item[0] + ' - ' + item[1]
-              : item}
-            <span>&times;</span>
+              ? item.value[0] + ' - ' + item.value[1]
+              : item.value}
+            <span onClick={() => removeFilterItem(item.param, item.value)}>
```
thinknoack, line 36: same question as the other removeFilterItem being on the parent

thinknoack (approved):

thinknoack (approved):

PR #2158, opened 2021-06-12
update to dataset type radios and date picker

thinknoack:


david-nishi (approved):
I encountered a formatting issue with the query string while reviewing, and while looking into it also found a bug with @apollo/client's useQuery hook. I've pushed some changes to resolve them, and the rest of the PR looks good.

PR #2161, opened 2021-06-17
Front page updates

thinknoack:
Front page edits are ready for review. I have one issue I need some additional support with. Activity sliders are throwing errors when loaded in @on/app.


packages/openneuro-app/src/scripts/refactor_2021/containers/front-page-container.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ActivitySlider,
+  Contributors,
+  GetUpdates,
+  Infographic,
+  Icon,
+  RecentData,
+  TopViewed,
+} from '@openneuro/components'
+
+const FrontPageContainer: FC = () => {
```
nellh:
```suggestion
const FrontPageContainer: React.FC = () => {
```

packages/openneuro-components/src/activity-slider/ActivitySlider.tsx
```diff
@@ -10,69 +10,81 @@ import 'slick-carousel/slick/slick-theme.css'
...
+  console.log(data)
   return (
-    <div className={'container activity-slider' + ' ' + sliderClass}>
+    <div className={'activity-slider' + ' ' + className}>
       <h3>{slideHeader}</h3>
-      <Slider {...props}>
-        {datasets.map((item, index) => (
+      <Slider {...sliderArgs}>
+        {data.map(({ node, index }) => (
           <>
```
nellh: Probably want to add the key here and use the node identity instead of the list index to keep updates from breaking the list.

nellh (changes requested):

packages/openneuro-components/src/activity-slider/ActivitySlider.tsx
thinknoack: I added the Key but still get a unique key error, maybe because Slick is injecting extra markup around the inner mapped children?

nellh: It should be key={node.id} to avoid the ordering issue, otherwise looks like it should work.

packages/openneuro-components/src/activity-slider/ActivitySlider.tsx
```diff
@@ -1,96 +1,139 @@
...
-                  <h4>{item.node.latestSnapshot.description.Name}</h4>
-                </div>
+      {/* @ts-expect-error */}
+      <Carousel
+        infinite={infinite}
+        keyBoardControl={keyBoardControl}
+        containerClass={containerClass}
+        itemClass={itemClass}
+        showDots={showDots}
+        responsive={responsive}
```
nellh, line 82: Do you know how this library handles responsive? I think we'll want this to render in SSR mode and my guess is it won't (can't test due to broken SSR on master).

packages/openneuro-components/src/activity-slider/ActivitySlider.tsx
```diff
@@ -1,96 +1,139 @@
...
-                  </div>
-                </div>
+            <div className="ds-modality">
+              <div className="hexagon-wrapper">
+                {node.latestSnapshot.summary !== null ? (
+                  <>
+                    <div
+                      className={
+                        'hexagon ' +
+                        node.latestSnapshot.summary.modalities[0].substr(0, 4)
```
nellh, line 121: This is probably good for now but we should add a primaryModality field.

nellh (changes requested):
Running into one issue with the front page without the cookie set (grey bars like this).
image

Also SSR is broken but it looks like this is an issue on master.

packages/openneuro-components/src/activity-slider/ActivitySlider.tsx
thinknoack: The front page w/out cookies set is getting some styling from openneuro-components it looks like the css switch for enabled/disabled is no longer working?

the multi carousel module does have support for ssr
https://www.npmjs.com/package/react-multi-carousel

packages/openneuro-components/src/activity-slider/ActivitySlider.tsx
thinknoack: yeah this is also used on the dataset page. I added a TODO for later.

PR #2170, opened 2021-06-22
Modality Portals

david-nishi:
# Major Changes
- bring mri modality page to app
- connect modality select to links (expanded header & general search)
- remove modality select from modality search
- restrict modality_selected to modalities on their portal pages

### DIFF
--- a/packages/openneuro-app/src/scripts/refactor_2021/search/search-params-ctx.tsx
+++ b/packages/openneuro-app/src/scripts/refactor_2021/search/search-params-ctx.tsx
@@ -21,6 +21,7 @@ export const SearchParamsProvider: FC<SearchParamsProviderProps> = ({
 export const removeFilterItem = setSearchParams => (param, value) => {
   switch (param) {
     case 'modality_selected':
+      break
     case 'datasetType_selected':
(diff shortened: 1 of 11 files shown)


packages/openneuro-app/src/scripts/refactor_2021/search/search-params-ctx.tsx
```diff
@@ -21,6 +21,7 @@ export const SearchParamsProvider: FC<SearchParamsProviderProps> = ({
 export const removeFilterItem = setSearchParams => (param, value) => {
   switch (param) {
     case 'modality_selected':
+      break
     case 'datasetType_selected':
     case 'datasetStatus_selected':
     case 'ageRange':
```
nellh, line 27: The behavior here would be a good unit test.

nellh (changes requested):
I think this could use some additional handling of routes, /search/modality or /search/modality/doesnotexist it could redirect to the default (MRI) page. Right now it just serves a blank page.

nellh (approved):
Will add the test in a new PR since it depends on some test fixes on master.

PR #2179, opened 2021-06-24
Search query refactor

david-nishi:
# Major Changes
- Refactor redesign facet filters and sorters to use the Elasticsearch Query DSL.
- Update elasticsearch dataset mapping with new filterable fields.

# Minor Changes
- Reset search params when running keyword search from landing page.

PR #2183, opened 2021-06-25
Front page and Search/portal updates

thinknoack:
Closes gitlab issues:  118, 223, 229, 221,234, 228, 242, 240, 227


packages/openneuro-components/src/mock-content/front-page-content.jsx
```diff
@@ -25,9 +26,9 @@ export const frontPage = {
   pageDescription: (
     <h2>
       A free and open platform for validating and sharing BIDS-compliant{' '}
-      <a href="/search/mri">MRI</a>, <a href="/search/pet">PET</a>,{' '}
-      <a href="/search/meg">MEG</a>, <a href="/search/eeg">EEG</a>, and{' '}
-      <a href="/search/asl">iEEG</a> data
+      <Link to="/search/mri">MRI</Link>, <Link to="/search/pet">PET</Link>,{' '}
+      <Link to="/search/meg">MEG</Link>, <Link to="/search/eeg">EEG</Link>, and{' '}
+      <Link to="/search/asl">iEEG</Link> data
```
nellh: These should use the full URL instead of the redirect, `/search/modality/eeg` to avoid search engines confusing the canonical version.

nellh (changes requested):
- [x] 118
- [x] 223
- [x] 229
- [x] 221 You can't x out of the modality yet? Is that intentional?
- [x] 234
- [x] 228
- [x] 242
- [x] 240
- [x] 227

Ran into one weird thing but not caused by this branch.

image

Search with no input leads to that.

packages/openneuro-components/src/mock-content/front-page-content.jsx
david-nishi: I'll handle the empty keyword chiclet and x-ing out the modality along with the change requests to #2179.

nellh (approved):

PR #2194, opened 2021-06-30
fix(components): Re-enable activity slider with workarounds

nellh:
This gets ActivitySlider working again - eventually found that optimizeDeps was just not being read and when it was added to the inline config this gets the module working until it's updated with ESM support.


nellh:
Test failure here is fixed in #2192

thinknoack (approved):

thinknoack:
One test failing but the activity slider / carousel stuff is good.

PR #2198, opened 2021-07-01
Pre refactor dataset

thinknoack:
updates refactor routes to use a styled container component for datalad/dataset/dataset.jsx


david-nishi (changes requested):
Lost some styles/functionality on the metadata page.

david-nishi (approved):

PR #2205, opened 2021-07-06
adding form submit to getupdates

thinknoack:


packages/openneuro-components/src/front-page/GetUpdates.tsx
```diff
@@ -2,23 +2,95 @@ import React from 'react'
...
+          setMode(
+            result.data.subscribeToNewsletter ? modes.SUCCESS : modes.ERROR,
+          )
+        }
+      })
+      break
+  }
+}
+
+export const getText = mode => {
```
david-nishi: Super minor, this doesn't need to be exported.

packages/openneuro-components/src/front-page/GetUpdates.tsx
```diff
@@ -2,23 +2,95 @@ import React from 'react'
...
+      <div className="get-updates-label">
+        <h2>{headingText}</h2>
+        <h3>{messageText}</h3>
+      </div>
+      <form
+        onSubmit={submitHandler(subscribe, mode, setMode, value)}
+        className="get-updates-form">
+        <>
+          <div className="">{}</div>
+          {mode === modes.ERROR || mode === modes.SUCCESS ? null : (
```
david-nishi:
```suggestion
          {mode === modes.GET && (
```
I think this would be more readable.

david-nishi (changes requested):

david-nishi (approved):

PR #2207, opened 2021-07-06
Results conditional

thinknoack:


david-nishi:
I found an error i made in the modality search page, so i corrected it and moved the duplicate logic for that and the new conditional in FilterBlock to the search-params-ctx.

david-nishi (approved):

ckrountree:
Closes Gitlab 264

PR #2213, opened 2021-07-09
1136 snapshot deprecation

david-nishi:
# Major Changes
- Adds fields `deprecated`, `deprecatedBy`, `deprecatedAt`, and `deprecatedFor` to snapshot.
- Adds `deprecateSnapshot resolver.


packages/openneuro-server/src/datalad/snapshots.js
```diff
@@ -248,12 +249,28 @@ export const getSnapshot = (datasetId, tag) => {
...
-        return snapshot
+        return {
+          ...body,
+          created,
+          hexsha,
+          deprecated,
+          deprecatedAt,
+          deprecatedBy: deprecatedBy
+            ? await User.findOne({ id: deprecatedBy })
+            : deprecatedBy,
```
nellh: This can use the implicit resolver since it's just the user id, e.g. it should work if you just return the ID because of the type definition in the schema.

packages/openneuro-server/src/datalad/snapshots.js
```diff
@@ -248,12 +249,28 @@ export const getSnapshot = (datasetId, tag) => {
...
+          deprecatedFor,
+        }
```
nellh: This is going to conflict with #2212... maybe we should make it a separate type and resolver so that a query that does not use these fields can avoid a round trip to Mongo?

nellh (approved):

PR #2214, opened 2021-07-09
200 modality aggregation queries

david-nishi:
## Major Changes
- Make publicDatasetsCounts and participantsCounts filterable by modality.
- Connect AggregateCounts on modality portals to data.


packages/openneuro-app/src/scripts/redesign.tsx
```diff
@@ -3,17 +3,17 @@ import Uploader from './uploader/uploader.jsx'
...
       <SearchParamsProvider>
-        <UserModalParamsProvider>
+        <UserModalOpenProvider>
           <HeaderContainer />
           <Routes_REFACTOR />
-        </UserModalParamsProvider>
+        </UserModalOpenProvider>
       </SearchParamsProvider>
     </Uploader>
   )
```
nellh: This got fixed on master.

nellh (approved):
One merge issue because the same change was made on master, otherwise working great. We may need to profile it a bit with the real database as the extra aggregate looks potentially heavy but I don't expect it to be a major issue.

PR #2216, opened 2021-07-09
fix(ssr): Don't remove accessToken cookie from the SSR response, only while rendering

nellh:
This cookie was accidentally unset on any SSR request instead of just for the duration of render.


nellh:
This has some negative consequences, working on an new version that selects the CSS during the server render.

PR #2218, opened 2021-07-12
Add portal redirects to ingress.

david-nishi:


nellh (approved):
This looks correct but I think we may need to make another change the CloudFront configuration to support it. Will merge and we can review on staging.

PR #2225, opened 2021-07-14
Pet summary fields

david-nishi:
## Major Changes
- Add pet specific fields from validator to Summary in mongo.
- Display pet fields in in summary section of dataset page (draft & snapshot).
- Make pet fields filterable on global and PET search pages.

## Minor Changes
- Fix no-snapshot datasets in search results.

### DIFF
--- a/packages/openneuro-components/src/search-page/SearchResultItem.tsx
+++ b/packages/openneuro-components/src/search-page/SearchResultItem.tsx
@@ -107,6 +107,7 @@ export const SearchResultItem = ({ node, profile }: SearchResultItemProps) => {
   const datasetId = node.draft.id
   const numSessions = summary?.sessions.length > 0 ? summary.sessions.length : 1
   const numSubjects = summary?.subjects.length > 0 ? summary.subjects.length : 1
+  const noSnapshots = !!node.snapshots
 
   const accessionNumber = (
     <span className="result-summary-meta">
@@ -150,24 +151,19 @@ export const SearchResultItem = ({ node, profile }: SearchResultItemProps) => {
 
   const dateAdded = formatDate(node.created)
   const dateAddedDifference = formatDistanceToNow(parseISO(node.created))
+  const date = noSnapshots
+    ? node.created
+    : node.snapshots[node.snapshots.length - 1].created
+  const dateUpdated = formatDate(date)
+  const dateUpdatedDifference = formatDistanceToNow(parseISO(date))
 
-  let lastUpdatedDate
-  if (node.snapshots.length) {
-    const dateUpdated = formatDate(
-      node.snapshots[node.snapshots.length - 1].created,
-    )
-    const dateUpdatedDifference = formatDistanceToNow(
-      parseISO(node.snapshots[node.snapshots.length - 1].created),
-    )
-
-    lastUpdatedDate = (
-      <div className="updated-date">
-        <span className="divider">|</span>
-        <span>Updated: </span>
-        {dateUpdated} - {dateUpdatedDifference} ago
-      </div>
-    )
-  }
+  const lastUpdatedDate = node.snapshots?.length ? (
+    <div className="updated-date">
+      <span className="divider">|</span>
+      <span>Updated: </span>
+      {dateUpdated} - {dateUpdatedDifference} ago
+    </div>
+  ) : null
 
   const uploader = (
     <div className="uploader">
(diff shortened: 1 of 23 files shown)


ckrountree:
Gitlab 213 and 252

packages/openneuro-components/src/search-page/SearchResultItem.tsx
```diff
@@ -150,24 +151,19 @@ export const SearchResultItem = ({ node, profile }: SearchResultItemProps) => {
...
-      </div>
-    )
-  }
+  const lastUpdatedDate = node.snapshots?.length ? (
+    <div className="updated-date">
+      <span className="divider">|</span>
+      <span>Updated: </span>
+      {dateUpdated} - {dateUpdatedDifference} ago
+    </div>
+  ) : null
```
nellh: Looks like this accidentally reverts a fix here for when draft only datasets are resolved.

packages/openneuro-server/src/models/summary.ts
```diff
@@ -25,6 +34,7 @@ const summarySchema = new Schema({
   totalFiles: Number,
   size: Number,
   dataProcessed: Boolean,
+  pet: Object,
```
nellh: The Mongoose schema should also specify the types.

packages/openneuro-server/src/graphql/schema.js
```diff
@@ -489,6 +489,15 @@ export const typeDefs = `
     size: BigInt!
     totalFiles: Int!
     dataProcessed: Boolean
+    pet: PetSpecificFields
+  }
+
+  type PetSpecificFields {
```
nellh: In isolation, this type name isn't very descriptive. SummaryPetFields?

packages/openneuro-server/src/graphql/resolvers/dataset.js
```diff
@@ -23,10 +24,22 @@ import { getDatasetWorker } from '../../libs/datalad-service.js'
...
+  try {
+    return await checkDatasetRead(id, user, userInfo).then(() => {
+      return datalad.getDataset(id)
+    })
+  } catch (err) {
+    // Don't throw the error if this is for a dataset search.
+    if (ignoreInaccessibleDatasets && err.message === states.READ.errorMessage)
+      return null
+    else throw err
+  }
```
nellh: I don't think the flag here makes sense, instead the error can be caught in the search resolver rather than overloading this one with search specific behavior. This also violates the signature for resolvers, which is `fieldName: (parent, args, context, info) => data`.

nellh (changes requested):

nellh:
One other piece of feedback, looks like this mostly duplicates #2223 but it's missing these new fields on SummaryInput so the validator output isn't actually saved.

david-nishi:
Rebased onto #2223 changes.

packages/openneuro-server/src/graphql/resolvers/dataset-search.js
```diff
@@ -35,24 +35,27 @@ export const decodeCursor = cursor =>
...
-    edges: body.hits.hits
-      .map(hit => ({
-        node: childResolvers.dataset(
-          null,
-          { id: hit._source.id },
-          { user, userInfo },
-        ),
-      }))
-      .filter(edge => edge !== null), // remove datasets that user does not have permissions for
+    edges: allowedEdges,
```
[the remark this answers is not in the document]
nellh: @david-nishi Looks like this still needs to be addressed.

packages/openneuro-server/src/graphql/resolvers/dataset-search.js
```diff
@@ -44,15 +45,19 @@ export const elasticRelayConnection = (
...
+      try {
+        const node = await childResolvers.dataset(
           null,
           { id: hit._source.id },
           { user, userInfo },
-        ),
-      }))
-      .filter(edge => edge !== null), // remove datasets that user does not have permissions for
+        )
+        return { node }
```
nellh: This can just return the dataset promise instead of wrapping it in another async function.

packages/openneuro-server/src/graphql/resolvers/dataset.js
```diff
@@ -23,8 +23,8 @@ import { getDatasetWorker } from '../../libs/datalad-service.js'
 import { getDraftHead } from '../../datalad/dataset.js'
 import { getFileName } from '../../datalad/files.js'
 
-export const dataset = (obj, { id }, { user, userInfo }) => {
-  return checkDatasetRead(id, user, userInfo).then(() => {
+export const dataset = async (obj, { id }, { user, userInfo }) => {
+  return await checkDatasetRead(id, user, userInfo).then(() => {
```
nellh: This is also an extra async/await. It could be refactored to use await instead of a promise chain within the await.

nellh (changes requested):

nellh (approved):

PR #2227, opened 2021-07-14
Grammatical cleanup and iEEG-sub removals (271)

ckrountree:


packages/openneuro-components/src/search-page/SearchResultItem.tsx
```diff
@@ -123,7 +123,7 @@ export const SearchResultItem = ({ node, profile }: SearchResultItemProps) => {
 
   const ages = (
     <span className="result-summary-meta">
-      <strong>Participants Ages: </strong>
+      <strong>Participant's Ages: </strong>
```
david-nishi: I think this would either be "Participants' Ages" or "Participant Ages".

david-nishi (approved):

PR #2232, opened 2021-07-15
Uploader

thinknoack:
adds the uploader component to the new site as well as styles to accommodate. This will be refactored in the future.


david-nishi (changes requested):
Couple notes (don't need to be addressed in this PR):
- There isn't an indicator for upload progress, success, or failure for the redesign implementation yet.
- The dataset does not show up in search after upload until the indexer is triggered.

nellh:
>     * The dataset does not show up in search after upload until the indexer is triggered.
For this, maybe we can just add the document (or a placeholder if not all info is available) after the initial upload?

nellh (changes requested):
I think this really needs some kind of progress, even if it's only inside the modal.

The toast styles might need to be imported earlier in the tree for the redesign as well.

ckrountree:
> I think this really needs some kind of progress, even if it's only inside the modal.
> 

We will definitely need a progress bar of some kind.  Let's do the MVP here and then address a fuller indicator post-beta.

david-nishi (approved):

thinknoack:
> I think this really needs some kind of progress, even if it's only inside the modal.
> 
> The toast styles might need to be imported earlier in the tree for the redesign as well.

Just for clarification. 

Currently when a user uploads a dataset, the "UPLOAD" link changes to "View Details" that pops up a modal with a progress bar and the current files being uploaded. Not sure if there is a success/fail notice though. There is no longer a upload progress bar in the navbar. And we may want to add toast notification.

PR #2247, opened 2021-07-19
Progress bar toasts

thinknoack:
adds styling for toasts and moved progress bar to header for visibility outside of the modal


nellh (approved):
Tested in Chrome/Firefox/Safari and this is a little odd for a very small dataset (page scrolling down and up quickly) but I think it works for most uploads.

PR #2258, opened 2021-07-21
Search pagination

david-nishi:
## Major Changes
- Enable "Load More" button on search.

### DIFF
--- a/packages/openneuro-app/src/client.jsx
+++ b/packages/openneuro-app/src/client.jsx
@@ -28,7 +28,14 @@ ReactDOM.render(
               fields: {
                 edges: {
                   keyArgs: ['query', 'datasetType', 'datasetStatus', 'sortBy'],
-                  merge: (existing = [], incoming, options) => {
+                  merge: (existing = [], incoming) => {
+                    existing = existing.filter(x => x)
+                    const existingIds = existing.map(({ node }) => node.__ref)
+                    incoming = incoming
+                      .filter(edge =>
+                        existingIds.includes(edge?.node.__ref) ? null : edge,
+                      )
+                      .filter(x => x)
                     return [...existing, ...incoming]
                   },
                 },
--- a/packages/openneuro-app/src/client.jsx
+++ b/packages/openneuro-app/src/client.jsx
@@ -29,8 +29,6 @@ ReactDOM.render(
                 edges: {
                   keyArgs: ['query', 'datasetType', 'datasetStatus', 'sortBy'],
                   merge: (existing = [], incoming, options) => {
-                    console.log('MERGE: DatasetConnection')
-                    console.log({ existing, incoming }, options)
                     return [...existing, ...incoming]
                   },
                 },
--- a/packages/openneuro-app/src/client.jsx
+++ b/packages/openneuro-app/src/client.jsx
@@ -14,6 +14,7 @@ import analyticsWrapper from './scripts/utils/analytics'
 import { version } from './lerna.json'
 import { config } from './scripts/config'
 import * as gtag from './scripts/utils/gtag'
+import { relayStylePagination } from '@apollo/client/utilities'
 
 gtag.initialize(config.analytics.trackingIds)
 
@@ -24,21 +25,9 @@ ReactDOM.render(
         clientVersion: version,
         cache: new InMemoryCache({
           typePolicies: {
-            DatasetConnection: {
+            Query: {
               fields: {
-                edges: {
-                  keyArgs: ['query', 'datasetType', 'datasetStatus', 'sortBy'],
-                  merge: (existing = [], incoming) => {
-                    existing = existing.filter(x => x)
-                    const existingIds = existing.map(({ node }) => node.__ref)
-                    incoming = incoming
-                      .filter(edge =>
-                        existingIds.includes(edge?.node.__ref) ? null : edge,
-                      )
-                      .filter(x => x)
-                    return [...existing, ...incoming]
-                  },
-                },
+                advancedSearch: relayStylePagination(),
               },
             },
           },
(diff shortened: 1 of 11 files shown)


packages/openneuro-app/src/scripts/datalad/dashboard/datasets/dataset-query.jsx
```diff
@@ -28,8 +28,9 @@ export const updateQuery = (previousResult, { fetchMoreResult }) => {
  * Load additional datasets based on next data cursor
  * @param {object} data Next data cursor
  * @param {function} fetchMore Apollo fetchMore function from the original query
+ * @param {number} version Apollo client version.
  */
-const loadMoreRows = (data, fetchMore) => {
+export const loadMoreRows = (data, fetchMore, version = 2) => {
```
nellh: There's only Apollo 3.0 clients so this doesn't seem necessary.

packages/openneuro-app/src/scripts/refactor_2021/search/search-container.tsx
```diff
@@ -92,11 +92,22 @@ const SearchContainer: FC<SearchContainerProps> = ({ portalContent }) => {
...
     useSearchResults()
+  const loadMore = loading
+    ? () => {}
+    : () =>
+        fetchMore({
+          variables: {
+            // ...variables,
+            cursor: data?.datasets?.pageInfo.endCursor,
+          },
+        }).then((...args) => console.log(args))
```
nellh: Looks like an extra console.log and both of these branches should return the same type (loadMore: void seems correct).

packages/openneuro-app/src/scripts/refactor_2021/search/use-search-results.tsx
```diff
@@ -285,6 +286,7 @@ export const useSearchResults = () => {
     errorPolicy: 'ignore',
     // fetchPolicy is workaround for stuck loading bug (https://github.com/apollographql/react-apollo/issues/3270#issuecomment-579614837)
     // TODO: find better solution
-    fetchPolicy: 'no-cache',
+    fetchPolicy: 'cache-and-network',
+    nextFetchPolicy: 'cache-first',
```
nellh, line 290: I think as long as we're using the anonymous SSR -> authenticated client renders this is the right strategy but we might want to revisit if we re-enable authenticated SSR to save a query.

packages/openneuro-app/src/client.jsx
```diff
@@ -22,8 +22,28 @@ ReactDOM.render(
...
+                edges: {
+                  keyArgs: ['query', 'datasetType', 'datasetStatus', 'sortBy'],
+                  merge: (existing = [], incoming) => {
+                    existing = existing.filter(x => x)
+                    const existingIds = existing.map(({ node }) => node.__ref)
+                    incoming = incoming
+                      .filter(edge =>
+                        existingIds.includes(edge?.node.__ref) ? null : edge,
+                      )
+                      .filter(x => x)
```
nellh: The fact this is required means this is still not correctly using keying the cached objects.

packages/openneuro-app/src/scripts/refactor_2021/search/search-container.tsx
```diff
@@ -176,9 +187,9 @@ const SearchContainer: FC<SearchContainerProps> = ({ portalContent }) => {
             <SearchResultsList items={resultsList} profile={profile} />
             {/* TODO: make div below into display component. */}
             <div className="grid grid-nogutter" style={{ width: '100%' }}>
-              {resultsList.length == 0 || resultsList.length < 25 ? null : (
+              {resultsList.length == 0 || numResultsRecieved < 25 ? null : (
```
nellh: Minor but hard coding the query size here is a bit brittle.

nellh (changes requested):

nellh:
Resolving some of my own feedback here. Simplifying to use the implicit merge on the DatasetEdge type and the generic relay pagination type policy allows for more reuse of these fields.

packages/openneuro-app/src/client.jsx
david-nishi: Thanks for your changes here!

packages/openneuro-server/src/graphql/resolvers/dataset-search.js
```diff
@@ -40,12 +40,14 @@ export const decodeCursor = cursor =>
 export const elasticRelayConnection = (
   { body },
   id,
+  size,
   childResolvers = { dataset },
   user = null,
   userInfo = null,
 ) => {
+  console.log(body)
```
nellh: Extra console.log

nellh (changes requested):
One test failure, this test probably just needs the test data updated.

```
dataset search resolvers › elasticRelayConnection() › returns a relay cursor for ApiResponse with results
    expect(received).toMatchObject(expected)
    - Expected  - 11
    + Received  +  5
      Object {
        "edges": Array [
          Object {
    -       "node": Object {
    -         "id": "testdataset1",
    -       },
    +       "node": Promise {},
          },
          Object {
    -       "node": Object {
    -         "id": "testdataset2",
    -       },
    +       "node": Promise {},
          },
          Object {
    -       "node": Object {
    -         "id": "testdataset3",
    -       },
    +       "node": Promise {},
          },
        ],
        "pageInfo": Object {
          "count": 10,
    -     "endCursor": "W251bGwsbnVsbF0=",
    -     "hasNextPage": true,
    +     "endCursor": null,
    +     "hasNextPage": false,
          "hasPreviousPage": false,
          "startCursor": null,
        },
      }
      89 |       )
      90 |       connection.edges = await Promise.all(connection.edges)
    > 91 |       expect(connection).toMatchObject(resultsRelayConnection)
         |                          ^
      92 |     })
      93 |   })
      94 | })
      at Object.<anonymous> (src/graphql/resolvers/__tests__/dataset-search.spec.js:91:26)
```

nellh (approved):

PR #2267, opened 2021-07-26
adding feedback form

thinknoack:
initial functionality for feedback via google forms


nellh (changes requested):

thinknoack:
added some default styles and interactions. ready for re-review

packages/openneuro-app/src/scripts/refactor_2021/redesign-form.jsx
```diff
@@ -0,0 +1,270 @@
...
+    }
+  }
+  myRequest = url => {
+    let response
+    try {
+      response = fetch(url, { method: 'POST' })
+    } catch (e) {
+      response = e
+    }
+    console.log(response)
```
nellh:
```suggestion
    fetch(url, { method: 'POST' })
```
Fetch returns a promise so this never throws.

packages/openneuro-app/src/scripts/refactor_2021/redesign-form.jsx
```diff
@@ -0,0 +1,270 @@
...
+
+  handleSubmit = e => {
+    e.preventDefault()
+    const id = '1GhgmqcdiPH1HPf8ZGqoQo33Lluds2IneAsn4U9Zv03M'
+    const formUrl = 'https://docs.google.com/forms/d/' + id + '/formResponse?'
+
+    let q = new URLSearchParams(this.state.entryFields).toString()
+
+    this.myRequest(formUrl + q)
+    console.log(this.state.entryFields)
```
nellh:
```suggestion
```

nellh (approved):
Couple suggestions but otherwise looks good.

thinknoack:
@nell committed suggestions and tested. all is working - i'll leave this here for @ckrountree  to review.  Ty

ckrountree:
Couple of comments:

1. The Feedback button may need a stronger call-to-action color.  I like the cleanliness and unobtrusiveness but it may also get lost.  Fine line. I leave this decision to you.
2. For the drop-down, I think that "OK" and "Soso" are redundant.  I'd also spell out "Okay". Trying to avoid a "bad" connotation, I might change the wording to "Poor" - gets the point across without being as strong.
Screen Shot 2021-08-06 at 1 00 59 PM

3. I would make it clearer what the two places you are asking about are:
first found modality portal

4. I would suggest adding an additional question to assess overall experience (not just Search).  Something with the tone such as, "How intuitive did you find the redesign functionality?" and then use the same _Great, Good, Okay, Poor_ options used in the Search drop-down. Maybe provide a follow-up if rating is not _Great_: "If not great, what was not intuitive?"

ckrountree:
Copied from [PR 2271](https://github.com/OpenNeuroOrg/openneuro/pull/2271#issuecomment-894494890): 
> 2\. For the drop-down, I think that "OK" and "Soso" are redundant.  I'd also spell out "Okay". Trying to avoid a "bad" connotation, I might change the wording to "Poor" - gets the point across without being as strong.


this is one eric wanted and he definitely didn't want good/bad. More like "The redesign is great, I find it very easy to use" and "I think the redesign is good overall".

I just need help with statements like that. They dont have to be good or bad.

thinknoack:
> 4\. I would suggest adding an additional question to assess overall experience (not just Search).  Something with the tone such as, "How intuitive did you find the redesign functionality?" and then use the same _Great, Good, Okay, Poor_ options used in the Search drop-down. Maybe provide a follow-up if rating is not _Great_: "If not great, what was not intuitive?"

I am a little confused by this. The dropdown is about the redesign. Do you want to add another that is for Search?

ckrountree:
> > 4. I would suggest adding an additional question to assess overall experience (not just Search).  Something with the tone such as, "How intuitive did you find the redesign functionality?" and then use the same _Great, Good, Okay, Poor_ options used in the Search drop-down. Maybe provide a follow-up if rating is not _Great_: "If not great, what was not intuitive?"
> 
> I am a little confused by this. The dropdown is about the redesign. Do you want to add another that is for Search?

Not necessarily.  You have the question "When searching for a dataset were you able to find relevant results?" in the form.  If we're asking about the ease of searching, I'd also like to know how intuitive the UI is for users.  That said, that could be accomplished if you decide to go with a rating drop-down more like:

**Which of the following best describes your experience with the redesign today?**
- Exciting and easy to navigate
- Visually pleasing, not quite as intuitive
- Couldn’t achieve my goal

 Then follow up with a Q if the user selects _Visually pleasing, not quite as intuitive_ or _Couldn’t achieve my goal_ (what was not intuitive, what was your goal)

thinknoack:
> > > 1. I would suggest adding an additional question to assess overall experience (not just Search).  Something with the tone such as, "How intuitive did you find the redesign functionality?" and then use the same _Great, Good, Okay, Poor_ options used in the Search drop-down. Maybe provide a follow-up if rating is not _Great_: "If not great, what was not intuitive?"
> > 
> > 
> > I am a little confused by this. The dropdown is about the redesign. Do you want to add another that is for Search?
> 
> Not necessarily. You have the question "When searching for a dataset were you able to find relevant results?" in the form. If we're asking about the ease of searching, I'd also like to know how intuitive the UI is for users. That said, that could be accomplished if you decide to go with a rating drop-down more like:
> 
> **Which of the following best describes your experience with the redesign today?**
> 
> * Exciting and easy to navigate
> * Visually pleasing, not quite as intuitive
> * Couldn’t achieve my goal
> 
> Then follow up with a Q if the user selects _Visually pleasing, not quite as intuitive_ or _Couldn’t achieve my goal_ (what was not intuitive, what was your goal)

I believe this has been accomplished with the changes I added to the form. Let's discuss when you have had a chance to review.

ckrountree:
Do we want to add "_Good overall experience_"?  It may seem redundant with the first option. I'll leave that to your discretion:
Screen Shot 2021-08-09 at 12 30 28 PM

I might be more explicit here: "_If the new design did not feel as intuitive or if you were unable to achieve your goal(s), please explain further._"
Screen Shot 2021-08-09 at 12 30 40 PM

Misspelling (`were`):

Screen Shot 2021-08-09 at 12 32 55 PM

ckrountree (approved):

PR #2271, opened 2021-08-05
Onpet feedback updates

thinknoack:
Gitlab- #295, #297, #306, #309, #310, #315, #314

### DIFF
--- a/packages/openneuro-app/src/sass/pre-refactor-with-wrapper-main.scss
+++ b/packages/openneuro-app/src/sass/pre-refactor-with-wrapper-main.scss
@@ -1220,6 +1220,10 @@ $primary: $on-dark-aqua default;
     line-height: 1.4em;
   }
 
+  //
+  // Tooltips
+  // --------------------------------------------------
+
   .alert {
     padding: $alert-padding;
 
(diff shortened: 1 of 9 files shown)


ckrountree:
Couple of comments:

1. The Feedback button may need a stronger call-to-action color.  I like the cleanliness and unobtrusiveness but it may also get lost.  Fine line.
2. For the drop-down, I think that "OK" and "Soso" are redundant.  I'd also spell out "Okay". Trying to avoid a "bad" connotation, I might change the wording to "Poor" - gets the point across without being as strong.
Screen Shot 2021-08-06 at 1 00 59 PM

3. I would make it clearer what the two places you are asking about are:
first found modality portal

4. I would suggest adding an additional question to assess overall experience (not just Search).  Something with the tone such as, "How intuitive did you find the redesign functionality?" and then use the same _Great, Good, Okay, Poor_ options used in the Search drop-down. Maybe provide a follow-up if rating is not _Great_: "If not great, what was not intuitive?"

thinknoack:
> 2\. For the drop-down, I think that "OK" and "Soso" are redundant.  I'd also spell out "Okay". Trying to avoid a "bad" connotation, I might change the wording to "Poor" - gets the point across without being as strong.


this is one eric wanted and he definitely didn't want good/bad. More like "The redesign is great, I find it very easy to use" and "I think the redesign is good overall".

I just need help with statements like that. They dont have to be good or bad.

ckrountree:
> > 2. For the drop-down, I think that "OK" and "Soso" are redundant.  I'd also spell out "Okay". Trying to avoid a "bad" connotation, I might change the wording to "Poor" - gets the point across without being as strong.
> 
> this is one eric wanted and he definitely didn't want good/bad. More like "The redesign is great, I find it very easy to use" and "I think the redesign is good overall".
> 
> I just need help with statements like that. They dont have to be good or bad.

ckrountree:
###  Looks good:

- `My Dataset` link to account dropdown (315)
- Affiliate block links (297)
- Section to Study Type change (314)
- Hidden sub-modalities (ASL, dynamic, static) (309)
- Portal title tags (295)

### Still pending:

- Dataset tooltips (uploading datasets locally is failing for me) (306)
Screen Shot 2021-08-06 at 4 59 29 PM

- Activity label change NEWLY ADDED >>> RECENTLY PUBLISHED (310) - can't check because...:
Screen Shot 2021-08-06 at 4 30 56 PM

ckrountree:
Regression: only on portal pages (so NOT landing or global Search) - the upload drop-down is hidden:
Screen Shot 2021-08-09 at 12 43 13 PM


Activity is now visible and working:

Screen Shot 2021-08-09 at 2 08 03 PM

Icons have tooltips.

ckrountree:
Regression fixed! 
Screen Shot 2021-08-09 at 2 26 37 PM

ckrountree (approved):

PR #2286, opened 2021-08-12
Git backend service performance refactors

nellh:
The avoids some subprocess calls and simplifies some of the steps to serve git repositories. Generally drops get_dataset for working with paths instead in most places.


david-nishi (approved):
Awesome work @nellh!

PR #2302, opened 2021-08-20
linking modalities on datasets to portals

thinknoack:
Addresses Gitlab 333

### DIFF
--- a/packages/openneuro-components/src/dataset/ModalitiesMetaDataBlock.tsx
+++ b/packages/openneuro-components/src/dataset/ModalitiesMetaDataBlock.tsx
@@ -1,0 +1,38 @@
+import React from 'react'
+import Markdown from 'markdown-to-jsx'
+
+export interface ModalitiesMetaDataBlockProps {
+  heading: string
+  isMarkdown?: boolean
+  item: React.ReactNode | string[] | number
+  className?: string
+}
+
+export const ModalitiesMetaDataBlock = ({
+  heading,
+  isMarkdown,
+  item,
+  className,
+}: ModalitiesMetaDataBlockProps) => {
+  const arrayToMarkdown = arr => {
+    console.log(arr)
+    return arr
+      ? arr
+          .map(
+            element =>
+              ` * <a href="/search/${element.toLowerCase()}">${element}</a>\n`,
+          )
+          .join('')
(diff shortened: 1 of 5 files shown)


packages/openneuro-components/src/dataset/ModalitiesMetaDataBlock.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const ModalitiesMetaDataBlock = ({
+  heading,
+  isMarkdown,
+  item,
+  className,
+}: ModalitiesMetaDataBlockProps) => {
+  const arrayToMarkdown = arr => {
+    console.log(arr)
```
nellh: Extra console.log
```suggestion
```

nellh (changes requested):

packages/openneuro-components/src/dataset/ModalitiesMetaDataBlock.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isMarkdown?: boolean
+  item: React.ReactNode | string[] | number
+  className?: string
+}
+
+export const ModalitiesMetaDataBlock = ({
+  heading,
+  isMarkdown,
+  item,
+  className,
```
david-nishi: Since this component is specific to a single feature, i think we can just hardcode these props.

packages/openneuro-components/src/dataset/ModalitiesMetaDataBlock.tsx
```diff
@@ -0,0 +1,37 @@
...
+      : ''
+  }
+  return (
+    <div className={'dataset-meta-block ' + className}>
+      <h2 className="dmb-heading">{heading}</h2>
+      {isMarkdown ? (
+        <Markdown>{arrayToMarkdown(item) || 'N/A'}</Markdown>
+      ) : (
+        item || 'N/A'
+      )}
```
david-nishi: Instead of going through markdown-to-jsx, we could just use jsx. This would also let us use <Link/> from react-router-dom instead of anchor tags.

david-nishi (changes requested):

david-nishi (approved):

PR #2307, opened 2021-08-24
track My Datasets and My Bookmarks state in URL query params

david-nishi:
## Major Changes
- Clicking to / away from My Datasets updates the url query param mydatasets.
- Removing the My Datasets chiclet removes mydatasets from the url query params.
- Clicking to / away from My Bookmarks updates the url query param bookmarks.
- Removing the My Bookmarks chiclet removes bookmarks from the url query params.


thinknoack (approved):

ckrountree:
Gitlab 321

nellh (commented):
Maybe it would be better to just add these as URLs and detect them as routes? That way there's no URL manipulation logic needed in these search components.

thinknoack (changes requested):
removed approval to discuss @nellh comments

david-nishi:
Closed in favor of route param approach.

PR #2323, opened 2021-09-07
Publish dataset tool

thinknoack:


thinknoack:
also, moved tools into a component that can be shared by draft and snapshots

packages/openneuro-components/src/dataset/DatasetTools.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const DatasetTools = ({
+  hasEdit,
+  rootPath,
+  isPublic,
+  isSnapshot,
+}: DatasetToolsProps) => {
+  return (
+    <>
+      <>
```
david-nishi: Don't need two of these.

david-nishi (approved):

PR #2327, opened 2021-09-08
update git config url and worker

thinknoack:


thinknoack:
also, update to alert text and conditionals

david-nishi (approved):

PR #2330, opened 2021-09-15
Anon review

thinknoack:
Still TODO - add subscription or work on cache update to show changes when a reviewer is created or removed. 

update apollo cache
https://www.apollographql.com/docs/react/data/mutations/#the-update-function


david-nishi (approved):

packages/openneuro-server/src/graphql/resolvers/reviewer.ts
```diff
@@ -14,16 +14,26 @@ import { generateReviewerToken } from '../../libs/authentication/jwt.js'
  */
 export async function createReviewer(obj, { datasetId }, { user, userInfo }) {
   await checkDatasetAdmin(datasetId, user, userInfo)
-  const reviewer = new Reviewer({ datasetId, creator: user })
-  await reviewer.save()
+  let reviewer = new Reviewer({ datasetId, creator: user })
+  reviewer = await reviewer.save()
```
nellh:
```suggestion
  const reviewer = new Reviewer({ datasetId, creator: user })
  await reviewer.save()
```

nellh (changes requested):
Creating the reviewer doesn't make the URL available at any point. Delete works well.

PR #2335, opened 2021-09-20
fix: Increase dataset title and authors search priority and order score

nellh:
This raises the importance of title and authors fields on keyword matches and fixes the inverted score ordering that was giving some strange results when using the default sort.


ckrountree:
Addresses Gitlab 351, 352

david-nishi (approved):

david-nishi:
Code looks good, still needs testing on staging.

PR #2336, opened 2021-09-20
fix: Avoid first load css issues when enable/disable feature flags

nellh:
This avoids the first load served from cache issue that occurs when enabling or disabling the redesign pages.


david-nishi (approved):

ckrountree:
Resolves Gitlab 346

PR #2337, opened 2021-09-21
File tree styles

thinknoack:
style and cleanup for file tree.


packages/openneuro-app/src/scripts/refactor_2021/dataset/snapshot-container.tsx
```diff
@@ -338,17 +337,14 @@ const SnapshotContainer: React.FC<SnapshotContainerProps> = ({
             />
             <MetaDataBlock heading="License" item={description.License} />
             <MetaDataBlock
-              isMarkdown={true}
```
david-nishi, line 341: One more of these to remove on line 209.

david-nishi (changes requested):
Really like what you did with the file tools!

david-nishi (approved):

Issue #355, opened 2018-01-23
Support Schema.org Dataset metadata for indexing


nellh:
This is not yet included in Google search results but once enabled this would make datasets more discoverable with Google search and other crawlers using the schema.org data types.

http://schema.org/Dataset
https://developers.google.com/search/docs/data-types/dataset

chrisgorgo:
Isn't this a duplicate of #352?

chrisgorgo:
(great minds...)

nellh:
Oh, yes it is. I'll close this one. #352 should also keep Google's use of this data in mind for improving search results.

PR #2343, opened 2021-09-22
update text for when worker ID is undefined adding worker to clone dropdown

thinknoack:
resolves gitlab #355


nellh (changes requested):
This is easily fixed by including the worker value in the snapshot query, it's always available for the dataset.

nellh (approved):

PR #2356, opened 2021-09-24
Follow bookmark login

david-nishi:
## Major changes
- Add redirect option to client login.
- When a user clicks a dataset's follow or bookmark button when not logged in, the login modal is displayed. Once logged in, the user is taken back to the dataset page they were on and the button they clicked is triggered for them.


thinknoack:
Looks good. One thing I saw

If a user has already favorited the dataset, when logged out it redirects them back to the dataset and "unfollows".

thinknoack (approved):

packages/openneuro-server/src/libs/authentication/jwt.js
```diff
@@ -153,10 +154,13 @@ export const authenticate = (req, res, next) => {
 }
 
 export const authSuccessHandler = (req, res, next) => {
+  const redirectPath = req.query.state 
+    ? atob(req.query.state)
```
nellh: You can avoid adding atob (unmaintained package and obsolete in Node 16) like this:
```suggestion
    ? Buffer.from(req.query.state, 'base64').toString()
```

nellh (changes requested):

nellh (approved):

PR #2357, opened 2021-09-24
chore: Update datalad-service lockfile to resolve jsmin incompatiblity

nellh:
jsmin 2.2.2 as locked no longer builds with the upstream python container


thinknoack (approved):
I have uploaded, edited, and deleted with success.

PR #2358, opened 2021-09-24
fix: Support new summary field in bids-validator@1.8.4

nellh:


thinknoack (approved):
I have uploaded, edited, and deleted with success.

PR #2361, opened 2021-09-27
fix: Summary is not required, make toObject() call optional when undefined

### LINKED ISSUE #2359: My dashboard bugs
**Describe the bug**

Multiple users are now reporting that the "My dashboard" (https://openneuro.org/dashboard/datasets) panel is presenting type errors.

```
TypeError: Cannot read property 'node' of null
```

```
TypeError: this.props.datasets[t] is null
```

I believe these are users with datasets with no snapshots.

Will report additional context as I receive it.

nellh:
Fixes #2359


nellh:
Reproduction case is pretty simple, just need any dataset commit that doesn't any summary associated with it.

thinknoack (approved):

PR #2381, opened 2021-09-28
adding optional to summary to fix error for null value

thinknoack:


thinknoack:
I was getting an Error on the front page for primaryModality.

`Uncaught TypeError: Cannot read properties of null (reading 'primaryModality')`

david-nishi (approved):

PR #2405, opened 2021-10-04
fixing a few issues phone on the iphone -effecting all smaller screens

thinknoack:


david-nishi (approved):
Looks good. I'd like to double check the logged in features on staging too

PR #2411, opened 2021-10-05
fix(search): Use snapshot instead of draft for search result display

nellh:
This updates the search result page to use latestSnapshot instead of draft for most of its information.


thinknoack (approved):
looks good locally.

PR #2431, opened 2021-10-20
adding module to pretty print json

thinknoack:
fixes ticket https://gitlab.com/squishymedia/nimh/openneuropet/-/issues/395


nellh (changes requested):
This module is abandoned, having not received updates in 3+ years. I think it's too risky to include it. It also pulls in the rather large Flux package as a new dependency and some React 16 legacy shims.

nellh (approved):

PR #2448, opened 2021-10-26
hide invalid icon unless you have edit access

thinknoack:


nellh (approved):
Made one TypeScript fix for the stories, otherwise good.

PR #2449, opened 2021-10-26
fix: Cache participantCount query for one hour

nellh:
Don't run this query as often, this is causing some latency blocking other queries from running.

### DIFF
--- a/packages/openneuro-server/src/graphql/resolvers/snapshots.js
+++ b/packages/openneuro-server/src/graphql/resolvers/snapshots.js
@@ -68,7 +68,6 @@ export const deprecateSnapshot = async (
 }
 
 export const participantCount = (obj, { modality }) => {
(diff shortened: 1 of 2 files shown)


packages/openneuro-server/src/graphql/resolvers/snapshots.js
```diff
@@ -65,66 +67,73 @@ export const deprecateSnapshot = async (
...
-          },
-        ],
-      }
-    : queryHasSubjects
-  const aggregateResult = await DatasetModel.aggregate([
-    {
-      $match: {
-        public: true,
+export const participantCount = (obj, { modality }) => {
+  console.log(participantCount)
```
thinknoack:
```suggestion
```

thinknoack (approved):

PR #2463, opened 2021-11-04
chore: Cleanup extra dependencies from docz

nellh:
Drops a few large global dependencies since they are no longer required to generate the documentation site.


nellh:
Looking into the test issue here.

thinknoack (approved):

PR #2477, opened 2021-11-09
adding git and on datalad urls with perms

thinknoack:
adding readOnly and has Edit datalad urls to clone dropdown and download page.


thinknoack:
issue https://gitlab.com/squishymedia/nimh/openneuropet/-/issues/413

Issue #396, opened 2018-02-08
Client side archive status indicator


nellh:
The prototype from #353 does not report the progress for S3 files until they are fully downloaded. Ideally the status for each file would be available and an indicator for progress would be part of the UI for downloading.

chrisgorgo:
Could the download status be solved by passing the expected archive size (which we can calculate since no compression is used)? Or is this issue about something else?

nellh:
The idea here was to show which file and how far along you were downloading from S3 somewhere more visible than the network tab in Chrome or Firefox.

nellh:
We should address this with broader download improvements rather than fixing this for the specific case.

PR #2479, opened 2021-11-11
Test git history API formats and fix extra commas returned in references format

nellh:
Fixes a string concatenation issue in the git history API for datasets. Adds test coverage for the format this API returns beyond just checking the correct number of git log entries.

GitLab #396


thinknoack (approved):
looks good and tests pass

PR #2488, opened 2021-11-15
fix(search): Disable "relevance" sort when no query is set

nellh:
If no search filters have been set, disable the relevance sort and show "Newest" instead.

Adds test coverage for this behavior and some test utils for testing search parameters generally.


thinknoack:
Overall it looks good. There was one issue that is not happening on master.

The range/date filters causes an error if no other filter is present. 

filters effected:
Age of Participants
Number of Participants
Publication Date

Reproduce-
Visit search add participant age filter. Page is blank.
Select another filter first (keyword) use age filter. Page works as expected. 

I tested all the other filters (including pet specific) and they all seem to be working correctly. 

Screen Shot 2021-11-16 at 11 35 57 AM

thinknoack (approved):

PR #2489, opened 2021-11-16
Disable redesign flag

nellh:
This is a subset of the cleanup work disabling the redesign flag and cookie handling. This avoids serving the old CSS at all but doesn't drop most of the unused dependencies or components.


thinknoack (approved):
🙌

PR #2512, opened 2021-12-07
feat: Allow for version deprecation

nellh:
This adds API and frontend support for marking a version snapshot as deprecated. There's a new workflow that can be initiated as a dataset admin on any version.

image

From there you can specify an explanation.

image

Warning modal is displayed when visiting a version page.

image

This doesn't have any functional effect on usage of the version, it's only informational on the website and a field for the API.

Remaining:
- [x] Use banner instead of modal
- [x] Undo button
- [x] TypeScript build issue


effigies:
Is it possible to un-deprecate or change the deprecation reason? I can imagine cases where someone makes a mistake, changes their mind, or wants to provide more detail.

Another thing someone might want to do is "Deprecate this and all previous versions".

I would prefer a banner (similar to "This dataset is not yet published", possibly red background) to a modal.

nellh:
> Is it possible to un-deprecate or change the deprecation reason? I can imagine cases where someone makes a mistake, changes their mind, or wants to provide more detail.
> 
> Another thing someone might want to do is "Deprecate this and all previous versions".
> 
> I would prefer a banner (similar to "This dataset is not yet published", possibly red background) to a modal.

You can replace the reason but there isn't a way to remove it. I agree on the banner, I'll switch that over and add a button to remove it to the banner.

effigies:
Is it possible to deprecate the most recent version and have no valid versions? Is that a use case we want to support, or should the user create a valid version before being able to declare an old one invalid?

nellh:
> Is it possible to deprecate the most recent version and have no valid versions? Is that a use case we want to support, or should the user create a valid version before being able to declare an old one invalid?

Wouldn't it be useful to deprecate a snapshot if you uploaded it and found an error but didn't have the new one ready yet? I think that's a valid use case to inform someone downloading it has an issue even if a new one isn't ready yet.

effigies:
I'm kind of going back and forth on it. I guess the question is the time frame. Disallowing deprecation without replacement could encourage more timely replacement, but leaves a longer window between "known bad" and "advertised bad". On the other hand, if a dataset is perpetually left in a "known bad" state, that's not great either.

I assume it's simplest to allow any version to be deprecated, so that makes sense in the short term. Perhaps we should monitor if there are datasets with a deprecated latest version and revisit if this becomes a problem.

nellh:
image

Here's what it looks like as a banner. Undo icon is available if you have access to mark snapshots as deprecated.

effigies:
Looks great!

Issue #2529, opened 2022-01-19
DOC: Failures to set up dev environment


effigies:
Node environment:

```Shell
$ which node
/usr/local/lib/nodejs/node-v14.16.0-linux-x64/bin/node
$ node --version
v14.16.0
```

OpenNeuro:

```Shell
$ git describe
v4.3.2-1-gd1c805d4a
```

Yarn according to docs:

<details>
<summary><code>$ yarn install</code></summary>

```
 YN0000: ┌ Resolution step
➤ YN0002: │ @aws-sdk/client-s3@npm:3.33.0 doesn't provide @aws-sdk/signature-v4-crt (p3df12), requested by @aws-sdk/middleware-sdk-s3
➤ YN0002: │ @mdx-js/loader@npm:1.6.22 doesn't provide react (p12550), requested by @mdx-js/react
➤ YN0002: │ @octokit/rest@npm:16.43.2 doesn't provide @octokit/core (p84523), requested by @octokit/plugin-request-log
➤ YN0060: │ @openneuro/app@workspace:packages/openneuro-app provides react (ped6c8) with version 17.0.2, which doesn't satisfy what react-virtualized requests
➤ YN0060: │ @openneuro/app@workspace:packages/openneuro-app provides react-dom (pfa5dd) with version 17.0.2, which doesn't satisfy what react-virtualized requests
➤ YN0002: │ @openneuro/app@workspace:packages/openneuro-app doesn't provide rollup (p7a7fb), requested by rollup-plugin-polyfill-node
➤ YN0002: │ @openneuro/cli@workspace:packages/openneuro-cli doesn't provide graphql (p71dc4), requested by @apollo/client
➤ YN0002: │ @openneuro/client@workspace:packages/openneuro-client doesn't provide @babel/core (pfd9b4), requested by @babel/preset-typescript
➤ YN0002: │ @openneuro/components@workspace:packages/openneuro-components doesn't provide @babel/core (p2fde8), requested by @storybook/addon-essentials
➤ YN0002: │ @openneuro/components@workspace:packages/openneuro-components doesn't provide @babel/core (p00817), requested by babel-jest
➤ YN0002: │ @openneuro/components@workspace:packages/openneuro-components doesn't provide babel-loader (pa9189), requested by @storybook/addon-essentials
[...]

PR #2530, opened 2022-01-19
docs: Update dev setup documentation

nellh:
Fixes some developer setup documentation issues reported in #2529


README.md
```diff
@@ -16,29 +16,46 @@ OpenNeuro is a free and open platform for analyzing and sharing neuroimaging dat
...
 ```shell
-yarn install && yarn bootstrap
+yarn install
 ```
 
 You can run tests with `yarn test` at the top level of the project. For each package, `yarn test --watch` will interactively run the tests for changes since the last commit.
 
-Before starting up the services, you will need to copy the example `.env.example` file to `.env` and `config.env.example` to `config.env`. Many of the values are optional, and most that aren't have default values included in their `.example` file. The exception is `JWT_SECRET` in `config.env`, which you will need to set to a large random string.
+Before starting up the services, you will need to copy the example `.env.example` file to `.env` and `config.env.example` to `config.env`. Many of the values are optional, and most that aren't have default values included in their `.example` file. Required values below:
+
```
rwblair, line 25: I had to set `CRN_SERVER_URL` in `config.env` to `http://localhost:9876` in order to get front end to fully load, may be worth mentioning here, or setting as default in example.

nellh: Went ahead and added it as the default value.

PR #2532, opened 2022-01-24
fix(app): Link to snapshot download page from snapshots

### LINKED ISSUE #2526: Snapshot download link does not link to snapshot download page
**Describe the bug**
This is linking to the draft download page.

**To Reproduce**
1. Visit any snapshot
2. Click download

nellh:
Replaces DatasetTools isSnapshot with a snapshot id to allow for this link to work.

Fixes #2526


packages/openneuro-components/src/dataset/DatasetTools.tsx
```diff
@@ -89,7 +90,9 @@ export const DatasetTools = ({
         </Tooltip>
       )}
       <span>
-        <Link className="dataset-tool" to={`/datasets/${datasetId}/download`}>
+        <Link
+          className="dataset-tool"
+          to={`/datasets/${datasetId}/versions/${snapshotId}/download`}>
```
rwblair: If I'm logged in and navigate to a published dataset I own it defaults to the draft view and the snapshotId is undefined, Following through with the download link the api produces a 5xx error.

rwblair (changes requested):

packages/openneuro-components/src/dataset/DatasetTools.tsx
nellh: Added a fix for this with new test coverage for the issue.

rwblair (approved):

PR #2544, opened 2022-02-14
ENH: Show orientation markers on NIfTI images

effigies:
From my reading of https://github.com/rii-mango/Papaya/wiki/Configuration#display-parameters, this should show L/R, P/A, and I/S markers on the images, to help users determine whether they are viewing as radiological vs neurological, and detect orientation errors.


nellh:
Gave it a try and it works as expected.

PR #2553, opened 2022-02-18
fix: Create pygit2.Signature at commit time

nellh:
Fixes an issue where commits were sometimes generated with an out of date commit time.


effigies (approved):
Seems reasonable.

PR #2574, opened 2022-03-16
refactor(app): Reorganize all refactor_2021 modules

nellh:
Drops refactor_2021 and moves all the associated modules.


packages/openneuro-app/src/scripts/datalad/fragments/freshdesk-widget.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { useCookies } from 'react-cookie'
-import { getProfile } from '../../refactor_2021/authentication/profile'
+import { getProfile } from '../../authentication/profile'
```
effigies, line 4: Should this file have gone into dataset/fragments instead of datalad/fragments?

nellh: Yeah, it probably shouldn't be in there. This component is only used in the error boundary component, so I think it should be next to that.

PR #2576, opened 2022-03-16
Show file size in dataset file tree

### LINKED ISSUE #2510: Show file sizes in file tree

Have a user frustrated with the inability to find some file metadata in the file tree, in particular the file size.

File size and mtime are commonly shown in file browsers. Could add one or both of these.

None.

**Additional context**

Freshdesk 1067.

nellh:
Adds the file size in human readable sizes as part of the download link tooltip.

Fixes #2510


rwblair (approved):
Just to double check my understanding, the only use of the File component is in scripts/dataset/files/file-tree.jsx and the new size field is passed to it when the file variable gets unpacked into its props?

nellh:
> Just to double check my understanding, the only use of the File component is in scripts/dataset/files/file-tree.jsx and the new size field is passed to it when the file variable gets unpacked into its props?

That's correct.

PR #2581, opened 2022-03-29
feat(app): Display content size instead of validator summary size for datasets

nellh:
Uses the new API implemented in #2578 to show a better estimate of dataset size in bytes.

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/snapshot-container.tsx
+++ b/packages/openneuro-app/src/scripts/dataset/snapshot-container.tsx
@@ -111,7 +111,7 @@ const SnapshotContainer: React.FC<SnapshotContainerProps> = ({
           <div className="col col-8 col-lg">
             {summary && (
               <DatasetHeaderMeta
-                size={summary.size}
+                size={snapshot.size}
                 totalFiles={summary.totalFiles}
                 datasetId={datasetId}
               />
--- a/packages/openneuro-components/src/search-page/SearchResultItem.tsx
+++ b/packages/openneuro-components/src/search-page/SearchResultItem.tsx
@@ -165,7 +165,7 @@ export const SearchResultItem = ({
   const size = (
     <span className="result-summary-meta">
       <strong>Size: </strong>
-      <span>{bytes(node?.latestSnapshot?.size)}</span>
+      <span>{bytes(node?.latestSnapshot?.size) || "unknown" }</span>
     </span>
   )
   const files = (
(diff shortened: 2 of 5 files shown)


packages/openneuro-components/src/search-page/SearchResultItem.tsx
```diff
@@ -251,8 +250,7 @@ export const SearchResultItem = ({
     <Tooltip
       tooltip="Visable to all viewers"
       flow="up"
-      className="result-icon result-publlic-icon"
-    >
+      className="result-icon result-publlic-icon">
```
effigies, line 253: Just noticing this is "publlic" (two `l`s). I'm not finding anywhere else in the repo that uses "result-publlic-icon" or "result-public-icon", so unclear if this is a typo (and if so, if it has any meaningful effect).

nellh: Probably a class in the previous design that can just be removed.

packages/openneuro-app/src/scripts/dataset/snapshot-container.tsx
```diff
@@ -111,7 +111,7 @@ const SnapshotContainer: React.FC<SnapshotContainerProps> = ({
           <div className="col col-8 col-lg">
             {summary && (
               <DatasetHeaderMeta
-                size={summary.size}
+                size={snapshot.size}
                 totalFiles={summary.totalFiles}
```
effigies, line 115: Need `snapshot.totalFiles`.

packages/openneuro-components/src/search-page/SearchResultItem.tsx
```diff
@@ -164,7 +165,7 @@ export const SearchResultItem = ({
   const size = (
     <span className="result-summary-meta">
       <strong>Size: </strong>
-      <span>{bytes(summary?.size)}</span>
+      <span>{bytes(node?.latestSnapshot?.size)}</span>
```
rwblair: For the curious reader:
```
> bytes(x?.bad)
null
```

effigies: Good `Maybe` monad.

nellh: Nice, so that's mostly fine (it will render "Size: ").

effigies: Would this be a reasonable failure mode?

```suggestion
      <span>{bytes(node?.latestSnapshot?.size) || "unknown" }</span>
```

PR #2583, opened 2022-03-30
Rework dataset page for tabbed tools

nellh:
This restructures all the dataset tools into tabs.

image

Couple things left to do:
- [x] Display the active tab
- [x] Add a file tree link to return to the default view
- [x] Fix site admin tab column sizing for git history


nellh:
Here's some additional screenshots to help with refining the layout here:

Publish:
image

Share:
image

Versioning:
image

Site admin:
image

Download:
image

Metadata:
image

Deprecate version:
image

rwblair:
Publish/Versioning:
- showing the "create version" button disabled before any notes are added, alongside the existing note about needing a note would make the flow of publishing more explicit.
- When a version is published we are dropped into the version view, two of the four the tabs take us back to draft view. A more explicit visual indicator of what state/version you are viewing of a dataset could be nice. Draft mode can have the, "This dataset has not been published!" banner which catches the eye. This information is reflected in the url or the grey versions box in the side column both which I find subtle compared to the excitement of the dataset tools part of the page.
- Show/link to all versions and notes in the versioning tab in snapshot mode.

Share:
- "Create Anonymous Reviewer" -> "Create Anonymous Review Link" or "Create Link for Anonymous Reviewer", copy below refers to the thing being created as a link, title should reflect that.

Deprecate tab:
- Additional text about what it means to be deprecated.
- Deprecate and delete buttons could be different color to indicate their action.

It is so nice to have all of these forms in tabs, comparing it to the current site its so much smoother this way.

effigies:
> Publish/Versioning:
> 
> * showing the "create version" button disabled before any notes are added, alongside the existing note about needing a note would make the flow of publishing more explicit.

It seems useful to show the existing changelog, so people know what they've written the past and can use a consistent style/voice. We could even show a preview of what the changelog will look like once they snapshot.

nellh:
> Show/link to all versions and notes in the versioning tab in snapshot mode.

For linking to them, I think the sidebar is enough for this since it's visible while using this tab now.

> It seems useful to show the existing changelog, so people know what they've written the past and can use a consistent style/voice. We could even show a preview of what the changelog will look like once they snapshot.

Implemented showing the CHANGES file if there is an existing snapshot. Otherwise the section disappears.
image

> Additional text about what it means to be deprecated.

Suggestions for what to add here? I think it's pretty accurate as is.

> When a version is published we are dropped into the version view, two of the four the tabs take us back to draft view. A more explicit visual indicator of what state/version you are viewing of a dataset could be nice. Draft mode can have the, "This dataset has not been published!" banner which catches the eye. This information is reflected in the url or the grey versions box in the side column both which I find subtle compared to the excitement of the dataset tools part of the page.

I agree that this is still hard to immediately notice. I made one change which is to preserve if you're looking at the snapshot or draft page on the metadata tab since especially as tabs this no longer makes sense the way it was working. I think this is a good issue to break out from this PR.

Thanks for the feedback, I've made the other changes suggested.

rwblair:
> Suggestions for what to add here? I think it's pretty accurate as is.

You're right. I wasn't sure if there were any further implications for availability of the data, but looking into it more its only used to display the deprecation note.

rwblair (approved):

PR #2589, opened 2022-04-05
feat(server): Update search index immediately after snapshots are created

nellh:
This refactor splits the current indexer into two packages, the search package is the elastic client tools needed to interact with the search index and some other shared search definitions. The indexer package is now just the periodic runner and some bulk indexing functions. This should still run to pick up metadata changes external to snapshots regularly. Most of this PR is just rearranging these parts into smaller packages that the indexer and server packages can share.

Snapshot creation will now immediately index the dataset and show in search results.


effigies (approved):
LGTM

PR #2603, opened 2022-04-28
fix(app): Prevent ReadMore component from hiding bulk delete add/remove toggle

### LINKED ISSUE #2602: Bulk delete icons missing
**Describe the bug**

Bulk delete icons are missing. I can find the span in the DOM, but its computed size is 0x0 px.

**To Reproduce**

Example: https://openneuro.org/datasets/ds002893

**Screenshots**
on-missing_bulk_delete

**Desktop (please complete the following information):**
- OS: Ubungu
- Browser: Brave
- Version: Recent

nellh:
Fix #2602

### DIFF
--- a/packages/openneuro-components/src/read-more/ReadMore.tsx
+++ b/packages/openneuro-components/src/read-more/ReadMore.tsx
@@ -33,7 +33,7 @@ export const ReadMore = ({
 
           <section>
             {children}
-            <label htmlFor={id}>
+            <label htmlFor={id} className="expand-collapse">
               <span>{expandLabel}</span>
               <span>{collapseLabel}</span>
             </label>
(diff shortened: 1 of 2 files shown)


packages/openneuro-components/src/read-more/ReadMore.tsx
```diff
@@ -33,7 +33,7 @@ export const ReadMore = ({
 
           <section>
             {children}
-            <label htmlFor={id}>
+            <label htmlFor={id} className="expand-collapse">
```
effigies, line 36: I guess `className` instead of `class` is because this is technically Javascript even though it looks a lot like HTML?

nellh: Yes, it's a convention used by React because of the potential keyword conflict. Alternatives like Preact just use `class=""`.

effigies (approved):
Either way, looks like it should do the right thing, given that `<article className="has-read-more readmore-file-tree">` appears here and `<article class="has-read-more readmore-file-tree">` shows up in the rendered page.

PR #2616, opened 2022-06-02
Add API and frontend support for displaying related dataset derivatives

nellh:
This is only currently populated by available derivatives in the OpenNeuroDerivatives organization but should generalize for most other DataLad derivatives. A tab is shown on the draft and snapshot pages, right now containing identical content.


rwblair (approved):
Looks good to me. I'm really digging the nested routes for adding tabs. This PR is a good example to point to for everything involved in adding additional information to a dataset outside of datalad.

None of the changes stand out to me that would make updating how we store and retrieve derivatives down line any more difficult.

PR #2619, opened 2022-06-21
Annex configuration fixes and switch to SHA256E for new datasets

nellh:
This updates the default annex configuration for new datasets to place a smaller limit on JSON and TSV files. dataset_description.json and LICENSE are now also always committed to the repo.

With several variants of .gitattributes being valid and with slightly different requirements, I simplified the push test to allow any .gitattributes configuration. Instead we just make sure the file exists and display a recommended example if it does not. The ultimate limit for pushing large objects still exists in the receive hook, so I don't think this really opens up any potential for abuse but it does make it much more flexible for datasets with special requirements or evolving the configuration like this in the future.

### DIFF
--- a/services/datalad/datalad_service/tasks/dataset.py
+++ b/services/datalad/datalad_service/tasks/dataset.py
@@ -12,15 +12,17 @@ from datalad_service.common.annex import init_annex
 from datalad_service.common.git import git_commit, COMMITTER_EMAIL, COMMITTER_NAME
 
 # A list of patterns to avoid annexing in BIDS datasets
-GIT_ATTRIBUTES = """* annex.backend=MD5E
+GIT_ATTRIBUTES = """* annex.backend=SHA256E
 **/.git* annex.largefiles=nothing
 *.bval annex.largefiles=nothing
 *.bvec annex.largefiles=nothing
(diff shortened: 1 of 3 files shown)


services/datalad/datalad_service/tasks/dataset.py
```diff
@@ -12,15 +12,17 @@
...
 *.bval annex.largefiles=nothing
 *.bvec annex.largefiles=nothing
-*.json annex.largefiles=nothing
-*.tsv annex.largefiles=nothing
+*.json annex.largefiles=largerthan=1mb
+*.tsv annex.largefiles=largerthan=1mb
+dataset_description.json annex.largefiles=nothing
 .bidsignore annex.largefiles=nothing
 CHANGES annex.largefiles=nothing
 README annex.largefiles=nothing
```
rwblair: README will be allowed to have extensions in the future, not sure if we want to enumerate them here or just use wildcard.

nellh: A wildcard sounds fine, I doubt there's any files named README* we want to annex.

rwblair (approved):

PR #2626, opened 2022-07-05
Create new datasets with a DataLad ID

### LINKED ISSUE #2625: Recent datasets in OpenNeuroDatasets aren't true datalad datasets
It looks like all OpenNeuro datasets added to the [datasets repo ](https://github.com/OpenNeuroDatasets) on/after 21 October 2021 are not complete datalad datasets, where those added before then are. The non-datalad datasets have a `git-annex` branch, but no `.datalad` directory in the master/main branch, and specifically no `.datalad/config` file that specifies the datalad dataset `id`. See e.g. https://github.com/OpenNeuroDatasets/ds003778.

Operating on these datasets with datalad (cloning, getting data) is still possible, but other operations aren't , such as metadata handling with `datalad-metalad` for which the dataset `id` is required. It is possible to turn these datasets into true datalad datasets on the user side (i.e. after cloning from OpenNeuroDatasets), but then there wouldn't be a consistent and globally unique `id` for the dataset unless that force create is committed and merged back into the master/main branch of the affected datasets. 

Is it a specific decision to not turn new OpenNeuro datasets into true datalad datasets? And if so, is it possible to change that decision for future datasets? And/or to incorporate "fixes" for existing affected datasets?

nellh:
This fixes #2625 for new datasets.

### DIFF
--- a/services/datalad/datalad_service/tasks/dataset.py
+++ b/services/datalad/datalad_service/tasks/dataset.py
@@ -33,7 +33,7 @@ DATALAD_CONFIG = """[datalad "dataset"]
 
 def create_datalad_config(dataset_path):
     config = DATALAD_CONFIG.format(str(uuid.uuid4()))
-    os.makedirs(os.path.join(dataset_path, '.datalad'))
+    os.makedirs(os.path.join(dataset_path, '.datalad'), exist_ok=True)
     with open(os.path.join(dataset_path, '.datalad/config'), 'w') as configfile:
         configfile.write(config)
 
--- a/services/datalad/datalad_service/tasks/dataset.py
+++ b/services/datalad/datalad_service/tasks/dataset.py
@@ -5,6 +5,7 @@ Any operations that affect an entire dataset (such as creating snapshots)
 """
 import os
 import stat
+import uuid
 
 import pygit2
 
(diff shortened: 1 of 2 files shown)


services/datalad/datalad_service/tasks/dataset.py
```diff
@@ -25,6 +26,17 @@
...
+DATALAD_CONFIG = """[datalad "dataset"]
+	id = {}
+"""
+
+
+def create_datalad_config(dataset_path):
+    config = DATALAD_CONFIG.format(str(uuid.uuid4()))
+    os.makedirs(os.path.join(dataset_path, '.datalad'))
+    with open(os.path.join(dataset_path, '.datalad/config'), 'w') as gitattributes:
+        gitattributes.write(config)
```
effigies:
```suggestion
    with open(os.path.join(dataset_path, '.datalad/config'), 'w') as configfile:
        configfile.write(config)
```

services/datalad/tests/test_datalad.py
```diff
@@ -13,8 +14,16 @@ def test_create_dataset(datalad_store):
     ds_id = 'ds000002'
     author = pygit2.Signature('test author', 'test@example.com')
     create_dataset(datalad_store, ds_id, author)
-    assert Dataset(os.path.join(
-        datalad_store.annex_path, ds_id)).repo is not None
+    ds = Dataset(os.path.join(
+        datalad_store.annex_path, ds_id))
```
effigies:
```suggestion
    ds = Dataset(os.path.join(datalad_store.annex_path, ds_id))
```

effigies (approved):
Nitpicks. This looks right to me.

PR #2627, opened 2022-07-05
fix: Add a datalad id while creating a snapshot if one does not exist

nellh:
This is the second fix for #2625. When a snapshot is created, check if the .datalad/config file exists and if it does not, add one with an id.

I had tried validating the .datalad/config contents but that leads to needing to merge them safely. I think we should stick with just assuming it exists and is as expected, or it doesn't exist and we should add our minimum version that provides dataset identity, other cases can still be handled by the uploader with git access or via the API (OpenNeuro currently hides any .datalad files from the UI).

### DIFF
--- a/services/datalad/datalad_service/tasks/snapshots.py
+++ b/services/datalad/datalad_service/tasks/snapshots.py
@@ -5,6 +5,7 @@ import re
 import pygit2
 
 from datalad_service.common.git import git_show, git_tag
+from datalad_service.tasks.dataset import create_datalad_config
 from datalad_service.tasks.description import update_description
 from datalad_service.tasks.files import commit_files
 
@@ -132,6 +133,16 @@ def validate_snapshot_name(store, dataset, snapshot):
             'Tag "{}" already exists, name conflict'.format(snapshot))
 
 
+def validate_datalad_config(store, dataset):
+    """Add a .datalad/config file if one does not exist."""
+    dataset_path = store.get_dataset_path(dataset)
+    try:
+        git_show(dataset_path, 'HEAD', '.datalad/config')
+    except:
+        create_datalad_config(dataset_path)
+        commit_files(store, dataset, ['.datalad/config'])
+
+
 def save_snapshot(store, dataset, snapshot):
     repo = pygit2.Repository(store.get_dataset_path(dataset))
     repo.references.create(f'refs/tags/{snapshot}', repo.head.target.hex)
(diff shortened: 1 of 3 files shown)


services/datalad/datalad_service/tasks/snapshots.py
```diff
@@ -132,6 +133,16 @@ def validate_snapshot_name(store, dataset, snapshot):
             'Tag "{}" already exists, name conflict'.format(snapshot))
 
 
+def validate_datalad_config(store, dataset):
+    """Add a .datalad/config file if one does not exist."""
+    dataset_path = store.get_dataset_path(dataset)
+    try:
+        git_show(dataset_path, 'HEAD', '.datalad/config')
+    except:
```
effigies: Do we know what kind of exception to expect here? Bare `except`s are generally considered bad practice.

services/datalad/tests/test_snapshots.py
```diff
@@ -39,6 +40,29 @@ def test_create_snapshot(client, new_dataset):
...
+def test_create_snapshot_no_config(datalad_store, client, new_dataset):
+    """Validate adding a datalad config if one is missing during snapshot creation"""
+    ds_id = os.path.basename(new_dataset.path)
+    snapshot_id = '1'
+    # Delete the default config first
+    response = client.simulate_delete('/datasets/{}/files'.format(
+        ds_id), body='{ "filenames": [".datalad/config"] }')
+    assert response.status == falcon.HTTP_OK
+    assert json.loads(response.content)['deleted'] == [
+        '.datalad/config']
```
effigies, line 52: Since our later check is that Datalad could find an ID, I think it would make sense to verify that it can't pre-snapshot.

```suggestion
        '.datalad/config']
    ds = Dataset(os.path.join(datalad_store.annex_path, ds_id))
    assert ds.id is None
```

services/datalad/datalad_service/tasks/snapshots.py
nellh: Testing it, I can narrow it to pygit2.GitError. The other exceptions here should raise instead of just overwriting the file (invalid decoding or missing git objects).

nellh: Actually KeyError is better, GitError just happens when there's some other issue (read error or a locked repo) rather than the file not existing in the commit referenced, and we probably don't want to try to replace it in that case either.

services/datalad/tests/test_snapshots.py
nellh: Added a variant of this but have to close the dataset or datalad caches the id despite removing the config.

services/datalad/tests/dataset_fixtures.py
```diff
@@ -98,6 +98,7 @@ def new_dataset(datalad_store):
     with open(changes_path, 'w') as f:
         f.write(CHANGES)
     ds.save(changes_path)
+    ds.close()
     return ds
```
effigies, line 102: Is the dataset still usable once closed?

nellh: Yes, it just forces DataLad to reopen the repo from the filesystem on the next access.

effigies (approved):

PR #2638, opened 2022-07-22
Fix missing JSON-LD header on snapshot pages

nellh:
Majority of the changes here are just to add some test coverage for this. A new test renders the full snapshot container with mock data and keeps a snapshot for the page and head content.


effigies (approved):
Looks sensible.

PR #2642, opened 2022-07-29
ENH: Support viewing MGH/MGZ files in browser

effigies:
This will allow us to more easily look in FreeSurfer derivatives. `.mgh`/`.mgz` is the FreeSurfer volumetric format, roughly equivalent to NIfTI. I've verified that NiiVue supports it through https://niivue.github.io/niivue/features/basic.multiplanar.html.

Please let me know if there's a test that I can provide.

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/files/file-viewer-type.jsx
+++ b/packages/openneuro-app/src/scripts/dataset/files/file-viewer-type.jsx
@@ -19,7 +19,7 @@ const FileViewerType = ({ path, url, data }) => {
     path.endsWith('.txt')
   ) {
     return <FileViewerText data={data} />
-  } else if (path.endsWith('.nii.gz') || path.endsWith('.nii') || path.endswith('.mgh') || path.endswith('.mgz') {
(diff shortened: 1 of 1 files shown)


packages/openneuro-app/src/scripts/dataset/files/file-viewer-type.jsx
```diff
@@ -19,7 +19,7 @@ const FileViewerType = ({ path, url, data }) => {
     path.endsWith('.txt')
   ) {
     return <FileViewerText data={data} />
-  } else if (path.endsWith('.nii.gz') || path.endsWith('.nii')) {
+  } else if (path.endsWith('.nii.gz') || path.endsWith('.nii') || path.endswith('.mgh') || path.endswith('.mgz') {
```
nellh:
```suggestion
  } else if (path.endsWith('.nii.gz') || path.endsWith('.nii') || path.endsWith('.mgh') || path.endsWith('.mgz')) {
```
Missing close paren and camel casing for endsWith.

PR #2667, opened 2022-09-16
deps: Update bids-validator to 1.9.8

nellh:
Update to the [latest release](https://github.com/bids-standard/bids-validator/releases/tag/v1.9.8).


nellh:
Raising the available memory for CI fixed this, we should see if we can get back down below 4GB for builds though.

PR #2670, opened 2022-09-21
fix(app): Prevent logged out users from attempting to generate API keys

### LINKED ISSUE #1203: Anonymous users can attempt to generate an API key
You aren't asked to login if you visit the /keygen as an anonymous user, the keygen doesn't actually work but it is misleading for users who may not have logged in yet.

nellh:
Fixes #1203


effigies (approved):
Well that's straightforward.

PR #2678, opened 2022-09-29
File cache tree support

nellh:
This replaces the commit files cache that many operations depend on with a per git tree object cache that can be loaded in incrementally.

- [x] Implement a size resolver that doesn't depend on reading all trees - currently returns dummy values in some cases (128 bytes).
- [x] Add tests for file tree rendering step.
- [x] Restore previous sorting behavior to put files first and directories last.
- [x] Update CLI downloads to use the new tree API
- [x] Update browser downloads to use the new tree API
- [x] Show any failed files when downloading in the browser


nellh:
One limitation here is this is returning the size reported by the validator again. I think we should address this on the validator side, it doesn't make sense for OpenNeuro to crawl the whole dataset just to report an accurate size when the validator has to do that anyways. I did try using git-annex info to get an estimate but this is still too slow for our largest datasets given storage performance available.

packages/openneuro-app/src/scripts/dataset/download/download-native.js
```diff
@@ -40,30 +40,90 @@ class DownloadAbortError extends Error {
...
+      const fileHandle = await openFileTree(
+        dirHandle,
+        path ? `${path}/${file.filename}` : file.filename,
+      )
+      // Skip files which are already complete
+      if (fileHandle.size == file.size) continue
+      const writable = await fileHandle.createWritable()
+      const { body, status, statusText } = await fetch(file.urls.pop())
+      let loaded = 0
+      const progress = new TransformStream({
```
rwblair, line 88: Very cool use of transform stream.

packages/openneuro-app/src/scripts/dataset/download/download-native.js
```diff
@@ -40,30 +40,90 @@ class DownloadAbortError extends Error {
...
+          })
+          loaded += chunk.length
+          controller.enqueue(chunk)
+        },
+      })
+      if (status === 200) {
+        await body.pipeThrough(progress).pipeTo(writable)
+      } else {
+        apmTransaction.captureError(statusText)
+        return requestFailureToast()
```
rwblair: Is the following accurate?: In earlier versions if a file failed the entire download would return then and there, where as this version should only result in the current directory download being abandoned and all other paths will still be downloaded. If so I could imagine a max failed download number to have `downloadNative` bail out on could be a thing.

packages/openneuro-app/src/scripts/dataset/download/download-native.js
```diff
@@ -40,30 +40,90 @@ class DownloadAbortError extends Error {
...
+      // Regular file
+      if (downloadCanceled) {
+        throw new DownloadAbortError('Download canceled by user request')
+      }
+      const fileHandle = await openFileTree(
+        dirHandle,
+        path ? `${path}/${file.filename}` : file.filename,
+      )
+      // Skip files which are already complete
+      if (fileHandle.size == file.size) continue
```
rwblair, line 84: Do fileHandle's have a size attribute? Wasn't seeing one on mdn and poking around locally but I may be incorrect in what type the one here is.

services/datalad/datalad_service/handlers/files.py
```diff
@@ -16,52 +15,42 @@ def __init__(self, store):
         self.store = store
         self.logger = logging.getLogger('datalad_service.' + __name__)
 
-    def on_get(self, req, resp, dataset, filename=None, snapshot='HEAD'):
+    def on_get(self, req, resp, dataset, filename, snapshot='HEAD'):
```
nellh, line 18: This was overloaded so that the files request with an argument would return a file but without the file argument it would return the listing. The change here is to make filename required and move all the listing functionality to the TreeResource class.

packages/openneuro-app/src/scripts/dataset/download/download-native.js
nellh: Yes, they do.

packages/openneuro-app/src/scripts/dataset/download/download-native.js
nellh: That's right. You can cancel the entire download by closing the toast but otherwise I think we want to try to get as much as possible, so that if something is missed you can resume with a second attempt and only need a small subset to get a complete dataset.

That brings up a fix here though, I think this should show you if any files failed due to network or other interruptions at the end and suggest a retry. Otherwise it's easy to assume you have the whole dataset when one file is missing.

packages/openneuro-app/src/scripts/dataset/download/download-native.js
rwblair: Turns out I was looking at a FileSystemFileHandle and not a FileSystemFileEntry that the getFile returns.

rwblair:
Does this call to get_repo_files need to be amended, wasn't sure if `snapshot` there was treeish:
https://github.com/OpenNeuroOrg/openneuro/blob/69816248878551c40892d3c8bc56306410c2bd0b/services/datalad/datalad_service/handlers/annex_objects.py#L21

 \* _output of above used by pygit2 repo.index.remove_all(...) which is a recursive delete._

nellh:
> Does this call to get_repo_files need to be amended, wasn't sure if `snapshot` there was treeish:
> 
> https://github.com/OpenNeuroOrg/openneuro/blob/69816248878551c40892d3c8bc56306410c2bd0b/services/datalad/datalad_service/handlers/annex_objects.py#L21

The snapshot value here is the tag (or the commit hash, either way) and will return the commit tree.

nellh:
> 

Got it, fixed in b9e1243

services/datalad/datalad_service/common/annex.py
```diff
@@ -50,9 +50,9 @@ def read_ls_tree_line(gitTreeLine, files, symlinkFilenames, symlinkObjects):
...
         gitTreeLine)
     # Skip git / datalad files
-    if filename.startswith('.git/'):
+    if filename.startswith('.git'):
         return
-    if filename.startswith('.datalad/'):
+    if filename.startswith('.datalad'):
         return
     if filename == '.gitattributes':
         return
```
effigies, line 58: This will now be covered by L53. Are we happy to also be skipping `.gitmodules`, `.gitignore`, etc?

nellh: This just affects if OpenNeuro displays these, so do we want to display .gitignore? We don't allow upload of it but it could exist via git push.

rwblair (approved):

services/datalad/datalad_service/common/annex.py
effigies: I think it's fine to exclude from display. We should probably reject git pushes with gitignores until we drop the working tree.

PR #2697, opened 2022-10-19
Prevent clearing front page search parameter during navigation

### LINKED ISSUE #2696: search term doesn't get carried over from main page search field
**Describe the bug**
When a search term is entered into the search field on the main landing page, it doesn't actually get carried into the search page.

**To Reproduce**
1. Go to ['the main page'](https://openneuro.org/)
2. Enter a search term into the search field and press enter or click the search icon.
3. The resulting page will not have any keywords present, and will say "Showing all available datasets"

**Expected behavior**
The search term should carry over

**Desktop (please complete the following information):**
 - Browser: Chrome
 - Version 106.0.5249.119

nellh:
The search parameter was set here and then would be immediately cleared by the navigate call. This skips setting it on the context and includes it in the URL. Adds a test which verifies this behavior to prevent future regressions.

Fixes #2696

### DIFF
--- a/packages/openneuro-app/src/scripts/__mocks__/config.ts
+++ b/packages/openneuro-app/src/scripts/__mocks__/config.ts
@@ -23,4 +23,7 @@ export const config = {
   sentry: {
     environment: 'unit-tests',
   },
+  support: {
+    url: 'https://example.com/test-suite',
+  },
 }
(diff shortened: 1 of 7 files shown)


effigies (approved):
LGTM

PR #2700, opened 2022-10-19
fix(app): Don't depend on parent routes in DeprecateSnapshotPage

### LINKED ISSUE #2699: Cannot deprecate snapshot
**Describe the bug**
Attempting to deprecate a snapshot fails with no error.

**To Reproduce**
1. Go to https://openneuro.org/datasets/ds004218/versions/1.0.0/deprecate
2. Enter a deprecation reason (I used "This version contains incomplete files")
3. Click button
4. Nothing seems to happen

**Expected behavior**
Snapshot should be deprecated

nellh:
Fixes regression causing the snapshot tag value to be missing.

Fixes  #2699


effigies (approved):
LGTM!

PR #2715, opened 2022-11-16
Consolidate to native fetch implementations and switch to vitest

nellh:
Dropping fetch polyfills and always uses browser or Node.js native fetch. This makes the minimum NodeJS version 18 for all tools but this version is now widely available as LTS packages.

TODO: 
- [x] Fix HeaderContainer test


effigies (approved):
This all makes sense. (Only reviewed the second commit.)

nellh:
@rwblair Could you take a quick look at the test changes? This is ready now I believe.

PR #2738, opened 2022-12-15
fix(worker): Update local and remote defaults to main branch

nellh:
This updates some mismatched master branch usage to always use main.

* On new commits, this will rename the master branch to main if required before making the commit. Updates symbolic HEAD reference so that new operations implicitly use the main branch.
* When pushing to GitHub, both main and master branches are pushed. Sets the default branch and repo description/name on pushes.
* New test that creates a default master branch Datalad repo and then updates it to use main on git_commit.

### DIFF
--- a/services/datalad/datalad_service/common/git.py
+++ b/services/datalad/datalad_service/common/git.py
@@ -40,8 +40,22 @@ def git_tag_tree(repo, tag):
     return repo.get(tag_reference.target).tree_id
 
 
+def git_rename_master_to_main(repo):
+    # Make sure the main branch is used, update if needed
+    # Always update the symbolic reference for HEAD
+    repo.references['HEAD'].set_target('refs/heads/main')
+    master_branch = repo.branches.get('master')
+    if master_branch:
+        main_branch = master_branch.rename('main', True)
+        # Abort the commit if this didn't work
+        if not main_branch.is_head():
+            raise Exception('Unable to rename master branch to main')
+
+
 def git_commit(repo, file_paths, author=None, message="[OpenNeuro] Recorded changes", parents=None):
     """Commit array of paths at HEAD."""
+    # master -> main if required
+    git_rename_master_to_main(repo)
     # Refresh index with git-annex specific handling
     annex_command = ["git-annex", "add"] + file_paths
     subprocess.run(annex_command, check=True, cwd=repo.workdir)
@@ -61,6 +75,6 @@ def git_commit_index(repo, author=None, message="[OpenNeuro] Recorded changes", 
         parent_commits = parents
     tree = repo.index.write_tree()
     commit = repo.create_commit(
-        'refs/heads/master', author, committer, message, tree, parent_commits)
+        'refs/heads/main', author, committer, message, tree, parent_commits)
     repo.head.set_target(commit)
     return commit
--- a/services/datalad/datalad_service/common/git.py
+++ b/services/datalad/datalad_service/common/git.py
@@ -42,14 +42,14 @@ def git_tag_tree(repo, tag):
 
 def git_rename_master_to_main(repo):
     # Make sure the main branch is used, update if needed
-    # Always update the symbolic reference for HEAD
-    repo.references['HEAD'].set_target('refs/heads/main')
     master_branch = repo.branches.get('master')
     if master_branch:
         main_branch = master_branch.rename('main', True)
         # Abort the commit if this didn't work
         if not main_branch.is_head():
             raise Exception('Unable to rename master branch to main')
+    # Always update the symbolic reference for HEAD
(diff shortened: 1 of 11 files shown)


services/datalad/datalad_service/common/github.py
```diff
@@ -23,17 +23,30 @@ def create_github_repo(dataset_path, dataset_id):
...
+    # Make sure the default branch is correct
+    github_set_default_branch(dataset_id)
+
+
+def github_set_default_branch(dataset_id):
+    """Sets the repo default branch to 'main' and resets the description/homepage fields."""
+    ses = Github(DATALAD_GITHUB_TOKEN)
+    org = ses.get_organization(DATALAD_GITHUB_ORG)
+    repo = org.get_repo(dataset_id)
+    repo.edit(default_branch="main", description="OpenNeuro dataset",
```
effigies: Is it reasonably un-clunky to put the dataset name here? That would make the listing in https://github.com/OpenNeuroDatasets much more informative.

```suggestion
    repo.edit(default_branch="main", description=f"OpenNeuro dataset {dataset_id}: {description['Name']}",
```

nellh: As part of the changes for metadata tagging on GitHub I was planning on loading some information from OpenNeuro's API service and updating it here.

That said, we could just read the dataset_description.json inline here and use the name from that if it exists, not sure how soon we'll get back to those GitHub metadata changes with other priorities.

services/datalad/datalad_service/common/github.py
```diff
@@ -23,17 +24,38 @@ def create_github_repo(dataset_path, dataset_id):
...
+    subprocess.check_call(
+        ['git', 'push', 'github', f'{tag}:refs/heads/master'], cwd=dataset_path)
     subprocess.check_call(
         ['git', 'push', 'github', 'git-annex:refs/heads/git-annex'], cwd=dataset_path)
     # Update tags
     subprocess.check_call(
         ['git', 'push', '--tags', 'github'], cwd=dataset_path)
+    # Make sure the default branch is correct
+    description = read_dataset_description(dataset_path, tag)
+    github_set_repo_config(dataset_id, description.Name)
```
effigies: I think this is just a dict, not an object.

```suggestion
    github_set_repo_config(dataset_id, description["Name"])
```

nellh: Yeah, I replaced this commit once I'd tested it. I think I need to also add an exception case for when the dataset_description.json git object doesn't exist, even though that should really never happen.

services/datalad/datalad_service/common/git.py
```diff
@@ -40,8 +40,22 @@ def git_tag_tree(repo, tag):
     return repo.get(tag_reference.target).tree_id
 
 
+def git_rename_master_to_main(repo):
+    # Make sure the main branch is used, update if needed
+    master_branch = repo.branches.get('master')
+    if master_branch:
+        main_branch = master_branch.rename('main', True)
```
effigies, line 52: Is there any chance there's a dataset with both `master` and `main`? It looks like `Branch.rename(force=True)` will clobber, but if both are present, are we confident that `master` is authoritative and not `main`? I could imagine a situation where someone with git access pushes a stale `master` branch to a dataset that's had `master` renamed to `main`.

nellh: Right now there are no main branch datasets and OpenNeuro cannot edit them without these fixes. Maybe instead we block master for new pushes? I could add that here.

services/datalad/datalad_service/common/git.py
```diff
@@ -40,8 +40,22 @@ def git_tag_tree(repo, tag):
...
+        # Abort the commit if this didn't work
+        if not main_branch.is_head():
+            raise Exception('Unable to rename master branch to main')
+    # Always update the symbolic reference for HEAD
+    repo.references['HEAD'].set_target('refs/heads/main')
```
effigies: What do you think of this?

```suggestion
    HEAD = repo.references["HEAD"]
    # Make sure the main branch is used, update if needed
    if HEAD.target == "refs/heads/master":
        main_branch = repo.branches.get('master').rename('main', True)
        # Abort the commit if this didn't work
        if not main_branch.is_head():
            raise Exception('Unable to rename master branch to main')
    # Always update the symbolic reference for HEAD
    HEAD.set_target('refs/heads/main')
    # Remove master if it's still around
    if "master" in repo.branches:
        repo.branches["master"].delete()
```

nellh:
Okay, adding two more test cases:

- [x] Test for unusual state where the default is some other branch and the commit should be aborted.
- [ ] Test for existing dataset with history including a tag to validate the rename preserves history in that case.

One more fix:

- [x] Verify if master is head before renaming unless it's an empty branch.

Issue #2742, opened 2022-12-23
Unable to upload dataset with CLI after passing bids-validator


koreywylie:
Unable to upload dataset using command line interface v4.14.0, despite this dataset passing bids-validator with minor warnings, command-line output attached.
[openneuro_upload_output.txt](https://github.com/OpenNeuroOrg/openneuro/files/10297330/openneuro_upload_output.txt)
[bids-validator_output.txt](https://github.com/OpenNeuroOrg/openneuro/files/10297331/bids-validator_output.txt)

nellh:
Can you provide the version of Node.js and what OS you're uploading from?

koreywylie:
OSX 10.15.7

$ npm list -g
/usr/local/lib
├── @***@***.***
├── ***@***.***
├── ***@***.***
└── ***@***.***

Thanks & best,  --Korey--
________________________________

Arshitha:
Hi I'm having the same issue.

I'm running the command from HPC system with 
```bash
Linux version 3.10.0-1160.80.1.el7.x86_64 (mockbuild@x86-vm-41.build.eng.bos.redhat.com) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) )
```
and software versions as follows:   
```bash
openneuro v4.14.0
node.js v18.12.1
```

I'm trying to update an existing dataset (ds004215) but it gives me the following stdout:
```bash
[arshithab@biowulf v1.0.2_changes]$ openneuro upload -d ds004215 -i -n --ignoreSubjectConsistency .
Adding files to "ds004215"
bids-validator@1.9.9
        Summary:                  Available Tasks:        Available Modalities: 
        512 Files, 11.01MB                                MEG                   
        66 - Subjects                                                           
        1 - Session                                                             


        If you have any questions, please post on https://neurostars.org/tags/bids.

Checking remote files...
(node:110534) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
=======================================================================
512 files to be uploaded with a total size of 11.0 MB
? Begin upload? Yes
=======================================================================
Starting a new upload (58924465) to dataset: 'ds004215'
d. 11 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limittenersExceededWarning: Possible EventTarget memory leak detecte
Retrying upload for dataset_description.json: TypeError: fetch failed
[...]

PR #2745, opened 2023-01-04
Fix retry file upload errors with native fetch on Node 18

nellh:
Two fixes here as reported in #2742

1. File upload retries were broken after the switch to native fetch, it's standard that a Request object cannot be reused and our implementation reused this since it works on non-standard polyfills we had been using.
2. The npm install of the CLI tool was not working because Node's implementation of fetch does not support the standard argument of a web stream as the body of the request. Instead we switch to a non-standard Node only method of building the request here passing in an AsyncIterable as the body.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": ["packages/*"],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "4.14.0"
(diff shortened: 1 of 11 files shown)


rwblair (approved):
Does the README.md for @openneuro/cli need to be updated requiring node 18?

PR #2770, opened 2023-02-13
Normalize nested directory paths from annexed objects to eliminate failure to resolve '..' in symlinked paths

### LINKED ISSUE #2769: Web UI downloads failing for annexed files removed in draft
### What went wrong?

https://openneuro.org/datasets/ds001907 has 46 participants. All but 9 are removed in draft. Attempting to download an image from a removed participant in the latest snapshot fails.

Data is still on S3 and can be retrieved through DataLad.

### Expected behavior

Unpurged files should be downloadable even if removed in draft or later snapshots.

### How to reproduce

1. Go to https://openneuro.org/datasets/ds001907/versions/3.0.2
2. Attempt to view/download NIfTI from the final subject (e.g., https://openneuro.org/datasets/ds001907/versions/3.0.2/file-display/sub-RC4227:ses-1:anat:sub-RC4227_ses-1_T1w.nii.gz)
3. Get `{"error": "file not found"}`

### Desktop

- OS: Ubuntu
- Browser: Firefox
- Version: 109

### Phone

- Device:
- OS:
- Browser:
- Version:

### Additional information

_No response_

nellh:
Includes a new test to cover this situation. This affects any nested annexed paths accessed with a commit identifier but I added a history step in the test since that wasn't covered and could have been related.

Fixes #2769

### DIFF
--- a/services/datalad/datalad_service/handlers/files.py
+++ b/services/datalad/datalad_service/handlers/files.py
@@ -22,8 +22,8 @@ class FilesResource(object):
             # If the file begins with an annex path, return that path
             if file_content[0:4096].find('.git/annex') != -1:
                 # Resolve absolute path for annex target
-                target_path = os.path.join(
-                    ds_path, os.path.dirname(filename), file_content)
+                target_path = os.path.normpath(os.path.join(
+                    ds_path, os.path.dirname(filename), file_content))
                 # Verify the annex path is within the dataset dir
                 if ds_path == os.path.commonpath((ds_path, target_path)):
                     fd = open(target_path, 'rb')
--- a/services/datalad/tests/test_files.py
+++ b/services/datalad/tests/test_files.py
@@ -43,6 +43,31 @@ def test_get_annexed_file(client):
     assert result.content.decode() == file_data
 
 
+def test_get_nested_annexed_file(client):
+    # Test accessing a subdir annexed file in git history
+    ds_id = 'ds000001'
+    file_data = 'test image, please ignore'
+    # Add it
+    response = client.simulate_post(
+        '/datasets/{}/files/sub-01:anat:data.nii.gz'.format(ds_id), body=file_data)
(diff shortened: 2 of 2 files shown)


services/datalad/datalad_service/handlers/files.py
```diff
@@ -22,8 +22,8 @@ def on_get(self, req, resp, dataset, filename, snapshot='HEAD'):
             # If the file begins with an annex path, return that path
             if file_content[0:4096].find('.git/annex') != -1:
                 # Resolve absolute path for annex target
-                target_path = os.path.join(
-                    ds_path, os.path.dirname(filename), file_content)
+                target_path = os.path.normpath(os.path.join(
+                    ds_path, os.path.dirname(filename), file_content))
                 # Verify the annex path is within the dataset dir
                 if ds_path == os.path.commonpath((ds_path, target_path)):
```
effigies, line 28: The test fails without this? I'm having trouble understanding how this changes the semantics at all. Because we get "file not found" and not "file not found in git tree", we're getting an `IOError` instead of failing this path check. We could plausibly be getting that on `git_show()` (L21), `open` (L29) or `fstat` (L31).

Is there a case where `open(path)` and `open(normpath(path))` should have any different result?

nellh: There is a difference, the path referenced has been removed. That's why this works for paths that still exist in the draft. You can't traverse up the directory tree from a directory that can't be opened. Normpath gives us the path without the potentially missing components by converting git-annex's dataset-root/../../.git paths into dataset-root/.git paths that do exist.

nellh: Oh, so the history part actually is a requirement then. Good thing my test makes more sense than my PR description.

effigies: Ah, thank you. That's what I was missing. This LGTM, then.

PR #2771, opened 2023-02-13
feat(app): Limit dataset full deletion to site admin users

nellh:
This is a basic version of #2734 that limits full dataset deletion to site admins only.


effigies (approved):
Makes sense to me, comparing with https://github.com/OpenNeuroOrg/openneuro/blob/621a63cbaf1a27845bc44b0faa227fafe8a92418/packages/openneuro-app/src/scripts/authentication/admin-user.jsx#L13-L21

PR #2797, opened 2023-04-11
Implement snapshot downloadFiles API call

nellh:
This creates a new flat API call that is indefinitely cached once a snapshot is fully exported. This is needed to allow for faster downloads without each client repeating tree creation.

Changes:
* Adds a doNotCache options to the CacheItem miss function to allow for a cache miss returning valid data to still avoid the cache
* Uses the doNotCache flag to not cache any file trees that are not from S3. This is probably the biggest change here because draft performance may suffer, it might make sense to add a flag for drafts to not give up caching them entirely but prevent their inclusion in the snapshotDownload cache before an export is complete.
* Adds Snapshot.downloadFiles which follows the existing DatasetFile interface.
* Limits the cacheClear mutation to the snapshotDownload item type only.
* Deprecates the unused DatasetChanges model.


nellh:
After some discussion, I think we should instead make the file tree cache stable and put the export URLs in their own cache that is invalidated when the export status changes.

nellh:
> After some discussion, I think we should instead make the file tree cache stable and put the export URLs in their own cache that is invalidated when the export status changes.

Going back on this idea after some additional development on it, it leads to a difficult to resolve pattern where the relationship between the git tree objects and the URLs can get stale that isn't easy to resolve without dropping the draft cache anyways. The existing solution works pretty well, if it does cause real performance issues with drafts we could set an expiration time on the unexported draft tree cache but I think this version should be measured first.

PR #2805, opened 2023-04-18
Use external URLs for FileTree and getFile APIs

nellh:
This change prefers the S3 URL for a file if an exported copy of the data is available on S3.

* The FileTree component already had this information available to it but it wasn't passed into the display components for use.
* The getFile API itself now resolves the tree needed for a path (usually from cache) and will serve a 302 redirect to S3 if one is found.


effigies (approved):
This makes sense to me.

PR #2822, opened 2023-06-05
fix(app): Disable reviewer links creation without a dataset version

nellh:
This disables the redirect to snapshot for reviewer users, allowing them to browse a draft. Also catches some rare cases where the redirect can trigger when snapshots are being created and avoids forcing a redirect before one exists by checking the tag value before attempting the redirect.

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/draft-container.tsx
+++ b/packages/openneuro-app/src/scripts/dataset/draft-container.tsx
@@ -76,7 +76,6 @@ const DraftContainer: React.FC<DraftContainerProps> = ({ dataset }) => {
   const [cookies] = useCookies()
   const profile = getUnexpiredProfile(cookies)
   const isAdmin = profile?.admin
-  const isReviewer = profile?.scopes?.includes('dataset:reviewer')
   const hasEdit =
     hasEditPermissions(dataset.permissions, profile?.sub) || isAdmin
   const hasDraftChanges =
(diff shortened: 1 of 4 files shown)


effigies:
I guess the question is whether we want to allow reviewers draft access. As an external reviewer, it would not be my goal to make the 1.0.0 release perfect, but to have something I can point to and say I reviewed it when it looked like this.

I think I'd rather disable reviewer links (or just give them sensible error messages) until the first snapshot.

nellh:
> I guess the question is whether we want to allow reviewers draft access. As an external reviewer, it would not be my goal to make the 1.0.0 release perfect, but to have something I can point to and say I reviewed it when it looked like this.
> 
> I think I'd rather disable reviewer links (or just give them sensible error messages) until the first snapshot.

Went ahead and implemented that instead. This still includes the fix for the case which would cause the redirect to lead to a bad URL if the dataset has no snapshots but reviewers will still get redirected from draft to snapshot. Instead the reviewer link button is hidden with a warning to create a version first.

image

effigies:
What happens to the reviewer links that already exist for a dataset without snapshots? Just looking at this, it seems like they'll be allowed to see the draft because the redirect doesn't get hit, but I might be misreading.

nellh:
> What happens to the reviewer links that already exist for a dataset without snapshots? Just looking at this, it seems like they'll be allowed to see the draft because the redirect doesn't get hit, but I might be misreading.

That's correct, the draft will be available.

effigies (approved):
Okay. This works. The dataset with reviewer links for the draft has now made a snapshot, so the redirect should work correctly going forward.

PR #2852, opened 2023-07-13
fix(app): Rename mock-content to content

nellh:
Fixes an extra character and missing punctuation on the front page as well.


packages/openneuro-components/src/content/activity-slider-content.jsx
```diff
@@ -11,8 +11,7 @@ export const RecentData = {
             latestSnapshot: {
               tag: '1.0.0',
               description: {
-                Name:
-                  'Runabout: A mobile EEG study of auditory oddball processing in laboratory and real-world conditions',
+                Name: 'Runabout: A mobile EEG study of auditory oddball processing in laboratory and real-world conditions',
```
effigies: I think this file may actually be mock content, as it's hard-coding most viewed and recently published, and these are definitely not what is shown there.

nellh: Thanks, I've cleaned up the unused content. There's some only partially used content remaining but anything entirely unused is now removed.

PR #2853, opened 2023-07-14
DOC: Update docs to be more user-facing

effigies:
This PR does the following:

1) Reformats the FAQ into a doc site Markdown page: https://effigies.github.io/openneuro/faq.html
2) Removes GitHub README from index.html content, instead just linking to the repository in the table of contents
3) Splits TOC into user guide+faq, policies, advanced usage, developer resources.
4) Removes the policy/index page, which existed only for a TOC we can handle by creating multiple TOCs

Test render here: https://effigies.github.io/openneuro


agt24:
This looks great @effigies! 

I think this is an improvement on the current version of the documentation, but the navigation is a still a touch confusing as there is some repetition between menu items on the left (on this page) and menu items on the right

There's also significant overlap between content in the landing page ("OpenNeuro Documentation") and the "User Guide".
Both seems to be striving to be an entry point for someone who is looking for help taking different guesses about what info will be most helpful to them. To some extent the FAQ is also trying to be an entry point with different guesses about what to highlight. Ideally the docs would have a single entry point (be it called OpenNeuro Documentation or User Guide) for all users from which they could be re-directed based on their needs (Finding/downloading data, Uploading/sharing data, Advanced topics/developers resources (which seem mergable), and Policy stuff (maybe doesn't need to be top level?),

agt24:
Adding another related point that @effigies and I were discussing on Slack last week. These are all really good docs, but they aren't very prominently linked on the OpenNeuro site (I'd never seen them before last week).

Maybe consider replacing the "FAQ" and "Support"  links at the top of the OpenNeuro.org landing page with a single link to "Help" or whatever that then provides the user with options for reading docs, watching video tutorials, or submitting a ticket.

PR #2854, opened 2023-07-14
Fix typos detected by codespell

effigies:
Encountered while checking the doc PR (#2853) for typos. Only the `middlware -> middleware` has any effect on code.


nellh (approved):
Testing the middleware fix, this works either way but only because this isn't cached for other reasons. It should never be cached, so this looks good.

Issue #2921, opened 2023-10-04
Search by Sex doesn't catch 'M' and 'F' in participants.tsv


jbwexler:
### What went wrong?

If you search by Sex and click 'Male' or 'Female', only a small number of datasets come up. That's because it seems to only recognize datasets that include the full words 'Male' or 'Female' in participants.tsv, but most only include the letter 'M' or 'F'. 

More generally, we should do some testing of the search functions to make sure they're functioning properly.

### Expected behavior

It should be recognizing 'M' and/or 'F' from participants.tsv.

### How to reproduce

1. Click Browse by Modalities
2. Click MRI
3. Click Sex
4. Click Male (or Female)

### Desktop

- OS: Mac
- Browser: Librewolf
- Version: 112.0.2-1

### Phone

- Device:
- OS:
- Browser:
- Version:

nellh:
Fixed in #2932, thanks @ehavener!

PR #2932, opened 2023-10-31
Fixes search by sex not catching 'M' and 'F' in participants.tsv

ehavener:
This PR resolves the open issue [here](https://github.com/OpenNeuroOrg/openneuro/issues/2921). It adds a multiMatchQuery to search for returning all possible valid strings per BIDS.  I tested this locally on a few datasets and it seems to work fine.


nellh:
Thank you for your contribution! I've tested this and it looks good to me.

PR #2940, opened 2023-11-10
feat(dataset): Annex TSV files in phenotype directory

effigies:
This adds `phenotype/*.tsv` to the annex by default.

Should we (and how should we) inject this into legacy datasets?

I couldn't find a place where we test that expected files get annexed or not. If a test is needed, I'll need a little guidance of what to use as a model.


effigies:
Also not positive of the convention for tagging PRs...

nellh:
> Also not positive of the convention for tagging PRs...

Squishymedia was using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and I've mostly kept going with that.

PR #2947, opened 2023-11-27
Add clickthrough agreement for downloads

nellh:
This just needs text content to be ready for review.

The agreement is a floating banner on the bottom of the browser window. It disappears once you have agreed and enables download/view file links. Download pages redirect to the dataset page if the agreement has not been accepted. The agreement is persisted in local storage and shown again if the user is downloading from another browser or clears local storage.

image


effigies:
Can this be pushed to staging for an interactive view?

nellh:
> Can this be pushed to staging for an interactive view?

It's on staging now. I'm planning on making a couple minor fixes once I have the text as far as styling. The button gets pushed off the screen on mobile sometimes. Spacing can be a little off where the button gets cut off at some screen sizes. Let me know if you see other issues.

effigies:
When the button is visible, it can be pushed down a bit and get cropped by the bottom of the window:

image

Also, the "I Agree" seems like it should have a non-breaking space or for the text to be centered in the button.

Would it be possible to allow the text to scroll when the window gets too narrow? I assume the goal of not increasing the height is to prevent the page contents from being covered.

effigies:
Overall this works smoothly. Attempting to browse and search works fine, but attempting to view a file or a download page is blocked until I press the button, at which point the buttons/tab are instantly made visible.

mgxd:
This looks good as is, but what about instead of hiding the `Download` tab, it is always present but is un-clickable until the user agrees? I think that'll make it clear that the download is always there, but blocked.

nellh:
> This looks good as is, but what about instead of hiding the `Download` tab, it is always present but is un-clickable until the user agrees? I think that'll make it clear that the download is always there, but blocked.

Thanks, I added this.

image

nellh:
Text and styling for mobile is done now.

image

effigies (approved):

PR #2969, opened 2024-01-12
Add codespell: workflow, config + get more typos fixed

yarikoptic:
In the past there were just PRs with results of applying codespell, not to use it in CI.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,6 @@
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
 [codespell]
-skip = .git,*.pdf,*.svg,go.sum,*.lock,.codespellrc
+skip = .git,*.pdf,*.svg,go.sum,*.lock,.codespellrc,.yarn,node_modules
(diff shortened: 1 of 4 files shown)


.pnp.cjs
```diff
@@ -40592,7 +40592,7 @@ function makeApi(runtimeState, opts) {
         if (result === false) {
           throw makeError(
             ErrorCode.BUILTIN_NODE_RESOLUTION_FAILED,
-            `The builtin node resolution algorithm was unable to resolve the requested module (it didn't go through the pnp resolver because the issuer was explicitely ignored by the regexp)
+            `The builtin node resolution algorithm was unable to resolve the requested module (it didn't go through the pnp resolver because the issuer was explicitly ignored by the regexp)
```
effigies: L40595 makes me think this is auto-generated. Probably want to add this file to the skip list.

yarikoptic: I ignored that file now

effigies (approved):

PR #2970, opened 2024-01-12
Add into User Guide preamble pointers to ReproNim 5 steps and OREONI

yarikoptic:
Following our discussion at ReproNim EAB, I am daring to suggest a possible point to link into ReproNim and possibly link/guide OpenNeuro users into better practices and to position BIDS and OpenNeuro to become an inherent part of their workflow and not just a dump of data.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -2,9 +2,13 @@
 
 Welcome to OpenNeuro! This guide is dedicated to new users to the site who would like to get started on using the [OpenNeuro web platform](https://openneuro.org/) to freely store and share [Brain Imaging Data Structure (BIDS) datasets](https://bids.neuroimaging.io/).
 
-We hope that OpenNeuro would become a useful resource for your research projects well beyond being merely a dump of your data to satisfy mandates of the funding agencies.
-We invite you to e.g. review [ReproNim 5 Steps](https://www.repronim.org/5steps) and [OREONI](https://oreoni.github.io) for a guidance and motivation on how to make your neuroimaging research more efficient and reproducible.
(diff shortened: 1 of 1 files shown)


docs/user_guide.md
```diff
@@ -2,6 +2,10 @@
 
 Welcome to OpenNeuro! This guide is dedicated to new users to the site who would like to get started on using the [OpenNeuro web platform](https://openneuro.org/) to freely store and share [Brain Imaging Data Structure (BIDS) datasets](https://bids.neuroimaging.io/).
 
+We hope that OpenNeuro would become a useful resource for your research projects well beyond being merely a dump of your data to satisfy mandates of the funding agencies.
+We invite you to e.g. review [ReproNim 5 Steps](https://www.repronim.org/5steps) and [OREONI](https://oreoni.github.io) for a guidance and motivation on how to make your neuroimaging research more efficient and reproducible.
+
```
effigies:
```suggestion
OpenNeuro is a data archive that follows the [FAIR principles for data management](https://www.go-fair.org/fair-principles/),
and which we hope can be a useful resource for your research projects.
In this guide, we describe browsing, downloading, uploading and managing OpenNeuro datasets.

For additional guidance and motivation on reproducible neuroimaging research,
we recommend the [Open and reproducible neuroimaging](https://oreoni.github.io/) guide
and the [ReproNim 5 Steps](https://www.repronim.org/5steps).
```

effigies (commented):
An alternative suggestion for the text. I think it would also make sense for us to expand this guide to include "How to make your dataset more findable", where we could link out to NeuroBagel and other annotation/query tools. Though that should probably be another PR.

Issue #2873, opened 2023-08-17
Misformed HTML in README results in react error


effigies:
### What went wrong?

A user added Markdown with `<br></br>` in it through the web README editor. The dataset now shows:

`Error: Minified React error #137; visit https://reactjs.org/docs/error-decoder.html?invariant=137&args[]=br for the full message or use the non-minified dev environment for full errors and additional helpful warnings.`

Following the [URL](https://legacy.reactjs.org/docs/error-decoder.html/?args%5B%5D=br&invariant=137):

> ``br is a void element tag and must neither have `children` nor use `dangerouslySetInnerHTML`.``

### Expected behavior

The retrieval process should probably shield contents from React inspection, or fall back to preformatted text.

### How to reproduce

1. ~~https://openneuro.org/datasets/ds004694 to see error~~
2. ~~`datalad clone https://openneuro.org/git/2/ds004694` to fetch the dataset and `git show` the latest change.~~
3. Add `<br></br>` to a dataset's README on https://staging.openneuro.org. e.g., https://staging.openneuro.org/datasets/ds001057

Edit: The offending code is resolved in the original datasets. You can still look at the history, or just look at the dataset on staging.

### Desktop

Irrelevant

### Phone

Irrelevant

### Additional information

![image](https://github.com/OpenNeuroOrg/openneuro/assets/83442/40708d3f-b5d3-4169-98e6-cebe8020718b)

PR #2982, opened 2024-01-24
fix(app): Use our own Markdown component that sanitizes HTML following GitHub-like rules

nellh:
This passes through the raw HTML strings instead of parsing it as JSX if present. All HTML will be displayed as raw text.

See #2873


packages/openneuro-app/src/scripts/utils/__tests__/markdown.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const { asFragment } = render(<Markdown>{hrefExample}</Markdown>)
+    expect(asFragment()).toMatchSnapshot()
+  })
+  it("does not allow href with unknown protocols", () => {
+    const hrefExample =
+      '<a href="about:memory">Example link that should not work.</a>'
+    const { asFragment } = render(<Markdown>{hrefExample}</Markdown>)
+    expect(asFragment()).toMatchSnapshot()
+  })
+})
```
effigies, line 35: Should we check the close-br specifically?

```suggestion
  it("filters close-break tags", () => {
    const hrefExample =
      '<br>sample text</br>'
    const { asFragment } = render(<Markdown>{hrefExample}</Markdown>)
    expect(asFragment()).toMatchSnapshot()
  })
})
```

packages/openneuro-app/src/scripts/utils/__tests__/__snapshots__/markdown.spec.tsx.snap
```diff
@@ -0,0 +1,47 @@
...
+      <br />
+      <ul>
+        <li>
+          test content
+        </li>
+      </ul>
+    </li>
+  </ul>
+</DocumentFragment>
+`;
```
effigies, line 57:
```suggestion
`;

exports[`Test <Markdown> component > filters close-break tags 1`] = `
<DocumentFragment>
  <br>
  sample text
</DocumentFragment>
`;
```

nellh:
```
exports[`Test <Markdown> component > filters close-break tags 1`] = `
<DocumentFragment>
  <span>
    <br />
    sample text
    <br />
  </span>
</DocumentFragment>
`;
```

This is the result for that input, kind of an interesting choice for fixing that syntax.

effigies: It's probably closer to the original intent, so that's fair. I was just making a guess as to what it would look like.

effigies (approved):

PR #3009, opened 2024-02-29
fix(app): Update to React 18 and replace jsx-to-markdown with react-markdown

nellh:
Replaces #3000 by replacing the library used with the more capable react-markdown.

The main extra change needed here was updating to React 18.


packages/openneuro-app/src/scripts/utils/__tests__/__snapshots__/markdown.spec.tsx.snap
```diff
@@ -2,56 +2,68 @@
...
-        <li>
-          test content
-        </li>
-      </ul>
     </li>
+    
+
+    <li>
+      test content
+    </li>
```
effigies, line 64: This seems to lose a level of nesting. Is this a reasonable interpretation of the input markdown?

nellh: This is the input.
```
* Markdown document
<br><br>
 * test content
```
react-markdown is more accurate to GitHub's behavior (not nesting) since the second `*` is not indented as much as the content of the first list item.

effigies (approved):
Lgtm, then.

PR #3017, opened 2024-03-19
Deno command line tools

nellh:
Adds command line tools for upload and download using Deno.

Some known issues:
- [ ] Hashing a dataset with a lot of annexed objects can take a while and doesn't show progress while hashing
- [x] Broken symlinks throw an error
- [x] Download only downloads git objects
- [x] Needs CI for deno tests


nellh:
Merging this so I can publish it on deno.land and I'll break out the remaining issue as a new issue.

PR #3025, opened 2024-04-05
DOC: Recommend "provider = generic" in .gitconfig

effigies:
At least when credential.credentialStore == secretservice, failing to include this produces a large number of sections like:

```cfg
[credential "https://openneuro.org/git/0/ds002278"]
provider = generic
```

Adding this prevents that duplication. I am not positive that this has no knock-on effects.


nellh:
I added an explanation of what the provider field is useful for so that it can be searched for if someone has this problem. It doesn't look like it will cause any issues.

PR #3032, opened 2024-04-23
deps: Support Node 20

nellh:
Drops the MongoDB vitest plugin due to lack of updates for it. Instead calls MongoDB memory server directly as needed. Upgrades to the latest version of Vitest and the 6.12 branch of Mongoose to support this.


nellh:
Deno test failures here are unrelated.

PR #3052, opened 2024-05-08
Bust cache for draft files when file exists on S3

### LINKED ISSUE #3047: Viewing participants.tsv in draft shows old versions of file
### What went wrong?

https://openneuro.org/datasets/ds004889/file-display/participants.tsv has been recently updated and published.

The snapshot shows updated contents (https://openneuro.org/datasets/ds004889/versions/1.1.2/file-display/participants.tsv) but the draft does not. Deleting the cache does not fix the issue, or perhaps the cache isn't getting deleted by the button.

I've checked and AWS and GitHub both have updated files. CHANGES and dataset_description.json both show the (old) 1.1.0 versions in the web UI, as well.

### Expected behavior

The draft should show the contents of the file as it exists on the server.

### How to reproduce

1. Go to https://openneuro.org/datasets/ds004889/file-display/participants.tsv 
2. See old version

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

### Additional information

_No response_

nellh:
Fixes #3047. This cache was created with the key "HEAD" and the cached value would be returned if the file existed on S3. Instead this caches the draft with the commit hash. This does mean an extra worker API call for each file during a download, so added a short lived cache to avoid bulk file requests from repeating this request to the worker.

### DIFF
--- a/packages/openneuro-server/src/cache/types.ts
+++ b/packages/openneuro-server/src/cache/types.ts
@@ -12,4 +12,5 @@ export enum CacheType {
   snapshotIndex = "snapshotIndex",
   participantCount = "participantCount",
   snapshotDownload = "download",
+  draftRevision = "revision",
 }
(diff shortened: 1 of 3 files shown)


effigies:
Makes sense to me.

Issue #3046, opened 2024-05-07
Transition from per-site analysis buttons to collapsable "Analyze" button


effigies:
### What would you like to see added?

Increasingly many third parties are providing their users access to OpenNeuro data. Eventually we're going to run out of button space. Another approach could be to do something like we do with the "Clone" drop-down. This would also give us more space to explain what the services are.

## Current

![image](https://github.com/OpenNeuroOrg/openneuro/assets/83442/1873b45a-70e6-49eb-be9f-767e6900e618)

## Proposed

![image](https://github.com/OpenNeuroOrg/openneuro/assets/83442/e0c00897-e7b1-4a62-a53a-ebe60cdd4264)

### Alternatives

Another approach would be to add another tab, like "Derivatives".

### Do you have any interest in helping implement the feature?

Yes, but I would need guidance

nellh:
I was thinking of implementing something like this after the discussion yesterday. Looks good to me.

prioux:
I like the new Analyse menu. It will work for at least a dozen new services. When there are more, then we'll need to think of something else, I guess.

prioux:
Also, the Analyse menu is the perfect place to put at the top a short disclaimer (`these are external sites blah blah blah not responsible if they reboot your data center`)

poldrack:
the new menu looks good to me as well!

surchs:
Looks great! Is the plan to make this button for anything "analysis" or could it be a more generic "do more" button for anything third-party related? I'm mainly asking because we're working on providing people with nice tools to add annotations to (for now forks of) OpenNeuro datasets over in https://github.com/OpenNeuroDatasets-jsonld and it would be great if we could eventually add a "Contribute annotations to this dataset" button into such a dropdown.

bryancaron:
Hi all, just checking in to see if there is any update regarding this item and timeline for implementation?  I'm particularly thinking in regard to the inclusion of a link to CBRAIN as a resource provider to analyze the OpenNeuro datasets.  Thanks!

effigies:
The update is that we're good to go ahead with it. We do not have a definite timeline yet.

@surchs I'm not opposed to "Do more" or similar. We can think about what looks good and makes sense as this is implemented, but we won't be permanently tied to any particular button text.

nellh:
This has been deployed in OpenNeuro 4.25.0.

PR #3077, opened 2024-06-12
Snapshot page analyze dropdown menu for external tools

nellh:
Implements #3046

image


packages/openneuro-app/src/scripts/dataset/components/AnalyzeDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export const AnalyzeDropdown: React.FC<CloneDropdownProps> = (
+  { datasetId, snapshotVersion },
+) => {
+  const brainlifeUrl =
+    `https://brainlife.io/openneuro/${datasetId}/${snapshotVersion}`
+  const nemarUrl =
+    `https://nemar.org/dataexplorer/detail?dataset_id=${datasetId}`
+  const cbrainUrl =
+    `https://portal.cbrain.mcgill.ca/openneuro/${datasetId}/versions/${snapshotVersion}`
```
effigies, line 26: Looks like you found it. Sorry for completely forgetting about this.

packages/openneuro-app/src/scripts/dataset/components/AnalyzeDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <div className="dataset-git-access">
+          <span>
+            <h4>Analyze this dataset</h4>
+          </span>
+          <p>
+            Work with OpenNeuro datasets using any of these third-party tools.
+          </p>
+          <hr />
+          <img
+            src="https://brainlife.io/favicon.ico"
```
effigies: Should we host our own copies, so they can rearrange their sites without breaking this?

effigies (approved):
Test failure will be a quick fix. One small question.

packages/openneuro-app/src/scripts/dataset/components/AnalyzeDropdown.tsx
nellh: There's trade offs either way but I think it does make sense. I'll make that change.

PR #3113, opened 2024-08-07
deps(worker): Update to pygit2 0.15.1

nellh:
Resolves breaking changes from pygit2 0.14 -> 0.15.

https://github.com/libgit2/pygit2/blob/master/CHANGELOG.md#1150-2024-05-18


effigies (approved):
This is a deeply obnoxious refactor, as literally every object can be `str()`'d. I started looking into it a while back and backed off when I realized it wasn't all `X.hex -> str(X)`. So hats off to you.

There's one line missing coverage. Are you confident in that one?

nellh:
> There's one line missing coverage. Are you confident in that one?

Added a new test to check for the expected tag reference to cover that change.

PR #3170, opened 2024-10-07
feat(app): Generate a parent directory for dataset download scripts

nellh:
Adds test coverage for this component and the changes to the directory structure.

Instead of `stimuli/meg/f074.bmp` we save to `ds0001234-1.0.0/stimuli/meg/f074.bmp`.


effigies (approved):
LGTM

PR #3179, opened 2024-10-22
Linting refactor and enable GitHub actions

nellh:
The linter hasn't been run on CI in a while and some of the rules in use do not make sense for the current version of the project. This PR replaces the linting config and includes all the required fixes for the new ruleset.

This also adds the new react-compiler beta linter, which can do a better job of catching misuse of React and in the future may be useful for automatic memoization of component rendering.

TODO:
- [x] About 100 errors left
- [x] GitHub action


eslint.config.mjs
```diff
@@ -0,0 +1,53 @@
...
+  ignores: [
+    "**/build/**",
+    "**/dist/**",
+    "**/node_modules/**",
+    ".yarn/*",
+    ".pnp.cjs",
+    ".pnp.loader.mjs",
+    "**/coverage/**",
+    "packages/openneuro-app/pluralize-esm.js",
+    "packages/openneuro-app/src/scripts/utils/schema-validator.js",
```
effigies, line 16: Might be helpful to future readers to know why we're ignoring these. I think this is the reason?

```suggestion
    // Generated from external sources
    "packages/openneuro-app/pluralize-esm.js",
    "packages/openneuro-app/src/scripts/utils/schema-validator.js",
```

eslint.config.mjs
```diff
@@ -0,0 +1,53 @@
...
+  ],
+}, {
+  files: [
+    "packages/**/*.ts",
+    "packages/**/*.tsx",
+    "packages/**/*.jsx",
+    "packages/**/*.js",
```
effigies, line 23: Is it reasonable for this thing to lint itself?

```suggestion
    "packages/**/*.js",
    "eslint.config.mjs",
```

packages/openneuro-app/src/scripts/dataset/fragments/dataset-history.jsx
```diff
@@ -80,7 +79,6 @@ const DatasetHistory = ({ datasetId }) => {
                 </div>
                 <div className="col-lg col col-2 grid actions">
                   <Revalidate datasetId={datasetId} revision={commit.id} />
-                  <UpdateRef datasetId={datasetId} revision={commit.id} />
```
effigies, line 83: Why are we removing this feature? I have occasionally used it.

effigies (commented):
Partial review. Will finish tomorrow.

effigies:
Finished review. Main concern is dropping the `UpdateRef` mutation, but it's not an indispensable feature.

packages/openneuro-app/src/scripts/dataset/fragments/dataset-history.jsx
nellh: The mutation it uses hasn't worked since 2022. It would move the HEAD reference in MongoDB but we no longer do that, always following the git repo HEAD reference instead. The git version would be more destructive and I think it's better handled via the git interface itself.

effigies: Do we allow force-pushes to revert? If not, then it's just logging into the worker and manually doing the job.

Anyway, if it hasn't worked, then this is no loss, so no further objections.

PR #3184, opened 2024-10-25
feat: Pass custom config to validator

effigies:
In prep to migrate to the schema validator, we need to promote some warnings to errors.

Reminded that we need to do this by bids-standard/bids-specification#1928, which proposes to downgrade SUBJECT_FOLDERS from an error to a warning. I don't think OpenNeuro wants to downgrade that, so this preemptively pegs it as an error.


effigies:
xref https://github.com/bids-standard/bids-validator/pull/2176, since that's needed to support microscopy blacklisting properly.

nellh:
Included in #3206

Issue #3186, opened 2024-10-31
NeuroSift viewer


effigies:
### What would you like to see added?

[NeuroSift](https://neurosift.app) is a viewer for electrophysiology data that can readily stream from S3 sources. This could be used in OpenNeuro, which lacks an ephys viewer.

An example URL using an OpenNeuro S3 file: https://neurosift.app/?p=/edf&url=https://s3.amazonaws.com/openneuro.org/ds003374/sub-01/ses-01/ieeg/sub-01_ses-01_task-jokeit_run-01_ieeg.edf?versionId=fmme9zou6CJaupJWGNfKx54PzDIruE1f&embedded=1

This can be embedded in an iframe, which is the current best practice for integration.

I would estimate this to be fairly low-effort with a high return in terms of UX. Do we collect any stats on attempts to view files in-browser?

We would need a list of file types that can be handled. It appears EDF and NWB, but I'm not immediately finding documentation on other formats.

cc @magland

### Alternatives

None

### Do you have any interest in helping implement the feature?

No

magland:
Thanks @effigies 

In terms of which file types are supported... yes, this would just be EDF and NWB (do you have any NWB on OpenNeuro?)

effigies:
A quick `find` didn't find any, but they are permitted under BIDS, so I would expect that we will get some eventually.

nellh:
I've implemented this in #3187 using the iframe embed approach.

PR #3187, opened 2024-10-31
feat(app): Add Neurosift viewer for EDF and NWB files

nellh:
Implementation of #3186 using the iframe approach.

Screenshot From 2024-10-31 10-22-40

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/files/file-viewer-type.jsx
+++ b/packages/openneuro-app/src/scripts/dataset/files/file-viewer-type.jsx
@@ -6,6 +6,7 @@ import FileViewerJson from "./viewers/file-viewer-json.jsx"
 import FileViewerTsv from "./viewers/file-viewer-tsv.jsx"
 import FileViewerCsv from "./viewers/file-viewer-csv.jsx"
 import FileViewerHtml from "./viewers/file-viewer-html.jsx"
+import { FileViewerEdf } from "./viewers/file-viewer-edf"
 
 /**
  * Choose the right viewer for each file type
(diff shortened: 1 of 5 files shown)


magland:
Thanks @nellh 

This won't work as is for NWB files. The URL for those looks a bit different:

https://neurosift.app/?p=/nwb&url=https://api.dandiarchive.org/api/assets/027567cf-e9e0-4aba-83c0-ef2d22c8c1b7/download/

Notice the p=/nwb instead of p=/edf

magland:
I wonder what's the best way to select the height. Is the plan to just fix it as something sensible?

nellh:
> Thanks @nellh
> 
> This won't work as is for NWB files. The URL for those looks a bit different:
> 
> https://neurosift.app/?p=/nwb&url=https://api.dandiarchive.org/api/assets/027567cf-e9e0-4aba-83c0-ef2d22c8c1b7/download/
> 
> Notice the p=/nwb instead of p=/edf

Thanks, fixed!

> I wonder what's the best way to select the height. Is the plan to just fix it as something sensible?

I set it to 50% viewport height as this is what our HTML iframe viewer was doing already and generally gives a good amount of space for the viewer without using a static height.

magland:
This is great, thanks @nellh . When will this go live?

nellh:
> This is great, thanks @nellh . When will this go live?

It will be included in the next OpenNeuro release, aiming to release early next week.

magland:
@nellh I think there's a problem here. It seems that when you visit the page of an EDF file, the entire file gets automatically downloaded to the browser before the iframe is rendered. This is probably because that's how other file types are handled. But in this case, we don't need the file to be downloaded - and that's the point of neurosift - that's it's lazy-loading. Sometimes the files can be quite large.

PR #3196, opened 2024-11-07
Allow sharing permissions by ORCID

### LINKED ISSUE #3190: Allow permissions to be shared by ORCID
Administrative permissions should require an email so that we can't get to a state with datasets with uncontactable authors.

If we have not already made it impossible for dataset admins to remove the last admin, that could probably be done at the same time.

nellh:
Adds the option to add dataset permissions via ORCID.

Fixes #3190.

Screenshot From 2024-11-07 13-18-42


thinknoack:
@nellh - I know you are updating a few tests so send this back when you have those.

in addition here is a small note, when removing perms, the client/ui doesn't update the state for visual feedback until a refresh. 

After a confirmed removal 
##
Screenshot 2024-11-08 at 12 39 04 PM

After refresh
##
Screenshot 2024-11-08 at 12 39 19 PM

nellh:
> in addition here is a small note, when removing perms, the client/ui doesn't update the state for visual feedback until a refresh.

Thanks! That's because the remove mutation has the same issue we discussed where the API should return the entire fragment it modifies (all permissions instead of just the user) to update the client cache. I think that's a good bug to note (#3198) now that this generally fixes the client state issues with adding/removing permissions.

nellh:
@thinknoack Added the test case we discussed in 85ddbfc.

thinknoack:
*Approving with small note

I have reviewed the code/file changes and tested with emails for both a new user and a previous user. That all looks great. 

*note - I will not be able to test the ORCID integration until I get my sandbox sorted out.

thinknoack (approved):
*Approving with small note

I have reviewed the code/file changes and tested with emails for both a new user and a original upload user. That all looks great.

*note - I will not be able to test the ORCID integration until I get my sandbox sorted out.

PR #3227, opened 2024-11-19
Vite config update

thinknoack:
minor changes to docker-compose and vite.config

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,8 +33,6 @@ services:
       - "9992:9992"
     environment:
       - NODE_ENV=development
-    volumes:
-      - ./:/srv:rw,z
     depends_on:
       node:
         condition: service_healthy
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,8 @@ services:
       - "9992:9992"
     environment:
       - NODE_ENV=development
+    volumes:
+      - ./:/srv:rw,z
     depends_on:
       node:
         condition: service_healthy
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
       - "9992:9992"
     environment:
       - NODE_ENV=development
+    volumes:
+      - ./:/srv:rw,z
```
nellh:
```suggestion
```

This should be redundant with the app config later.

packages/openneuro-app/vite.config.js
```diff
@@ -27,6 +27,10 @@ export default defineConfig({
     port: 80,
     host: "0.0.0.0",
     cors: true,
+    watch: {
+      usePolling: true, // Enable polling for file watching
```
nellh: This has a pretty significant performance cost if it isn't needed for a large project like this. Might be better to use these overrides on macOS and document it for developer setup instead.

In `.env`:
```
CHOKIDAR_USEPOLLING=true
CHOKIDAR_INTERVAL=1000
```

nellh (changes requested):

docker-compose.yml
thinknoack: this has been removed

packages/openneuro-app/vite.config.js
thinknoack: removed this from vite.config and added polling config to docker-compose and config.env

nellh (approved):

PR #3226, opened 2024-11-19
fix(app): Prevent loading files for viewers with built in data transfer

nellh:
Skip data requests for some files when the viewer includes data loading.

Fixes #3225

PR #3230, opened 2024-11-20
fix(app): prevent unnecessary download of EDF/NWB data to browser

magland:
Recently support for viewing EDF and NWB files was added by embedding a Neurosift iframe
https://github.com/OpenNeuroOrg/openneuro/pull/3187

However I recently noticed that when navigating to such a file, the entire content of the file gets downloaded to the browser before the view is rendered. This is unnecessary (and potentially very resource demanding) as neurosift handles efficient lazy-loading of the file without downloading the entire thing.

This PR solves the issue by conditionally skipping the fetch step in the case of these file extension types.

tag: @nellh

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/files/file-view.jsx
+++ b/packages/openneuro-app/src/scripts/dataset/files/file-view.jsx
@@ -1,13 +1,18 @@
 import React, { useEffect, useState } from "react"
 import { Loading } from "@openneuro/components/loading"
 import FileViewerType from "./file-viewer-type.jsx"
-import { isNifti, isNwb } from "./file-types"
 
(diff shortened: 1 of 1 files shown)


nellh:
@magland Thanks for the contribution but I made a similar fix based on your comment and it was merged in #3226 for inclusion in our next release. This also affected Niivue, thank you for catching the issue.

magland:
> @magland Thanks for the contribution but I made a similar fix based on your comment and it was merged in #3226 for inclusion in our next release. This also affected Niivue, thank you for catching the issue.

Ah okay, great. I was able to set up a dev environment (for the frontend at least) so that was a plus. :)

PR #3242, opened 2024-12-02
User routes

thinknoack:


packages/openneuro-app/src/scripts/users/user-account-container.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+
+const UserAccountContainer: React.FC<AccountContainerProps> = ({ user }) => {
+
+  return (
+    <>
+        <div className="user-info">
+            <div className="user-account-card">{user.id}</div>
+        <div className="user-account-tab-links">ComponentsLinks to datasets, account, and notifications
+            <Link to={'account'}> Test </Link>
```
nellh:
```suggestion
            <Link to="account"> Test </Link>
```

packages/openneuro-app/src/scripts/users/user-query.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * Routing wrapper for USER query
+ *
+ * Expects to be a child of a react-router Route component with UserId
+ */
+const UserQuery = () => {
+    const { user } = useParams()
+    return (
+        <>
+            <UserQueryHook user={user} />
+        </>
```
nellh:
```suggestion
        <UserQueryHook user={user} />
```

packages/openneuro-app/src/scripts/routes.tsx
```diff
@@ -33,6 +35,7 @@ const AppRoutes: React.VoidFunctionComponent = () => (
     <Route path="/import" element={<ImportDataset />} />
     <Route path="/metadata" element={<DatasetMetadata />} />
     <Route path="/public" element={<Navigate to="/search" replace />} />
+    <Route path="/account/:orchid/*" element={<UserQuery/>} />
```
nellh:
```suggestion
    <Route path="/account/:orcid/*" element={<UserQuery/>} />
```

packages/openneuro-app/src/scripts/users/user-routes.tsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from "prop-types"
+import { Route, Routes } from "react-router-dom"
+import UserAccountContainer from "./user-account-container"
+
+const DatasetRoutes = ({ user, error }) => {
+  useEffect(() => {
+    if (error) {
+      throw error
+    }
+  }, [user, error])
```
nellh:
```suggestion
```
I don't think this is actually useful. The error will bubble up anyways.

packages/openneuro-app/src/scripts/users/user-routes.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect } from "react"
+import PropTypes from "prop-types"
+import { Route, Routes } from "react-router-dom"
+import UserAccountContainer from "./user-account-container"
+
+const DatasetRoutes = ({ user, error }) => {
```
nellh:
```suggestion
export const UserRoutes = ({ user, error }) => {
```

packages/openneuro-app/src/scripts/users/user-routes.tsx
```diff
@@ -0,0 +1,29 @@
...
+    </Routes>
+  )
+}
+
+DatasetRoutes.propTypes = {
+  dataset: PropTypes.object,
+  error: PropTypes.object,
+}
+
+export default DatasetRoutes
```
nellh:
```suggestion
```
We have a lot of default exports but generally better to use named exports for components.

packages/openneuro-app/src/scripts/users/user-query.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
+import * as Sentry from '@sentry/react'
+import PropTypes from 'prop-types'
+import { useParams } from 'react-router-dom'
+import { Loading } from '@openneuro/components/loading'
+import UserRoutes from './user-routes'
```
nellh:
```suggestion
import { UserRoutes } from './user-routes'
```

nellh (commented):
Looks like this is on the right track. Added some suggestions but it's mostly minor fixes.

packages/openneuro-app/src/scripts/users/user-query.tsx
thinknoack: I've updated files I created to be named exports. Should I go ahead and do the same for ones I see like `<FourOFour />` etc.

thinknoack:
> Looks like this is on the right track. Added some suggestions but it's mostly minor fixes.

ok, lint/ts issues have been resolved.

packages/openneuro-app/src/scripts/users/__tests__/user-account-view.tsx
```diff
@@ -0,0 +1,79 @@
...
+    links: ["https://example.com", "https://example.org"],
+    github: "johndoe",
+  };
+
+describe('<UserAccountView />', () => {
+  it('should render the user details correctly', () => {
+    render(<UserAccountView user={baseUser} />);
+
+    // Check if user details are rendered
+    expect(screen.getByText('Name:')).toBeInTheDocument();
```
nellh: I think checking for this by name is fine but generally looking for the dynamic data is going to be more important and this more tests than needed can be brittle.

packages/openneuro-app/src/scripts/users/__tests__/user-account-view.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should render links with EditableContent', () => {
+    render(<UserAccountView user={baseUser} />);
+
+    // Editable Links section
+    const linksSection = screen.getByText('Links');
+    expect(linksSection).toBeInTheDocument();
+
+    const linkItems = screen.getAllByRole('listitem');
+    expect(linkItems).toHaveLength(2); // initially 2 links
```
nellh: Get by role here is great, this is a good example of a test that checks the functionality of the component.

packages/openneuro-app/src/scripts/users/__tests__/user-account-view.tsx
nellh, line 1: By convention this should be user-account-view.spec.tsx. These files should probably all be .spec.tsx to match the rest of the project.

nellh (approved):
Approved with minor changes for the test filenames.

packages/openneuro-app/src/scripts/users/user-query.tsx
```diff
@@ -0,0 +1,81 @@
+// UserQuery.tsx
+
+import React from 'react';
+import { useParams } from 'react-router-dom';
+import { UserRoutes } from './user-routes';
+import FourOFourPage from '../errors/404page';
+
+// ORCID validation regex pattern
+const orcidPattern = /^\d{4}-\d{4}-\d{4}-\d{4}$/;
```
nellh: The ORCID can end in X as the last value is a checksum.

We should reuse this implementation from here:
https://github.com/OpenNeuroOrg/openneuro/blob/08e28d29991668fd032aabbeb1f2185d3b24238c/packages/openneuro-app/src/scripts/dataset/mutations/update-permissions.tsx#L10-L18

nellh (changes requested):

packages/openneuro-app/src/scripts/users/__tests__/user-account-view.tsx
thinknoack: will update with query by orcid

packages/openneuro-app/src/scripts/users/__tests__/user-account-view.tsx
thinknoack: updated

packages/openneuro-app/src/scripts/users/user-query.tsx
thinknoack: added new validationUtils.ts that is now available to on-app

PR #3258, opened 2024-12-17
update resolver to query user by orcid or id

thinknoack:
updating the resolver so a user can be queried by orcid or id.


thinknoack:
@nellh - I think I have this as we discussed. I will need some help to write test coverage for this
`packages/openneuro-server/src/graphql/resolvers/user.ts` what I tried was not working as expected. 

also I added the gql user query in the user-query.tsx  and I am writing/updating tests for that. I will also need to adjust templates to account for fields that are not required (like avatar). Putting back into draft but would love your feedback.

packages/openneuro-server/src/graphql/resolvers/user.ts
```diff
@@ -3,8 +3,20 @@
...
+
 export const user = (obj, { id }) => {
-  return User.findOne({ id }).exec()
+  
+  function isValidOrcid(orcid: string): boolean {
+    return /^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$/.test(orcid || '');
+  }
+
+  if (isValidOrcid(id)) {
+    return User.findOne({ "orchid": id }).exec()
```
nellh:
```suggestion
    return User.findOne({ "orcid": id }).exec()
```

nellh: One typo here but another limitation is there's two kinds of user records here:

```json
{
  "provider": "orcid",
  "providerId": "0000-0000..."
}
```

```json
{
  "provider": "google",
  "orcid": "0000-0000..."
}
```

This branch needs to find both of these. Once we've migrated Google users to ORCID logins, most users will fall into the first group that won't match here.

nellh (changes requested):
Most of this looks good. The precommit hook didn't get run so the formatting is wrong in some spots. One fix needed for the functionality but it would also be good to fix the formatting.

packages/openneuro-server/src/graphql/resolvers/user.ts
thinknoack: updated typo.

thinknoack: @nellh something like this?

```
export const user = (obj, { id, key }) => {
  function isValidOrcid(orcid: string): boolean {
    return /^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$/.test(orcid || "")
  }

  if (isValidOrcid(id)) {
    if (key === 'orcid') {
      return User.findOne({ "orcid": id }).exec(); // Search by orcid if key is 'orcid'
    } else {
      return User.findOne({ "providerId": id }).exec(); // Otherwise, search by providerId
    }
  } else {
    return User.findOne({ "id": id }).exec(); // Search by id if not a valid ORCID
  }
}
```

nellh: It should be a single query, you can use the `{$or: [{ "orcid": id }, {"providerId": id}]}` operator to check if the id is in either field. No need for a key argument.

thinknoack: ty - that was helpful. I have that added now

nellh (approved):

PR #3268, opened 2025-01-06
adding workaround for elastic search container on mac

thinknoack:


nellh:
Testing this it does break on anything other than Apple Silicon, so this will need to be moved to the environment file.

thinknoack:
> Testing this it does break on anything other than Apple Silicon, so this will need to be moved to the environment file.

I moved these to the config.env and updated the various files to correspond

PR #3272, opened 2025-01-09
update user schema to include new fields and ability to updateUser

thinknoack:
This contains the editing feature for user content on the about page. It also adds the hasEdit, loggedIn, and admin checks.


nellh (approved):
Tested locally with several kinds of users and it looks like it's in a pretty good state.

PR #3279, opened 2025-01-15
User datasets, general styling and layouts, getting query started, and various filtering and sorting

thinknoack:


packages/openneuro-app/src/scripts/users/components/dataset-card.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+  stars: [{ userId: string; datasetId: string }]
+  followers: [{ userId: string; datasetId: string }]
+  latestSnapshot?: {
+    id: string
+    size: number
+    issues: [{ severity: string }]
+    created?: string
+  }
+}
```
nellh: Not that it needs to be changed in this PR but we should probably try to generate these types from the GraphQL schema as we're repeating this type definition quite a bit.

packages/openneuro-app/src/scripts/users/user-datasets-view.tsx
```diff
@@ -7,47 +11,148 @@ interface User {
...
+  }
+  stars: [{ userId: string; datasetId: string }]
+  followers: [{ userId: string; datasetId: string }]
+  latestSnapshot?: {
+    id: string
+    size: number
+    issues: [{ severity: string }]
+    created?: string
+  }
 }
```
nellh: This for example is the same type repeated. I think at least within the context of the user components, these should be consolidated. I'd move these to a user-types.ts file for now and leave any generation from GraphQL and merger with other frontend uses of these types to a separate PR.

packages/openneuro-app/src/scripts/users/user-datasets-view.tsx
```diff
@@ -7,47 +11,148 @@ interface User {
...
+            }
+            description {
+              Authors
+            }
+          }
+        }
+      }
+    }
+  }
+`
```
nellh, line 21: Can this reuse any of the query fragments we already have defined?

https://github.com/OpenNeuroOrg/openneuro/blob/977d5a1facbc48bc9a8af027a1511f6a1517c425/packages/openneuro-app/src/scripts/datalad/dataset/dataset-query-fragments.js

packages/openneuro-app/src/scripts/users/user-datasets-view.tsx
```diff
@@ -7,47 +11,148 @@ interface User {
...
+          const aUpdated = a.latestSnapshot?.created || a.created
+          const bUpdated = b.latestSnapshot?.created || b.created
+          result = new Date(bUpdated).getTime() - new Date(aUpdated).getTime()
+          break
+        }
+        default: {
+          result = 0
+          break
+        }
+      }
```
nellh: It would be a good idea to break out this sorting function and unit test it.

nellh (commented):
Looks good so far, I added a few comments with some suggestions.

packages/openneuro-app/src/scripts/users/user-datasets-view.tsx
thinknoack: moved into a new Util.

packages/openneuro-app/src/scripts/users/user-datasets-view.tsx
thinknoack: I moved this into a new Fragment I did not find one that I could reuse.

packages/openneuro-app/src/scripts/users/components/dataset-card.tsx
thinknoack: I made user-types its own file and put it in the types dir.

packages/openneuro-app/src/scripts/users/user-datasets-view.tsx
thinknoack: I made user-types its own file and put it in the types dir.

thinknoack:
@nellh I would love to request another review.  I think this is good except for the query for user datasets. I have some questions about that.

nellh (approved):

PR #3299, opened 2025-01-27
chore(build): Upgrade bids-validator for web UI

effigies:
We updated the validator for the CLI but not the web UI. Not sure if this is best practice, but I manually edited `packages/openneuro-app/package.json` and ran `npx yarn up`. It seems to have done something sensible.

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -3510,21 +3510,21 @@ const RAW_RUNTIME_STATE =
       }]\
     ]],\
     ["@jsr/bids__schema", [\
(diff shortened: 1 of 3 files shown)


nellh:
> We updated the validator for the CLI but not the web UI. Not sure if this is best practice, but I manually edited `packages/openneuro-app/package.json` and ran `npx yarn up`. It seems to have done something sensible.

I'd do it with `yarn workspace @openneuro/app dlx jsr add @bids/validator` but looks like the result is identical.

PR #3314, opened 2025-01-30
Brain Initiative Landing Page

thinknoack:
adds to the search ctx and params to search for datasets where brain_initiative == "true". 

also contains the various styling and components to accommodate the new NIH Brain Initiative portal.


nellh (approved):
I rebased and made one small change to how the value was being queried but this looks ready to go.

PR #3320, opened 2025-02-04
feat: Demote PARTICIPANT_ID_MISMATCH to a warning for now

effigies:
Addressing problems with https://openneuro.org/datasets/ds005752/ that can't be reasonably resolved within the dataset. This is quite a minor issue, and we can hopefully get rid of this exception soon.


effigies:
@nellh Not sure if 2be2eda is what you had in mind...

services/datalad/datalad_service/tasks/assets/validator-config.json
effigies, line 1: I guess we could symlink this one too, then?

nellh: I tried to get that to work but gave up on it. I've also had to go back to using a non-symlinked one for the JSR CLI, so there's just going to be several copies.

Issue #3313, opened 2025-01-30
Validator data can cause 404 on slow connections


nellh:
### What went wrong?

A more frequent issue with the new validator but there are some legacy validator examples. A snapshot page will load this data twice.

### Expected behavior

Avoid loading validator data twice and avoid loading it prior to display.

### How to reproduce

1. Simulate a slower connection (slow enough that it takes at least 60 seconds to load 10MB).
2. https://openneuro.org/datasets/ds005165/versions/1.0.4

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

nellh:
This is fixed by limiting the maximum size of validator data and loading only as needed.

PR #3322, opened 2025-02-05
Progressive loading for schema validator output

nellh:
This avoids loading the BIDS validator output for schema validator datasets, instead loading this once the validation drop down is clicked. For many datasets this avoids several megabytes of data transfer and potentially expensive API requests until this is actually needed.

A related bug fix is included to more accurately disable creating snapshots when the schema validator output contains errors or the draft has no author values provided.

See #3313


thinknoack (approved):
reviewing the app locally looks good. I see loading animations for both draft and snapshots. Code looks good, tests pass, and no lint issues.

Issue #2936, opened 2023-11-07
Allow linking of GitHub users to OpenNeuro users


nellh:
### What would you like to see added?

For some new GitHub features, we need to allow a user to connect their GitHub account to their OpenNeuro account. This should use GitHub SSO and maintain the link until the user disconnects the account.

This should probably happen after #180.

### Do you have any interest in helping implement the feature?

Yes

PR #3326, opened 2025-02-10
adding github strategy that syncs data from github to user

thinknoack:
adds github oauth strategy and a button to sync data from github to a users profile.


packages/openneuro-server/src/libs/authentication/crypto.ts
```diff
@@ -1,7 +1,7 @@
 import crypto from "crypto"
 import config from "../../config"
 
-const secret = config.auth.jwt.secret
+const secret = config.auth.jwt?.secret || ""
```
nellh: What's the purpose of this change? I think this would make more sense to keep it from being defined as an insecure string. For tests, the config can be mocked with a testing value.

packages/openneuro-server/src/libs/authentication/github.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        // Only update institution if it is not already set
+        if (!orcidUser.institution && profile._json.company) {
+          orcidUser.institution = profile._json.company
+        }
+
+        // If email is empty, set it from GitHub
+        if (!orcidUser.email && profile.emails?.length > 0) {
+          orcidUser.email = profile.emails[0].value
+        }
```
nellh: If we want to use GitHub as an authority for email, I think we need to check that the email used here is marked primary and verified in the response.

packages/openneuro-server/src/routes.ts
```diff
@@ -157,6 +163,48 @@ const routes = [
...
+
+  // GitHub OAuth callback route
+  {
+    method: "get",
+    url: "/auth/github/callback",
+    handler: (req, res, next) => {
+      passport.authenticate("github", (err, user) => {
+        if (err) {
+          // eslint-disable-next-line
+          console.error("GitHub Auth Error:", err)
```
nellh: It would be good to also capture a Sentry error here, especially useful for debugging auth issues in production.

packages/openneuro-server/src/routes.ts
```diff
@@ -157,6 +163,48 @@ const routes = [
...
+          return res.redirect("/login?error=github_auth_failed")
+        }
+        if (!user) return res.redirect("/login")
+
+        // Store JWT in a secure cookie
+        res.cookie("jwt", user.jwt, {
+          httpOnly: true,
+          secure: process.env.NODE_ENV === "production",
+          sameSite: "strict",
+        })
```
nellh: This looks like a bad idea, why is there a need to set the JWT on a cookie here? It should be set already on the accessToken cookie when this callback resolves.

packages/openneuro-server/src/libs/authentication/github.ts
```diff
@@ -0,0 +1,80 @@
...
+          orcidUser.links.push(profile.profileUrl)
+        }
+
+        orcidUser.githubSynced = new Date()
+
+        await orcidUser.save()
+
+        return done(null, addJWT(config)(orcidUser.toObject()))
+      } catch (err) {
+        return done(err, null)
```
nellh, line 101: This should probably also report to Sentry.

nellh (changes requested):

packages/openneuro-server/src/libs/authentication/crypto.ts
thinknoack: I was getting an error before while running tests. I since changed the test and forgot to remove this.

packages/openneuro-server/src/libs/authentication/github.ts
thinknoack: I am not sure on this, I think when I login via orcid my email was blank, it is probably best to use orcid for emails. I can remove this.

packages/openneuro-server/src/routes.ts
thinknoack: added sentry

packages/openneuro-server/src/routes.ts
thinknoack: ah, thanks. that has been removed.

packages/openneuro-server/src/libs/authentication/github.ts
thinknoack: Sentry added

packages/openneuro-server/src/routes.ts
thinknoack: is there a way to test Sentry?

packages/openneuro-server/src/libs/authentication/github.ts
thinknoack: because there isn't really a normal login - I also added toastify to alert users of issues/success.

thinknoack:
This PR implements basic linking for GH accounts. ticket REF #2936

packages/openneuro-app/src/scripts/users/github-auth-button.tsx
```diff
@@ -0,0 +1,93 @@
...
+      }
+      &.synced .accordion-title {
+        left: 435px;
+      }
+    }
+  }
+`
+
+export const GitHubAuthButton: React.FC<GitHubAuthButtonProps> = ({ sync }) => {
+  const buttonText = sync ? "Re-sync" : "Sync"
```
nellh: Maybe this should be "Link GitHub"?

packages/openneuro-server/src/libs/authentication/github.ts
```diff
@@ -0,0 +1,74 @@
...
+        const token = jwtFromRequest(req)
+        const decoded = token ? jwt.verify(token, config.auth.jwt.secret) : null
+
+        if (!decoded || !decoded.sub) {
+          return done(new Error("No authenticated ORCID user found"), null)
+        }
+
+        const orcidUser = await User.findOne({
+          id: decoded.sub,
+          provider: "orcid",
```
nellh:
```suggestion
```
Even though we will migrate users to ORCID accounts, this should work for any account that's authenticated so limiting it to provider ORCID is less flexible.

packages/openneuro-server/src/libs/authentication/github.ts
```diff
@@ -0,0 +1,74 @@
...
+      scope: ["user:email", "read:user"],
+      passReqToCallback: true,
+    },
+    async (req, accessToken, refreshToken, profile, done) => {
+      try {
+        const token = jwtFromRequest(req)
+        const decoded = token ? jwt.verify(token, config.auth.jwt.secret) : null
+
+        if (!decoded || !decoded.sub) {
+          return done(new Error("No authenticated ORCID user found"), null)
```
nellh: This error case should probably return a redirect to an error page, it can just be the access denied page.

packages/openneuro-server/src/libs/authentication/github.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        const orcidUser = await User.findOne({
+          id: decoded.sub,
+          provider: "orcid",
+        })
+
+        if (!orcidUser) {
+          return done(null, false, {
```
nellh, line 68: This can also be access denied.

packages/openneuro-app/src/scripts/users/github-auth-button.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    if (error === "orcid_required") {
+      toast.error("Please login with your ORCID account")
+    } else if (error) {
+      toast.error(error.replace(/_/g, " ")) // Replace underscores with spaces for readability
+    }
+
+    const success = searchParams.get("success")
+    if (success === "github_auth_success") {
+      toast.success("GitHub data successfully synced!")
```
nellh:
```suggestion
      toast.success("GitHub linked!")
```
Linked or connected here too?

packages/openneuro-app/src/scripts/users/github-auth-button.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <AccordionWrap>
+        <AccordionTab
+          accordionStyle="plain"
+          label="?"
+          className={lastSyncedText
+            ? "keyword-accordion synced"
+            : "keyword-accordion"}
+        >
+          <span>
+            Here is some text for a description and/or policy for syncing data
```
nellh:
```suggestion
            Sync profile data from GitHub (avatar, institution, or location).
```

nellh (changes requested):

packages/openneuro-app/src/scripts/users/github-auth-button.tsx
thinknoack: this changes a portion of the text. 

the button now says 

Link user data from <i class="fab fa-github" aria-hidden="true"></i> GitHub

Re-sync user data from <i class="fab fa-github" aria-hidden="true"></i> GitHub

packages/openneuro-server/src/libs/authentication/github.ts
thinknoack: I removed the provider to make it more flexible.

packages/openneuro-server/src/libs/authentication/github.ts
thinknoack: add new error route user/github that goes to 403

packages/openneuro-server/src/libs/authentication/github.ts
thinknoack: add new error route user/github that goes to 403

packages/openneuro-app/src/scripts/users/github-auth-button.tsx
thinknoack: added

packages/openneuro-server/src/routes.ts
```diff
@@ -157,6 +164,58 @@ const routes = [
...
+              encodeURIComponent(info?.message || "github_auth_failed")
+            }`,
+          )
+        }
+
+        return res.redirect(`${cleanRedirectTo}?success=github_auth_success`)
+      })(req, res, next)
+    },
+  },
+
```
nellh: The implementation here looks good, but I think these handlers should be imported from their own module, like the Google and ORCID ones.

nellh (changes requested):

packages/openneuro-server/src/routes.ts
thinknoack: ok, those are moved to libs/authentication/github.ts

nellh (approved):

Issue #3316, opened 2025-01-31
Unrestrict user deletion before DOI creation


effigies:
### What would you like to see added?

Maybe add a notice at snapshot creation that deletion will become harder?

### Do you have any interest in helping implement the feature?

Yes

nellh:
This was implemented in #3334.

PR #3334, opened 2025-02-13
feat(app): Allow users to delete datasets if no versions have been created

nellh:
Updates messaging around version creation to indicate that creating the first version will disable deletion.

See #3316

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/routes/snapshot.tsx
+++ b/packages/openneuro-app/src/scripts/dataset/routes/snapshot.tsx
@@ -74,8 +74,8 @@ const SnapshotRoute = ({
           )}
           <p>
             Create a new version of this dataset for download and public access.
-            This will begin an export of this dataset to GitHub and S3 if it has
-            been made public.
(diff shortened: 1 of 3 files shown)


packages/openneuro-app/src/scripts/dataset/routes/snapshot.tsx
```diff
@@ -128,6 +128,14 @@ const SnapshotRoute = ({
               You must add at least one change message to create a new version
             </small>
           )}
+          {snapshots.length === 0
+            ? (
+              <p>
+                Once a version has been created and DOI assigned, the dataset
+                can no longer be deleted.
+              </p>
```
effigies, line 136: Does it make sense to make this a checkbox, e.g.:

* [ ] I acknowledge that by creating a snapshot and assigning a DOI, I will no longer be able to delete this dataset.

nellh: Added this and used your language for it.

PR #3350, opened 2025-02-24
update to persistent filters when switching between modalities

thinknoack:
contains updates to search filters

currently: 

- any time you select a modality it removes the query. 
- selecting BRAIN Initiative while on a modality will reset the query and take you to the brain initiative portal. 

update:

- when switching between modalities and even removal of modality the other params are saved
- when on a modality (kind of like another search portal) selecting brain inititative will not redirect you to the brain initiative portal but instead add a filter
- when on the BI portal page - selecting a modality will take you to the modality "search" with that filter present
- there is now an Initiative Filter 'chicket'/tag that you can X out of - if on the modality search you will stay within the modality, if on the brain initiative landing page you will be taken back to /search and the other params should all remain persistent 
- also fixes -removing the "raw" filter resets the query


packages/openneuro-app/src/scripts/search/filters-block-container.tsx
```diff
@@ -25,12 +25,26 @@ const FiltersBlockContainer: FC<FiltersBlockContainerProps> = ({
...
+  const pattern = "/search/modality/*"
+
+  const convertWildcardToRegex = (pattern) => {
+    const escapedPattern = pattern.replace(/[|\\{}()[\]^$+?.]/g, "\\$&")
+    const regexPattern = "^" + escapedPattern.replace(/\*/g, ".*") + "$"
+    return new RegExp(regexPattern)
+  }
+
+  const regex = convertWildcardToRegex(pattern)
+  const isModalityPath = regex.test(location.pathname)
```
nellh: I think useMatch() would be simpler, it will identify modality paths without the regular expressions and in an injection safe way.

packages/openneuro-app/src/scripts/search/filters-block-container.tsx
```diff
@@ -46,6 +60,19 @@ const FiltersBlockContainer: FC<FiltersBlockContainerProps> = ({
             Sentry.captureException(error)
           }
         }
+      } else if (!isModalityPath && param === "brain_initiative") {
+        removeFilterItem(setSearchParams)(param, value)
+        const queryParams = new URLSearchParams(location.search)
+        const query = queryParams.get("query")
```
nellh: Similarly the hook useSearchParams() can return the query parameter so constructing a new URLSearchParams isn't needed.

packages/openneuro-app/src/scripts/search/search-params-ctx.tsx
```diff
@@ -20,6 +20,10 @@ export const SearchParamsProvider: React.FC<SearchParamsProviderProps> = ({
     const query = searchQuery.get("query")
     if (query) {
       searchParams = JSON.parse(query)
+    } else {
+      Sentry.captureException(
+        "No query found in URL. Using initialSearchParams.",
+      )
```
nellh: This is probably common enough to not qualify as an exception. A breadcrumb tracking it might be useful though.
```suggestion
      Sentry.addBreadcrumb({
        category: "routing",
        message: "No query found in URL. Using initialSearchParams.",
        level: "info"
      })
```

nellh (changes requested):

packages/openneuro-app/src/scripts/search/filters-block-container.tsx
thinknoack: ah very nice

packages/openneuro-app/src/scripts/search/filters-block-container.tsx
thinknoack: that is helpful

thinknoack: updated

packages/openneuro-app/src/scripts/search/filters-block-container.tsx
thinknoack: had to redo a little after adding useSearchParams, but this update seems to be working as I expect it without constructing a new url params

nellh (approved):

PR #3354, opened 2025-02-25
feat: Progessive loading for legacy validation

nellh:
This speeds up a lot of page loads where the main bottleneck is fetching the validator issues. Instead we load a status which is just a count of errors and warnings, loading the real issues data only after the user expands the issues header. This removes the need to load all issues during search indexing as well, speeding up search reindexing considerably.

Refines the types for the Issues model and migrates part of the old issues components to TypeScript to simplify the refactor.


thinknoack (commented):
reviewed code and changes look good locally

PR #3355, opened 2025-02-26
search filters update for Invalid DS on My Datasets - use latestSnapshot.validation.errors over draft.issues.severity

thinknoack:
- update to the dataset mapping/query to include validation, 
- changes to search resolver to check if validation.errors is greater than zero, 
- removes the status filter if you are not viewing my datasets on search

### DIFF
--- a/packages/openneuro-search/src/query.ts
+++ b/packages/openneuro-search/src/query.ts
@@ -49,6 +49,9 @@ export const INDEX_DATASET_FRAGMENT = gql`
         ReferencesAndLinks
         DatasetDOI
       }
+      validation{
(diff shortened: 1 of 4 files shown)


packages/openneuro-search/src/query.ts
```diff
@@ -49,6 +49,9 @@ export const INDEX_DATASET_FRAGMENT = gql`
         ReferencesAndLinks
         DatasetDOI
       }
+      validation{
```
nellh:
```suggestion
      validation {
```

nellh (approved):

Issue #3353, opened 2025-02-25
Add the loading icon when "Load More" is clicked on the search page.


thinknoack:
### What went wrong?

currently no indication that something is happening when you click load more on search

### Expected behavior

clicking load more should add the loading animation letting users know a process is running. 

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

effigies:
I think this is done?

thinknoack:
> I think this is done?

Yes, this is complete. It did introduce the regression you noted in [#3449](https://github.com/...), which has since been fixed and merged. The fix is currently awaiting deployment.

PR #3357, opened 2025-02-27
adding a loading animation to page when load more is clicked

thinknoack:
closes issue #3353 

not sure if I should/could tap into loading coming from the useSearchResults hook, so I added a useState to manage the fetchMore loading state


nellh (changes requested):
I think this can be simplified to just `notifyOnNetworkStatusChange: true` on the query run. This should notify the component of the loading state while fetchMore is running.

thinknoack:
> I think this can be simplified to just `notifyOnNetworkStatusChange: true` on the query run. This should notify the component of the loading state while fetchMore is running.

Ok I have modified the loadmore to work with notifyOnNetworkStatusChange

packages/openneuro-app/src/scripts/search/search-container.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect } from "react"
+import React, { useContext, useEffect, useState } from "react"
```
nellh:
```suggestion
import React, { useContext, useEffect } from "react"
```

nellh (approved):

PR #3359, opened 2025-02-27
Make version in the footer into hyperlink to the github release

yarikoptic:
This would help users to see how recent release is and associated with it changes etc


effigies (approved):
LGTM.

PR #3368, opened 2025-03-05
fix(app): Match modality selections on value instead of labels

nellh:
Fix a bug where the modality search selection wasn't being set correctly when choosing filters on the front page.


thinknoack (approved):
filter on home page is working as expected.

PR #3369, opened 2025-03-05
fix(app): Show formatted modality string in filters

nellh:
Use the same ModalityLabel component for showing formatted values in search filters.


thinknoack (approved):
looks good

PR #3378, opened 2025-03-10
update to the /admin view on dataset pages to include starter component for dataset-events

thinknoack:
adds component to display dataset event log


packages/openneuro-app/src/scripts/dataset/mutations/dataset-events.tsx
```diff
@@ -0,0 +1,227 @@
...
+                      <button
+                        onClick={handleUpdateNote}
+                        className="on-button on-button--small on-button--primary"
+                      >
+                        Save
+                      </button>
+                    )}
+                  </div>
+                </div>
+              </li>
```
nellh: I'd break this out into its own component.

packages/openneuro-app/src/scripts/dataset/routes/admin-datalad.jsx
```diff
@@ -30,6 +31,8 @@ const AdminDataset = ({ dataset }) => (
     <HeaderRow4>Draft Head</HeaderRow4>
     {dataset.draft.head}
     <DatasetHistory datasetId={dataset.id} />
+    <hr />
+    <DatasetEvents datasetId={dataset.id} />
```
nellh, line 34: We might want to put this above the history? I think this will be more common to reference.

nellh (approved):

packages/openneuro-app/src/scripts/dataset/mutations/dataset-events.tsx
thinknoack: new component added.

packages/openneuro-app/src/scripts/dataset/routes/admin-datalad.jsx
thinknoack: moved above history

PR #3395, opened 2025-03-24
adding polling context

thinknoack:
adds startPolling and stopPolling apollo functions to context, so various components can update when validation changes.

### DIFF
--- a/packages/openneuro-app/src/scripts/dataset/components/ValidationBlock.tsx
+++ b/packages/openneuro-app/src/scripts/dataset/components/ValidationBlock.tsx
@@ -60,7 +60,7 @@ export const ValidationBlock: React.FC<ValidationBlockProps> = ({
     if (!issuesStatus && !validation) {
       if (typeof startPolling === "function") {
         try {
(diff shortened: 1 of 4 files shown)


packages/openneuro-app/src/scripts/dataset/components/ValidationBlock.tsx
```diff
@@ -38,6 +40,40 @@ export const ValidationBlock: React.FC<ValidationBlockProps> = ({
...
+      }
+    }
+  }
+
+  // Function to start polling if ValidationPending is shown.
+  const startPollingValidation = () => {
+    if (!issuesStatus && !validation) {
+      if (typeof startPolling === "function") {
+        try {
+          startPolling(500) // 500ms poll interval
```
nellh:
```suggestion
          startPolling(10000) // 10s poll interval
```

This looks good but I'd raise the polling time to avoid flooding the server with API requests if validation is taking a long time.

nellh (approved):

PR #3411, opened 2025-04-09
Drop @openneuro/components package

nellh:
This simplifies our TypeScript build significantly as we no longer depend on Lerna for workspace linking in the app after this change.

I've replaced the use of index packages and master imports for scss here which also reduces some of the extra data loaded for a given page by shaking the unused objects.


thinknoack (approved):
yay! 

files reviewed and site tested locally.

PR #3413, opened 2025-04-09
fix(app): Prevent extra scrolling on content boxes

nellh:
I noticed these pages scrolled despite having less than one page of content.


thinknoack (approved):
reviewed and approved.

PR #3426, opened 2025-04-16
Prevent git index from being updated twice during commits

### LINKED ISSUE #2848: Large objects can fail to be annexed
### What went wrong?

Noticed this on ds004598, this dataset was recently uploaded and several AVI files should have been annexed but were not. Need to track down how this happened as the previous causes of it have been fixed.

### Expected behavior

Except for the files excluded in .gitattributes, everything should be annexed.

### How to reproduce

_No response_

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

### Additional information

_No response_

nellh:
`git-annex add` called here updates the git index on disk. We would throw away this work and immediately create a pygit2 index in memory. This was usually fine but slightly inefficient, however it can lead to a race condition where two attempts at creating a commit would end up with a different on disk and in memory index. By limiting this to only git-annex add and reading the on disk index, the index is always consistent and avoids the risk of adding large files as regular git objects accidentally.

Adding some additional error handling here that should help with debugging this kind of failure in the future.

Fixes #2848

### DIFF
--- a/services/datalad/datalad_service/common/git.py
+++ b/services/datalad/datalad_service/common/git.py
@@ -1,6 +1,8 @@
 import pathlib
 import re
 import subprocess
+import logging
+import sentry_sdk
 
 import pygit2
 from charset_normalizer import from_bytes
@@ -10,6 +12,7 @@ tag_ref = re.compile('^refs/tags/')
 COMMITTER_NAME = 'Git Worker'
 COMMITTER_EMAIL = 'git@openneuro.org'
 
+logger = logging.getLogger(__name__)
 
 class OpenNeuroGitError(Exception):
     """OpenNeuro git repo states that should not arise under normal use but may be a valid git operation in other contexts."""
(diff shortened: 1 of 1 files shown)


services/datalad/datalad_service/common/git.py
```diff
@@ -78,9 +81,22 @@ def git_commit(repo, file_paths, author=None, message="[OpenNeuro] Recorded chan
             'HEAD points at invalid branch name ({}) and commit was aborted'.format(repo.references['HEAD'].target))
     # Refresh index with git-annex specific handling
     annex_command = ["git-annex", "add"] + file_paths
-    subprocess.run(annex_command, check=True, cwd=repo.workdir)
-    repo.index.add_all(file_paths)
-    repo.index.write()
+    try:
+        subprocess.run(annex_command, check=True, cwd=repo.workdir)
```
effigies: I think you need this to access stdout/stderr below. I'm not looking at the docs right now, though.

```suggestion
        subprocess.run(annex_command, check=True, capture_output=True, cwd=repo.workdir)
```

effigies (commented):
Quick thing I noticed. Will look a bit more closely tomorrow.

effigies (approved):
LGTM.

PR #3427, opened 2025-04-16
[WIP] Coral embed

thinknoack:


config.env.example
```diff
@@ -4,6 +4,9 @@ CRN_SERVER_URL=http://localhost:9876
 # Secret authentication value for this instance
 JWT_SECRET=
 
+# Coral Talk Secret from SSO Config in Admin Panel
+CORAL_JWT_SECRET=ssosec_0178013126e99c4f68e06d47794a96c01c5a117b01e00687f24c310a2b49d9f993
```
nellh: This should be blank, it shouldn't be possible to accidentally start without setting a new random value.

packages/openneuro-app/src/scripts/dataset/comments/coral-embed.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useEffect, useRef, useState } from "react"
+
+declare global {
+  interface Window {
+    Coral?: {
+      createStreamEmbed: (config: any) => void
+    }
+  }
+}
```
nellh: Doing this is fine but we should probably move this to a global definitions typescript file near the entrypoint for the project, it can be surprising elsewhere.

packages/openneuro-app/src/scripts/dataset/comments/coral-embed.tsx
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+
+    const script = document.createElement("script")
+    script.src = "http://localhost:5001/assets/js/embed.js"
+    script.async = true
+    script.defer = true
+    script.onload = fetchAndInitializeCoral
+    script.onerror = () => console.error("Failed to load Coral embed.js")
+    document.head.appendChild(script)
```
nellh: This should be a script tag in the index. I was able to get it working like that, so I think you can make this change.

packages/openneuro-app/src/scripts/utils/global-polyfill.ts
```diff
@@ -1,16 +1,16 @@
...
+// function polyfillGlobal(): void {
+//   if (typeof global === "object") return
+//   Object.defineProperty(Object.prototype, "global", {
+//     get: () => globalThis,
+//     configurable: true,
+//   })
+// }
 
-polyfillGlobal()
+// polyfillGlobal()
```
nellh: Glad we found this. I'd just remove this file from the source if this PR is cleaning up all the draft.js components anyways.

packages/openneuro-server/src/handlers/users.ts
```diff
@@ -13,3 +16,33 @@ export function createAPIKey(req, res, next) {
...
+    const header = {
+      alg: "HS256",
+    }
+    const token = jwt.sign(payload, config.auth.coral.secret, { header })
+    res.json({ token })
+  } catch (error) {
+    Sentry.captureException(error)
+    res.status(500).json({ message: "Failed to generate Coral SSO token" })
+  }
+}
```
nellh, line 45: This looks pretty reasonable.

packages/openneuro-app/src/scripts/dataset/comments/coral-embed.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const isAuthenticated = document.cookie.includes("accessToken=")
+
+  useEffect(() => {
+    const fetchAndInitializeCoral = async () => {
+      const accessTokenFromCookie = document.cookie.split("; ").find((cookie) =>
+        cookie.startsWith("accessToken=")
+      )?.split("=")[1]
+      const headers = accessTokenFromCookie
+        ? { Authorization: `Bearer ${accessTokenFromCookie}` }
+        : {}
```
nellh, line 21: This shouldn't be necessary. The cookie is already set when the fetch request is made.

config.env.example
thinknoack: oof, cut and paste

packages/openneuro-app/src/scripts/dataset/comments/coral-embed.tsx
thinknoack: I think I have made the right adjustments.

packages/openneuro-app/src/index.html
```diff
@@ -18,6 +18,11 @@
       href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,700;1,300;1,400;1,700&display=swap"
       rel="stylesheet"
     />
+    <script
+    src="http://localhost:5001/assets/js/embed.js"
+    async
```
nellh: This shouldn't be async and should use a relative URL so it works for other domains. While it will usually work with async, it's possible for it to be parsed after the React app renders and fail intermittently.

thinknoack:
@nellh - Would you mind reviewing this again? I have made the above adjustments, I also added the admin role for Openneuro admins, connected the embed signin to the modal, and added some basic openneuro styling. 

currently going to fix the lint issues and write some test coverage, as well as wait for the coral teams response to my email

packages/openneuro-app/src/scripts/dataset/comments/__tests__/coral-embed.spec.jsx
```diff
@@ -0,0 +1,44 @@
...
+import { UserModalOpenCtx } from "../../../utils/user-login-modal-ctx"
+import { vi } from "vitest"
+
+vi.mock("../../../utils/user-login-modal-ctx", () => ({
+  UserModalOpenCtx: React.createContext({ setUserModalOpen: vi.fn() }),
+}))
+
+global.fetch = vi.fn().mockResolvedValue(Promise.resolve({
+  json: vi.fn().mockResolvedValue({ token: "mock-sso-token" }),
+}))
```
nellh, line 13: This should probably use the fetch mocking library. `vitest-fetch-mock`

nellh (changes requested):
Looks pretty good, a few minor comments.

packages/openneuro-app/src/scripts/dataset/comments/__tests__/coral-embed.spec.jsx
thinknoack: I am having an issue with vitest-fetch-mock. Is this some sort of version related vite issue?

just adding `import createFetchMock from 'vitest-fetch-mock'; `to my test file I get 

` FAIL  packages/openneuro-app/src/scripts/dataset/comments/__tests__/coral-embed.spec.jsx > CoralEmbed - Simple Embed.js Check (Vitest) > should check if window.Coral exists and initializeCoralEmbed is called
TypeError: Cannot set property testPath of #<Object> which has only a getter`

packages/openneuro-app/src/index.html
thinknoack: Can you explain the relative URL piece. Should I add embed.js to the openneuro-app? Sorry I am not following 100%.

nellh: I think the best way to approach it is to map localhost:5001 to a path in the load balancer. For local dev that's in the nginx config (before the final / location):

```
    location /coral {
        client_max_body_size 0;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header Connection "";
        proxy_http_version 1.1;
        proxy_request_buffering off;
        proxy_pass http://coral:5001;
    }
```

The URL then looks like this:

```html
<script src="/coral/assets/js/embed.js" />
```

thinknoack: with this setup and a restart of the container I am getting a 404 (`http://localhost:9876/coral/assets/js/embed.js`) and the file is still accessible from `http://localhost:5001/assets/js/embed.js`

thinknoack: I pushed the changes to nginx and index for debug/review

thinknoack: Moving this to the dev.conf for nginx works

I had to change the proxy_pass to http://coral:5000;

Issue #3342, opened 2025-02-18
Hide uploader in anonymous review mode


effigies:
### What would you like to see added?

Although uploaders can decline to add authorship information to datasets during a review period, they cannot hide the upload account, which is very likely to be identifying. Because this is not present in the dataset itself, this should be easy to redact.

### Do you have any interest in helping implement the feature?

Yes

effigies:
Had a quick look, and I don't see a way to access whether the current user is a reviewer in https://github.com/OpenNeuroOrg/openneuro/blob/master/packages/openneuro-app/src/scripts/dataset/snapshot-container.tsx. @nellh If you can see a quick way to access that infromation from there, I could probably do this, but it looks like it might be a somewhat complicated refactor.

nellh:
This was released.

PR #3431, opened 2025-04-23
check profile for anonymous reviewer and hide uploaded by field

thinknoack:
simple profile check to hide possible PII for anonymous reviewers. 

relates to issue #3342


nellh (changes requested):
A better way to check for this is to see if the token has scopes defined and one value will be `dataset:reviewer` if the token is scoped for review only.

nellh (approved):

PR #3437, opened 2025-04-30
Migrate Google accounts to ORCID

### LINKED ISSUE #180: Account consolidation - ORCID and Google
I logged in using ORCID and even though my gmail account is my primary contact email on ORCID I was not able to see the datasets I uploaded using my gmail account before. It seems the accounts are not consolidated.

Was this expected @anibalsolon?

nellh:
This will require ORCID for all logins on OpenNeuro. If a user has a Google account without a linked ORCID, they will be asked to link one.

Google only accounts will continue to function but a new website login will require linking ORCID to continue using the website. For deployment, the plan is to invalidate OpenNeuro's existing API keys.

The migration here is non-destructive, creating a log of the updated records and rolling back the transaction if the migration fails. A successful migration changes ownership of the content from the Google account to the new or existing ORCID account and marks the Google account as a "migrated" account.

After an account is migrated, it can still be used for login but in some cases will require the client to complete both the Google and ORCID oauth flows, so users should be encouraged to use their ORCID for logins.

Fixes #180.

### DIFF
--- a/config.env.example
+++ b/config.env.example
@@ -62,7 +62,7 @@ FRESH_DESK_URL=
 ENGINE_API_KEY=
 
 # MongoDB URL
-MONGO_URL=mongodb://mongo:27017/crn
+MONGO_URL=mongodb://mongo:27017/crn?replicaSet=rs0
 
 # DataLad service container host and port
 DATALAD_SERVICE_URI=datalad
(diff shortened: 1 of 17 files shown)


nellh:
@thinknoack You'll need to add `?replicaSet=rs0` to your MongoDB connection to test but the transaction error you had should be fixed by just restarting your containers with these changes.

thinknoack (approved):
It looks good. After some back and forth I have tested a few user situations. 

- New orcid, 
- New google
- existing google
- existing orcid

PR #3442, opened 2025-05-06
Update - User page datasets

thinknoack:
updates user page "Datasets/My Datasets" to filter datasets to a specific orcid user. 
extends filters/search/sort to use elasticsearch
adds a loadmore for users that have a lot of datasets


packages/openneuro-server/src/graphql/resolvers/user.ts
```diff
@@ -9,7 +9,7 @@ function isValidOrcid(orcid: string): boolean {
 export const user = (obj, { id }) => {
   if (isValidOrcid(id)) {
     return User.findOne({
-      $or: [{ "orcid": id }, { "providerId": id }],
+      $or: [{ "provider": "orcid", "providerId": id }],
```
nellh, line 12: @thinknoack I changed this because I was seeing a different set of datasets for different users based on what step in user migration was complete. I think it makes sense to only match this on ORCID users going forward now that the migration is live.

packages/openneuro-app/src/scripts/users/__tests__/datasest-card.spec.tsx
nellh, line 1: Filename here is misspelled.

packages/openneuro-app/src/scripts/users/__tests__/datasest-card.spec.tsx
```diff
@@ -183,19 +24,115 @@ const mockDataset = {
...
     expect(screen.getByText("Test Dataset")).toBeInTheDocument()
     expect(screen.getByText("ds000001")).toBeInTheDocument()
-    expect(screen.getByText("1.00 GB")).toBeInTheDocument()
+
+    expect(screen.getByText((content, element) => {
+      const isDateSpan = element?.tagName.toLowerCase() === "span" &&
+        element.textContent?.includes("Added:")
+      if (!isDateSpan) return false
+      const expectedDatePart = "12/31/2024"
+      const expectedAgoPart = "4 months ago"
```
nellh: Age is misspelled and this test is dependent on the current date, so this test will fail periodically. Skip matching on this string in the test is easiest, I don't think this is a very useful thing to check if the date itself is rendered.

nellh (changes requested):

nellh (approved):

Issue #1082, opened 2019-04-18
Admin users list does not show a loading state


nellh:
This is most apparent on production where loading the users list takes a few seconds. We should add a loading indicator since the "no results" message is confusing.

nellh:
This should get fixed, the admin management page is unusable at the moment.

effigies:
There is a loading indicator now, but the admin page is still unusable. Not sure whether to close this and open a new issue, or just bump this.

The most responsive way to use it at the moment is to toggle "admin", put in the search query, and untoggle.

@thinknoack Is this something we could add to your queue? Lowish priority.

https://openneuro.org/admin/users

nellh:
For @thinknoack - the main issue is the search is implemented client side. This should be paginated and the query needs to take the search argument to load only matching users.

effigies:
If we're going to keep live-updating search results, I think the search box needs to be debounced (though the apparent failure may just be the lags being larger than the debounce interval).

PR #3458, opened 2025-05-15
Update/admin user view

thinknoack:
addresses #1082 
contains updates to query users 

per discussion w/Nell

- adds filter and sorting to users query
   - the query takes the search argument to load only matching users
- adds pagination 
- The API itself should not return any users where migrated: true is set. 
- shows users from both google and orcid lists
- updates various styling elements and layout in the client.


packages/openneuro-server/src/graphql/resolvers/user.ts
```diff
@@ -17,13 +18,105 @@ export const user = (obj, { id }) => {
   }
 }
 
-export const users = (obj, args, { userInfo }) => {
-  if (userInfo.admin) {
```
nellh, line 21: Looks like this check is missing, which is here to prevent this query from being used by non-admin users.

nellh (changes requested):

packages/openneuro-server/src/graphql/resolvers/user.ts
thinknoack: ok, this is updated

nellh (approved):

PR #3461, opened 2025-05-20
various links to user pages updated. styling update for github link b…

thinknoack:
hides notifications tab/link on userpages
updates user menu to allow linking to account and datasets on userpages
various styling updates

### DIFF
--- a/packages/openneuro-app/src/scripts/users/user-account-view.tsx
+++ b/packages/openneuro-app/src/scripts/users/user-account-view.tsx
@@ -81,20 +81,20 @@ export const UserAccountView: React.FC<UserAccountViewProps> = ({
     }
   }
 
-  // --- NEW REGEX FOR URL VALIDATION ---
-  // This regex requires the URL to start with "http://" or "https://".
-  // It checks for:
-  // ^               - Start of the string
-  // (http|https)    - Matches "http" or "https"
-  // :\\/\\/         - Matches "://"
-  // [a-zA-Z0-9.-]+  - Matches the domain name (e.g., "example.com")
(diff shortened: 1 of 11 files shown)


packages/openneuro-app/src/scripts/users/user-account-view.tsx
```diff
@@ -114,9 +129,8 @@ export const UserAccountView: React.FC<UserAccountViewProps> = ({
         setRows={handleLinksChange}
         className="custom-class"
         heading="Links"
-        // eslint-disable-next-line no-useless-escape
-        validation={/((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=\+\$,\w]+@)?[A-Za-z0-9.-]+|(?:www.|[-;:&=\+\$,\w]+@)[A-Za-z0-9.-]+)((?:\/[\+~%\/.\w-_]*)?\??(?:[-\+=&;%@.\w_]*)#?(?:[\w]*))?)/} // URL validation regex
-        validationMessage="Invalid URL format. Please use a valid link."
+        validation={httpHttpsRequiredUrlValidation}
```
nellh: Might be better to just check if `new URL()` fails for a string to avoid a regexp for it.

nellh (changes requested):

packages/openneuro-app/src/scripts/users/user-account-view.tsx
thinknoack: cool, that is way better. I have it added.

nellh (approved):

PR #3471, opened 2025-05-27
fix: Enforce eol=lf for .json, .tsv, CHANGES, README and LICENSE

### LINKED ISSUE #3465: CHANGES files with carriage returns (`\r`) break snapshot creation
### What went wrong?

ds005555 manually fixed up their changelog, and uploaded a file with `\r\n` newlines. The snapshots failed with `Bad Request`. After running `dos2unix`, the snapshot proceeded.

### Expected behavior

We should auto-convert CRLF to LF. I believe we can do this by setting `* text=auto` in `.gitattributes`, which will force LF in files git detects as text.

It's possible that git could misdetect files and corrupt data, so we could limit it to known files, such as:

```gitattributes
CHANGES eol=lf
README* eol=lf
LICENSE* eol=lf
*.json eol=lf
*.tsv eol=lf
```

@yarikoptic Do you have any experience with using `* text=auto` causing problems in BIDS datasets?

### How to reproduce

_No response_

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

### Additional information

_No response_

effigies:
This PR marks `.json`, `.tsv`, `CHANGES`, `README[.*]`, `LICENSE[.*]` and `CITATION.cff` as text and instructs git to normalize to LF (as oppposed to CRLF) line endings.

Closes #3465.

### DIFF
--- a/services/datalad/datalad_service/tasks/dataset.py
+++ b/services/datalad/datalad_service/tasks/dataset.py
@@ -18,14 +18,14 @@ GIT_ATTRIBUTES = """* annex.backend=SHA256E
 **/.git* annex.largefiles=nothing
 *.bval annex.largefiles=nothing
 *.bvec annex.largefiles=nothing
-*.json annex.largefiles=largerthan=1mb
+*.json text eol=lf annex.largefiles=largerthan=1mb
+*.tsv text eol=lf annex.largefiles=largerthan=1mb
 phenotype/*.tsv annex.largefiles=anything
-*.tsv annex.largefiles=largerthan=1mb
 dataset_description.json annex.largefiles=nothing
 .bidsignore annex.largefiles=nothing
-CHANGES annex.largefiles=nothing
-README* annex.largefiles=nothing
-LICENSE annex.largefiles=nothing
+CHANGES text eol=lf annex.largefiles=nothing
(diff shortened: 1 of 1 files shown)


services/datalad/datalad_service/tasks/dataset.py
```diff
@@ -18,14 +18,15 @@
 **/.git* annex.largefiles=nothing
 *.bval annex.largefiles=nothing
 *.bvec annex.largefiles=nothing
-*.json annex.largefiles=largerthan=1mb
+*.json text eol=lf annex.largefiles=largerthan=1mb
+*.tsv text eol=lf annex.largefiles=largerthan=1mb
 phenotype/*.tsv annex.largefiles=anything
```
effigies, line 23: I did a little playing, and the final match wins. So we want `phenotype/*.tsv` to override `*.tsv`.

You can check on ds005237:

```
󰅂 git check-attr -a phenotype/abc.tsv
phenotype/abc.tsv: annex.backend: SHA256E
phenotype/abc.tsv: annex.largefiles: largerthan=1mb
phenotype/abc.tsv: filter: annex
```

nellh (approved):

PR #3473, opened 2025-05-29
Use annexed content for edits to dataset_description.json and CHANGES

### LINKED ISSUE #3464: CHANGES and dataset_description.json must not be annexed to create a snapshot
### What went wrong?

A few examples of this have come up, as users use the git interface to upload more data, we probably need to be more flexible about this.

### Expected behavior

OpenNeuro should handle edits to these files even if the files are annexed.

### How to reproduce

_No response_

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

### Additional information

_No response_

nellh:
This will use and maintain the annexed state for CHANGES and dataset_description.json when creating snapshots.

Fixes #3464

### DIFF
--- a/services/datalad/datalad_service/common/git.py
+++ b/services/datalad/datalad_service/common/git.py
@@ -1,9 +1,13 @@
+import asyncio
+import io
+import os
 import pathlib
 import re
 import subprocess
 import logging
 import sentry_sdk
 
+import aiofiles
 import pygit2
 from charset_normalizer import from_bytes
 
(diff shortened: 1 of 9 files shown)


services/datalad/datalad_service/tasks/description.py
```diff
@@ -11,26 +12,33 @@ def edit_description(description, new_fields):
...
                                 description, description_file_contents)
-            description_file.seek(0)
-            description_file.truncate(0)
-            description_file.write(json.dumps(
+        # If contents are sensible, reopen and overwrite (at the symlink target for annexed files)
+        async with aiofiles.open(path, "a+", encoding='utf-8', newline='') as description_file:
+            await description_file.seek(0)
+            await description_file.truncate(0)
+            await description_file.write(json.dumps(
                 updated, indent=4, ensure_ascii=False))
```
effigies: I think this will corrupt the file in the annex tree? I'd say it's probably better to unlink and then re-create.

```suggestion
        # If contents are sensible, overwrite (removing the file to avoid 
        await aiofiles.os.unlink(path)
        async with aiofiles.open(path, 'w', encoding='utf-8', newline='') as description_file:
            await description_file.write(json.dumps(
                updated, indent=4, ensure_ascii=False))
```

If we're worried about a crash between these two, you might consider https://boltons.readthedocs.io/en/latest/fileutils.html#atomic-file-saving.

nellh: Thanks, I added a test to cover a dataset_description.json edit and consolidated the write logic since CHANGES was properly handling this.

services/datalad/datalad_service/common/git.py
```diff
@@ -26,6 +30,59 @@ def git_show(repo, committish, obj):
...
+async def read_from_bytes_as_stream(data: bytes, chunk_size: int = 1024):
+    """
+    Reads from a bytes object as an async stream, yielding chunks.
+    """
+    buffer = io.BytesIO(data)
+    while True:
+        chunk = await asyncio.to_thread(buffer.read, chunk_size)
+        if not chunk:
+            break
+        yield chunk
```
effigies: I might be missing something here, but I suspect you're going to spend more time with `asyncio.to_thread()` than you will with `data.__getitem__`:

```suggestion
    buffer = io.BytesIO(data)
    for i in range(0, len(data), chunk_size):
        yield data[i:i + chunk_size]
```

Edit: I guess this won't be necessary if you switch to blobio...

services/datalad/datalad_service/common/git.py
```diff
@@ -26,6 +30,59 @@ def git_show(repo, committish, obj):
...
+async def git_show_content(repo, committish, filename):
+    """
+    Similar to git_show but resolves annexed content if possible.
+    
+    Returns a stream for the consumer.
+    Annex content is typically much larger and worth streaming.
+    """
+    dataset_root = str(pathlib.Path(repo.path).parent)
+    commit, _ = repo.resolve_refish(committish)
+    data_bytes = (commit.tree / filename).read_raw()
```
effigies: Are you sure you want to `read_raw`? Large files that get accidentally put into `git` could consume a lot of memory. It looks like there's a [BlobIO](https://www.pygit2.org/objects.html#streaming-blob-content) interface that could be used. Something like:

```py
blob = commit.tree / filename
blobio = pygit2.BlobIO(blob):
initial_bytes = blobio.read(4096)
if initial_bytes.find(b'.git/annex') != -1:
    with blobio:
        result = str(from_bytes(initial_bytes + blobio.read()).best())
    ...
    return stream_from_reader(file_obj), os.path.getsize(target_path)
else:
    return stream_from_reader(blobio, start=initial_bytes), blob.size
```

And then use a common:

```python
async def stream_from_reader(reader, start=None, chunk=1024):
    if start is not None:
        yield start
    try:
        while True:
            chunk = await reader.read(chunk_size)
            if not chunk:
                break
            yield chunk
    finally:
        await reader.close()

effigies (approved):
LGTM. IMO, we should make normalizing `.gitattributes` as part of the commit process, so hopefully saving these files as git-annex doesn't last long-term, but making sure we can get the contents out of annex and create new files without corrupting the annex is a solid first step.

PR #3476, opened 2025-06-04
feat(cli): Add support for symlinked content during uploads

### LINKED ISSUE #3421: Uploading datalad dataset does not push annexed files
### What went wrong?

I am working on uploading a datalad dataset to OpenNeuro with the CLI tool. One thing I noticed is that the upload succeeds, but doesn't include annexed files (i.e., NIfTIs). When I `datalad unlock` these annexed files, the upload step catches them.

### Expected behavior

I expected OpenNeuro's CLI tool to work on datalad datasets without any extra steps.

### How to reproduce

1. Create a datalad dataset.
2. Annex some files in the dataset.
3. `deno run -A jsr:@openneuro/cli upload --affirmDefaced path/to/dataset`
4. Check the dataset on OpenNeuro- the annexed files will not be included.

### Desktop

- OS: Linux
- Browser: n/a
- Version: n/a

### Phone

- Device: n/a
- OS: n/a
- Browser: n/a
- Version: n/a

### Additional information

I think documenting this behavior would be good enough for now.

nellh:
This will traverse symlinks during CLI uploads. This is for the current mode of creating a managed annex repo during upload.

Fixes #3421

### DIFF
--- a/cli/src/commands/upload.ts
+++ b/cli/src/commands/upload.ts
@@ -38,7 +38,7 @@ export async function addGitFiles(
   for await (
     const walkEntry of walk(dataset_directory_abs, {
       includeDirs: false,
-      includeSymlinks: false,
+      includeSymlinks: true,
     })
   ) {
     const relativePath = relative(dataset_directory_abs, walkEntry.path)
--- a/cli/src/worker/git.ts
+++ b/cli/src/worker/git.ts
@@ -114,7 +114,16 @@ async function shouldBeAnnexed(
  * git-annex add equivalent
  */
 async function add(event: GitWorkerEventAdd) {
-  const { size } = await context.fs.promises.stat(event.data.path)
+  let size
+  try {
+    const stats = await context.fs.promises.stat(event.data.path)
+    size = stats.size
+  } catch (_err) {
+    console.log(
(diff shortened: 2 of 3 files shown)


cli/src/worker/annex.ts
```diff
@@ -83,7 +83,9 @@ export async function annexAdd(
     ? computeHashMD5
     : computeHashSHA256
   computeHash.init()
-  const stream = context.fs.createReadStream(path, {
+  // Obtain the real path in case this is an annexed or symlinked file
+  const realPath = await context.fs.promises.realpath(path)
+  const stream = context.fs.createReadStream(realPath, {
```
effigies: I'm surprised that we need to worry about this. Opening a symlinked file should be transparent, and nothing in the docs for [createReadStream](https://nodejs.org/api/fs.html#fscreatereadstreampath-options) suggests otherwise. Is the `includeSymlinks: true` above not enough?

nellh: It wasn't actually needed, I pulled that out and added a nicer early exit if any symlinks or files are unreachable.

effigies (approved):
LGTM.

PR #3477, opened 2025-06-04
fix(cli): Don't exempt .gitattributes from dotfile filtering

nellh:
Users have accidentally added bad .gitattributes when pushing without history. It makes more sense to use the OpenNeuro default gitattributes in this case.

If a user intends to add a .gitattributes, they are likely using DataLad or git-annex directly.


effigies (approved):
Agreed. If we're not using their commit history, there's no good reason to use their git config.

PR #3480, opened 2025-06-05
fix(app): Allow public access to user pages

nellh:
Enables the public view of the user profile pages.


thinknoack (approved):
change is good for allowing anonymous users to view user pages

PR #3481, opened 2025-06-05
fix(server): Restrict the User.email field to admins or your own user

nellh:
Require admin access to retrieve the connected email address except for your own user.

This hides this value on the profile page except when viewing your own profile or for site admins.


thinknoack (approved):
updates successfully hide email value unless own/admin

PR #3494, opened 2025-06-12
 moving tools into the header for more space / pre-annotation feature

thinknoack:
Based on newest designs for Annotation feature. This will move the dataset tools towards that approach before we make any updates RE: annotation

ux/ui update to move dataset tools into the dataset header. 

- removes the gradient in the header
- subtle styling updates for various screen sizes
- allows for more room and additional tool tabs


nellh (changes requested):
Looks good for snapshots but the draft container needs to be updated as well.

thinknoack:
> Looks good for snapshots but the draft container needs to be updated as well.

I missed that, draft container is updated now.

nellh (approved):

PR #3495, opened 2025-06-17
Update/search results layouts details component

thinknoack:
Based on newest designs for Annotation feature. This will move the search results tools towards that approach before we make any updates RE: annotation

updates to search layouts

- full width contianer
- search teaser display to resemble pubmed
- new search details panel/drawer component
- adding fields README and Citation
- moves dataset filter/tabs to above results
- moves components from generic components dir to search/components


nellh (changes requested):
I'm still working on the code review. One functional issue:

Screenshot From 2025-06-25 10-17-25

We recently fixed this issue where this dataset cannot be clicked if it has a white space title. To keep the layout, we should probably just replace the title with the accession number if it is only white space.

thinknoack:
> I'm still working on the code review. One functional issue:
> 
> Screenshot From 2025-06-25 10-17-25
> 
> We recently fixed this issue where this dataset cannot be clicked if it has a white space title. To keep the layout, we should probably just replace the title with the accession number if it is only white space.


oh, looks like I wasn't checking for white space.

```
  const datasetId = node.id

  const heading = node.latestSnapshot.description?.Name
    ? node.latestSnapshot.description?.Name
    : datasetId
```
    
how about this - I updated to check for white space dataset names

`  const heading = node.latestSnapshot.description?.Name?.trim() || datasetId`

packages/openneuro-app/src/scripts/search/components/FiltersBlock.tsx
```diff
@@ -69,17 +69,17 @@ export const FiltersBlock = ({
   const subjectCountRangeIsNull =
     JSON.stringify(subjectCountRange) === JSON.stringify([null, null])
 
+  const labelText = modality_selected
```
nellh: There's a modalityShortMapping function that does this in a general way.

https://github.com/OpenNeuroOrg/openneuro/blob/3fa161f2befea57a28a5d26abdc1df863bc2f987/packages/openneuro-app/src/scripts/components/formatting/modality-label.tsx#L1-L22

packages/openneuro-app/src/scripts/search/components/SearchResultDetails.tsx
```diff
@@ -0,0 +1,180 @@
...
+                {item}
+              </span>
+            ))}
+          </div>
+        </div>
+      )
+    } else {
+      return null
+    }
+  }
```
nellh: This sort of thing would probably make sense as a component rather than just a function in the render step. Functions that don't create a closure over any value should generally be defined outside of rendering.

nellh (changes requested):

packages/openneuro-app/src/scripts/search/components/FiltersBlock.tsx
thinknoack: updated this lableText and two other places I was doing the same thing

packages/openneuro-app/src/scripts/search/components/SearchResultDetails.tsx
thinknoack: new component added for list items in the SearchDetails

nellh (approved):

PR #3510, opened 2025-06-27
WIP Adding new resolver creators/datacite.yml

thinknoack:
refactor of how contributor/(author) information is sourced, processed, and displayed across the application by prioritizing datacite.yml,

Notes

- contributor resolver
- two contributor Display Components - links to orcid and ON user datasets if possible
- New specs for contributor component in app/users and server/datalad


packages/openneuro-app/src/scripts/search/use-search-results.tsx
```diff
@@ -263,11 +270,11 @@ export const useSearchResults = () => {
       ]),
     )
   }
-  if (authors.length) {
+  if (authors.length) { // TODO - NELL - this was switched to contributors - is that correct?
```
nellh: Yeah, we want to match contributors once that's available.

packages/openneuro-server/src/datalad/contributors.ts
```diff
@@ -0,0 +1,199 @@
...
+  // 1. Try to get authors from datacite.yml
+  const dataciteCache = new CacheItem(redis, CacheType.dataciteYml, [
+    datasetId,
+    revisionShort,
+  ])
+  try {
+    const dataciteData = await dataciteCache.get(() =>
+      getDataciteYml(datasetId, revision)
+    )
+    if (dataciteData && Array.isArray(dataciteData.authors)) {
```
nellh: Looks like our example here was actually not very representative of a correct implementation.

Digging deeper into 4.6, the authors data is actually part of creators:

https://schema.datacite.org/meta/kernel-4.6/

Here's a better example of structured author metadata.

```yml
creators:
  - creatorName:
      nameType: Personal
      value: Smith, John
    givenName: John
    familyName: Smith
    nameIdentifiers:
      - nameIdentifierScheme: ORCID
        schemeUri: https://orcid.org
        nameIdentifier: https://orcid.org/0000-0002-1825-0097
```

There is also a contributor field that is meant for non-author roles.

packages/openneuro-server/src/datalad/contributors.ts
```diff
@@ -0,0 +1,199 @@
...
+        parsedContributors = normalizeBidsAuthors(
+          datasetDescriptionJson.Authors,
+        )
+        Sentry.captureMessage(
+          `Loaded contributors from dataset_description.json for ${datasetId}:${revision}`,
+        )
+      }
+    } catch (error) {
+      Sentry.captureException(error)
+    }
```
nellh: This is better if it calls the description resolver that exists to avoid reimplementing parts of it.

packages/openneuro-server/src/graphql/schema.ts
```diff
@@ -577,6 +579,8 @@ export const typeDefs = `
     size: BigInt
     # Single list of files to download this snapshot (only available on snapshots)
     downloadFiles: [DatasetFile]
+    # Authors list from datacite.yml || dataset_description.json
+    contributors: [Contributor]
```
nellh: Thinking about this, since datacite.yml makes the distinction between creators and contributors, we might want to call this creator so we can use contributors later (for datacite.yml contributor fields) instead of naming the creators field as contributors in our API schema. That is likely to be confusing if we want to use both fields and it looks like the correct way of adding things like curation contributors is the optional contributors field over adding them as authors (creators in datacite.yml).

packages/openneuro-server/src/datalad/contributors.ts
```diff
@@ -0,0 +1,199 @@
...
+interface RawDataciteAuthor {
+  firstname?: string
+  lastname?: string
+  id?: string
+}
+
+/**
+ * Attempts to read and parse datacite.yml.
+ */
+const getDataciteYml = async (
```
nellh: One more error case to consider is this:

```yaml
resourceType:
  resourceTypeGeneral: JournalArticle
```

Anything other than dataset in this field means that it is probably an unrelated datacite.yml. It's not that unlikely to annotated the software or article like this, so we should check that resourceTypeGeneral is `Dataset`.

nellh (changes requested):
Taking a deeper look at the specification, I found a few issues with how we've approached this. The implementation here looks good but I think there's a few changes required to align with the current datacite metadata schema.

https://datacite-metadata-schema.readthedocs.io/_/downloads/en/4.6/pdf/

PR #3516, opened 2025-07-07
adds modalityHEX and Dataset Citation to the my datasets - dataset cards

### LINKED ISSUE #3511: Add modalities to "My Datasets"
### What would you like to see added?

Datasets should list primary modality 

### Alternatives

_No response_

### Do you have any interest in helping implement the feature?

Yes

### Additional information / screenshots

_No response_

thinknoack:
closes #3511

adds primary modality and cite to dataset cards on the My Dataset user page

### DIFF
--- a/packages/openneuro-app/src/scripts/components/modality-cube/ModalityHexagon.tsx
+++ b/packages/openneuro-app/src/scripts/components/modality-cube/ModalityHexagon.tsx
@@ -4,7 +4,7 @@ import { modalityShortMapping } from "../../components/formatting/modality-label
 import "./modality-hexagon.scss"
 interface ModalityHexagonProps {
   primaryModality: string | null | undefined
-  size?: string | null | undefined
+  size?: string
 }
 
 //ModalityHexagon component displays a colored hexagon and label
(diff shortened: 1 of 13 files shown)


packages/openneuro-app/src/scripts/components/modality-cube/ModalityHexagon.tsx
```diff
@@ -21,7 +23,7 @@ export const ModalityHexagon: FC<ModalityHexagonProps> = ({
     : <i className="fa fa-circle-o-notch fa-spin"></i>
 
   return (
-    <div className="hexagon-wrapper">
+    <div className={`hexagon-wrapper ${size}`}>
```
nellh: If this can be undefined, it should map that to a string `hexagon-wrapper ` or `hexagon-wrapper` instead of `hexagon-wrapper undefined`. Caught by the snapshot test.

packages/openneuro-app/src/scripts/users/dataset-card.tsx
```diff
@@ -87,6 +89,14 @@ export const DatasetCard: React.FC<DatasetCardProps> = (
       datasetSize = `${sizeInBytes} bytes`
     }
   }
+
+  const year = getYear(parseISO(dataset.created))
+  const authors = dataset.latestSnapshot.description?.Authors
+    ? dataset.latestSnapshot.description.Authors.join(" and ")
```
nellh: For display of this, I think we should make two changes:
1. List authors comma separated with only one "and" (preceding the final author's name).
1. If there's many authors, limit the number and replace the list with ... or "et al.".

nellh (changes requested):

nellh (approved):

PR #3521, opened 2025-07-11
Fix: Display loading spinner on initial search load

### LINKED ISSUE #3486: Loading state is not displayed for search after navigation
### What went wrong?

The loading state isn't show but the search results do load after the query finishes.

### Expected behavior

Loading state should be shown when React router navigation changes to the search page routes.

### How to reproduce

_No response_

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

### Additional information

_No response_

thinknoack:
adds initialLoading state to show loading animation by default 

closes #3486

### DIFF
--- a/packages/openneuro-app/src/scripts/search/search-container.tsx
+++ b/packages/openneuro-app/src/scripts/search/search-container.tsx
@@ -120,6 +120,8 @@ const SearchContainer: FC<SearchContainerProps> = ({ portalContent }) => {
   const is_grant_portal = portalContent?.portal || false
   const grant = portalContent?.grant || null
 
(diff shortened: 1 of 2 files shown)


nellh (commented):
The approach of adding an additional loading state for this case didn't feel like the right fix. The cause here is the front page loads the advancedDatasetSearch query and then because the updated parameters on the search page are a refetch, the loading state was suppressed by `cache-first` fetch policy. I think we should use cache and network here anyways.

thinknoack:
thanks, I see now.

thinknoack (commented):
loading on initial page load.

PR #3525, opened 2025-07-21
Add Redis queue using taskiq to worker

nellh:
This adds a simpler asyncio capable queue using taskiq, initially only queueing validation tasks. Future tasks will be scheduled maintenance, QA, and export operations.

Includes helm chart, podman-compose setup (build with `podman-compose  build taskiq`), and fixes for get_docker_scale.py which is used to identify which worker is running.

Queues are per worker with separate Redis databases.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,5 @@
 # docker compose versions
-version: "2.3"
+version: '2.3'
 
 services:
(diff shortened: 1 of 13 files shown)


thinknoack (approved):
validation is loading and I can see the queued datasets running and completing

PR #3528, opened 2025-07-28
Add Contributor Request/Response Workflow and Events UI

thinknoack:
Adds EVENTS for managing contributor access to datasets. Allows users to formally request contributor status, and provides dataset administrators with a dedicated UI to view, accept, or deny these requests. All actions are tracked and displayed in a new, unified events log on the dataset and admins get a log in their account notifications. 

Actual contributor Roles/Perms are TODO - this PR just handles events resolvers and event related UI

Key Work & Changes

- New Contributor Request Workflow
- Request Contributor Button: allows users who do not have permissions to a dataset to formally request contributor access.
- New GraphQL mutations are used to log these requests as events.
- Admin UI: The DatasetEvents component can now managing these requests
- Events & Notifications System
- New Event Types: contributorRequest and contributorResponse
- The DatasetEvents component has been refactored 
- Detailed Event Items component is responsible for displaying each event 
- New CSS added for events UI and actions.

- tests have been adjusted for various issues related to updates. additional tests needed.


packages/openneuro-app/src/scripts/dataset/mutations/request-contributor-status.tsx
```diff
@@ -0,0 +1,89 @@
...
+        id
+        timestamp
+        user {
+          id
+          name
+        }
+        event {
+          type
+          # Include other fields relevant to event descriptions if needed
+          # e.g., version, public, target, level, commit, reference, admin, targetUserId, status, reason
```
nellh:
```suggestion
```
Seems like a generated note?

packages/openneuro-server/src/graphql/schema.ts
```diff
@@ -899,6 +916,8 @@ export const typeDefs = `
     success: Boolean
     # Notes associated with the event
     note: String
+    # top-level datasetId field
+    datasetId: ID
```
nellh, line 993: This seems to duplicate the datasetId from the parent in the model. Why does it need a second copy of it?

packages/openneuro-app/src/scripts/users/user-notification-accordion-actions.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+    try {
+      const backendStatus = newStatus.toUpperCase()
+
+      await updateNotificationStatus({
+        variables: {
+          datasetEventId: id,
+          status: backendStatus,
+        },
+      })
```
nellh: Ran into 400 error, Cannot query field "updateNotificationStatus" on type "Mutation". on triggering this.

packages/openneuro-app/src/scripts/dataset/components/dataset-event-item.tsx
```diff
@@ -0,0 +1,283 @@
...
+          <ToastContent title="Contributor request processed successfully" />,
+        )
+        refetchEvents()
+
+        setTimeout(() => {
+          setUpdatedNote("")
+          if (editingNoteId === event.id) {
+            startEditingNote(null, "")
+          }
+        }, 50)
```
nellh: Is setTimeout necessary here?

packages/openneuro-app/src/scripts/dataset/components/dataset-event-list.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+  event: {
+    type: string
+    requestId?: string
+    status?: string
+  }
+  success: boolean
+  hasBeenRespondedTo?: boolean
+  responseStatus?: string
+}
```
nellh: The event type seems to be defined a few times. It would be good to unify these and import the shared type.

packages/openneuro-app/src/scripts/dataset/mutations/dataset-events.tsx
```diff
@@ -66,8 +32,10 @@ export const DatasetEvents = ({ datasetId }) => {
       setShowForm(false)
       refetch()
     },
-    onError: (error) => {
-      Sentry.captureException(error)
+    onError: (mutationError) => {
+      toast.error(
```
nellh, line 37:
```suggestion
      Sentry.captureException(error)
      toast.error(
```

packages/openneuro-app/src/scripts/dataset/mutations/dataset-events.tsx
```diff
@@ -78,12 +46,50 @@ export const DatasetEvents = ({ datasetId }) => {
...
+            ...event,
+            hasBeenRespondedTo: true,
+            responseStatus: response.event.status,
+          }
+        }
+      }
+      return event
+    })
+    return enrichedEvents
+  }, [rawEvents])
```
nellh: This step could be done on the API side instead. If the client needs this, the API should just return it or provide an option for it.

packages/openneuro-app/src/scripts/dataset/mutations/request-contributor-status.tsx
```diff
@@ -0,0 +1,67 @@
...
+      Sentry.captureException(error)
+    }
+  }
+
+  // Determine if the current user already has any permissions (read, write, admin)
+  // If they do, the button should not be displayed.
+  const hasPermissions = datasetPermissions?.some((p) =>
+    p.user.id === currentUserId &&
+    (p.level === "admin" || p.level === "rw" || p.level === "read")
+  )
```
nellh: We should check if the user is a contributor, rather than if they have permissions.

nellh (changes requested):
It'd be good to try to break things like this up into smaller PRs in the future, this could probably be several component PRs and an API change one. Most of it is looking pretty good and working. Made a few recommendations. It looks intentional that the notification status buttons are failing, that is fine to save for another PR.

packages/openneuro-server/src/graphql/schema.ts
thinknoack: i was thinking that we might have various places outside of dataset that we are querying for events. The new notifications resolver queries for DatasetEvent objects by userID. Outside of a dataset- datasetId is the only way for the frontend to know which dataset an event belongs to. Does that make sense to do here?

packages/openneuro-app/src/scripts/dataset/components/dataset-event-item.tsx
thinknoack: not really. I was trying to account for a slow refetchEvents where there could be what feels like two "renders" in the UI. I don't imagine it to ever be an issue and when I remove it locally it is pretty unnoticeable. I will remove it.

packages/openneuro-app/src/scripts/dataset/components/dataset-event-list.tsx
thinknoack: will do

packages/openneuro-app/src/scripts/dataset/components/dataset-event-item.tsx
thinknoack: removed

packages/openneuro-app/src/scripts/users/user-notification-accordion-actions.tsx
thinknoack: yes, I put that on hold, and have a question about how to store that data on a user if many users will share the same notification.

packages/openneuro-app/src/scripts/dataset/components/dataset-event-list.tsx
thinknoack: done

packages/openneuro-app/src/scripts/dataset/mutations/request-contributor-status.tsx
thinknoack: I added a perm check for the dataset.contributors array. needs to be reviewed once that field is added.

packages/openneuro-app/src/scripts/dataset/mutations/dataset-events.tsx
thinknoack: moved to the backend

packages/openneuro-app/src/scripts/users/user-notification-accordion-actions.tsx
thinknoack: this has been updated in #3548

nellh (approved):

PR #3534, opened 2025-08-11
Add git-annex fsck checks for datasets

nellh:
This adds a task that runs after HEAD is updated or a tag is created to verify local or remote annex fsck state.

There is a 7 day delay before the same objects will be rechecked, any changed objects are always rechecked - the delay just limits how often we recheck an already validated file. This delay may need to be adjusted depending on the resource usage in production.

One included major change required is the move coroutines for commits.


thinknoack (commented):
tests pass and fsck checks run after update

PR #3535, opened 2025-08-11
Add FileCheck model and updateFileCheck mutation

nellh:
This adds a model for storing file checks for datasets. Initially this just records annex fsck output but additional fields can be added for other kinds of file based checks.

One mutation is added updateFileCheck which allows the worker to report annex fsck status on each commit.

For API authentication a token is generated using #3533's worker scope for a specific dataset.

### DIFF
--- a/packages/openneuro-server/src/graphql/schema.ts
+++ b/packages/openneuro-server/src/graphql/schema.ts
@@ -207,10 +207,10 @@ export const typeDefs = `
     createGitEvent(datasetId: ID!, commit: String!, reference: String!): DatasetEvent
     # Create or update a fileCheck document
     updateFileCheck(
-      datasetId: String!
+      datasetId: ID!
       hexsha: String!
       refs: [String!]!
-      annexFsck: [AnnexFsckInput!]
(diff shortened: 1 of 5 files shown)


packages/openneuro-server/src/graphql/schema.ts
```diff
@@ -205,6 +205,13 @@ export const typeDefs = `
...
     # Create a git event log for dataset changes
     createGitEvent(datasetId: ID!, commit: String!, reference: String!): DatasetEvent
+    # Create or update a fileCheck document
+    updateFileCheck(
+      datasetId: ID!
+      hexsha: String!
+      refs: [String!]!
+      annexFsck: [AnnexFsckInput!]!
+    ): FileCheck
   }
```
thinknoack: does this need `remote: String`?

thinknoack (approved):
added one question about schema.

packages/openneuro-server/src/graphql/schema.ts
nellh: Yes, fixed. Thanks!

thinknoack (approved):

PR #3540, opened 2025-08-14
feat(api): Add resolver for creators for datacite.yml 

thinknoack:
Changed BRANCH Name

old PR and comments here
https://github.com/OpenNeuroOrg/openneuro/pull/3510

refactor of how CREATORS/(author) information is sourced, processed, and displayed across the application by prioritizing datacite.yml,

Notes

CREATORS resolver
two contributor Display Components - links to orcid and ON user datasets if possible
New specs for CREATORS component in app/users and server/datalad


thinknoack:
need to rebase

packages/openneuro-app/src/scripts/search/use-search-results.tsx
```diff
@@ -168,7 +174,8 @@ export const useSearchResults = () => {
         "id^20",
         "latestSnapshot.readme",
         "latestSnapshot.description.Name^6",
-        "latestSnapshot.description.Authors^3",
+        "latestSnapshot.description.Authors^3", // TODO: Nell - do we need this still?
```
nellh, line 177: I don't think this is useful after this change, so I'd remove it.

```suggestion
```

packages/openneuro-app/src/scripts/search/use-search-results.tsx
```diff
@@ -263,11 +270,11 @@ export const useSearchResults = () => {
       ]),
     )
   }
-  if (authors.length) {
+  if (authors.length) { // TODO - NELL - this was switched to creators - is that correct?
```
nellh, line 273: This could be updated to match the new terminology, just has to be updated in the parents. I don't think it's strictly necessary though. The idea of authors is still accurate.

packages/openneuro-server/src/utils/orcid-utils.ts
nellh, line 1: Looks like this is copied from the user resolver? I like it in this spot more but would be good to have the user resolver use the same implementation.

nellh (approved):
Couple minor changes I'd recommend but I think this looks good.

PR #3543, opened 2025-08-14
feat(app): Query and display components for contributors 

thinknoack:


packages/openneuro-app/src/scripts/dataset/draft-container.tsx
```diff
@@ -194,7 +205,8 @@ const DraftContainer: React.FC<DraftContainerProps> = ({ dataset }) => {
                 editMode={hasEdit}
               >
                 {description?.Authors?.length ? description.Authors : ["N/A"]}
-              </EditDescriptionList>
+              </EditDescriptionList> */
```
nellh, line 208: I think we do want to be able to edit this still. It should create the datacite.yml entries, maybe this can be done in another PR?

packages/openneuro-app/src/scripts/dataset/__tests__/__snapshots__/snapshot-container.spec.tsx.snap
```diff
@@ -786,14 +786,14 @@ OCI-1131441 (R. Poldrack, PI) in any publications.
...
           </div>
           <div
-            class="dataset-meta-block dmb-inline-list"
+            class="dataset-meta-block "
           >
             <h2
               class="dmb-heading"
             >
-              Authors
+              Contributors
```
nellh: Even if this is contributors in datacite.yml, we might want to keep the authors label here? BIDS calls it authors and datacite.yml is the one exception.

packages/openneuro-app/src/scripts/dataset/draft-container.tsx
thinknoack: yeah. another PR is good.

packages/openneuro-app/src/scripts/dataset/__tests__/__snapshots__/snapshot-container.spec.tsx.snap
thinknoack: updated in both draft/snapshot-container.tsx and snapshot

nellh (approved):

PR #3548, opened 2025-08-19
Feat(api) add a user specific virtual notification status to dataset events

thinknoack:


packages/openneuro-server/src/graphql/resolvers/user.ts
```diff
@@ -211,23 +216,110 @@ export async function notifications(obj, _, { userInfo }) {
...
-    { "event.targetUserId": userId },
+  const pipeline: PipelineStage[] = [
+    {
+      $lookup: {
+        from: "datasets",
+        localField: "datasetId",
+        foreignField: "id",
+        as: "datasetInfo",
+      },
+    },
```
nellh: Is this first lookup needed? It seems like the datasets lookup can be avoided.

thinknoack: looking into this... if I am remembering correctly, I am doing this so the notification list on the user profile can have dataset specific data in the event.

thinknoack: upon review, the schema is not using this and the datasetID is already available to the frontend. I will remove

thinknoack: ok, lookup has been removed

nellh (approved):

PR #3555, opened 2025-08-25
fix(cli): Always merge the git-annex branch on fetch step

nellh:
This merge step is sometimes required to sync git-annex history when the local and remote have diverged. This should always be a clean merge, so this should prevent wiping out the local history or crashing.

### DIFF
--- a/cli/src/worker/git.ts
+++ b/cli/src/worker/git.ts
@@ -43,8 +43,27 @@ async function update() {
     )
     // Reset first to be sure that an interrupted run didn't create untracked files
     await resetWorktree(context, await getDefault(context))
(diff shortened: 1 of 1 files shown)


effigies:
Does `git annex merge` not do anything else we might need to emulate?

nellh:
> Does `git annex merge` not do anything else we might need to emulate?

It does have conflict resolution but if you're running into conflicts the OpenNeuro CLI just clones and readds changes instead of resolving the conflicts. You'd only need to handle them if you're using DataLad or git-annex directly anyways since we're always generating one set of commits for an upload based on the state on OpenNeuro.

thinknoack (approved):

Issue #3562, opened 2025-09-03
Avoid rekeying unchanged MD5E objects as SHA256E during uploads


nellh:
### What went wrong?

The command line tools will generate new SHA256E keys for annex objects even if the contents have not changed. We should try to avoid doing this as it results in duplicate keys and wastes storage.

### Expected behavior

When an MD5E file has not changed, maintain the old key format. If the file has changed, update to SHA256E.

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

Issue #3561, opened 2025-09-03
one or more files in the most recent draft are missing or do not match checksums (ds004569)


jpeelle:
### What went wrong?

After some challenges (https://github.com/OpenNeuroOrg/openneuro/issues/3559) I was able to upload new files to ds004569, and all files seem present.

However, when I go to "create a new version", I'm presented with the error: "one or more files in the most recent draft are missing or do not match checksums. Please reupload any files listed below to create a version"

However, these files exist.

It does not seem possible to create the new version with these errors present.

These are all .mat and .snirf files, which should be covered by my .bidsignore (we converted an fNIRS dataset to NIfTI, but are also including other nirsy format).

<img width="831" height="745" alt="Image" src="https://github.com/user-attachments/assets/2ceb3406-d660-44c7-9014-795e6dc4d3bb" />

### Expected behavior

Having uploaded new files and passed BIDS validation, expected to be able to commit the updated version.

### Desktop

- OS:
- Browser:
- Version:

### Phone

- Device:
- OS:
- Browser:
- Version:

nellh:
Thank you for detailing this issue, I believe I've found the cause looking at your dataset. It looks like the earlier version of the CLI created entries for these files and the contents (annexed data) did not get transferred correctly as part of your completed upload. A new version will be released soon to address this issue. You will need to rerun your upload but it should only transfer the contents for these missing files. I'll let you know once this fix has been released.

nellh:
@jpeelle This update has been release as part of OpenNeuro 4.38.1.

Could you please try rerunning your upload with the OPENNEURO_LOG=DEBUG environment variable and provide the log if these files are not shown as transferred by the log? They will take a bit to be acknowledged on the OpenNeuro website after the upload is complete but you should see log entries like this to confirm the missing annexed data is transferred:

```
INFO Stored 189475505 bytes for key "SHA256E-s189475505--94c00ab398b0f406e707ee4462f2b0e0d566ad65fa6b9af69ffbff8739b7abc5.mat" from path "sub-01/ses-01/nirs/sub-01_ses-01_task-movie.mat"
```

jpeelle:
Yes, I confirm that with 4.38.1 the upload worked and I was able to approve the new version. Thanks for your (quick) help @nellh!!!!

In case it's useful in the future for other folks: I had deleted a file from my local version, but because of the weird state things were left in the entire upload failed until I restored my local version of that file. I diagnosed this by seeing:

```
[...]

PR #3563, opened 2025-09-03
refactor(cli): Recheck annexed object contents added in previous upload runs

nellh:
This solves a case where the command line tools could be interrupted during upload and fail to retransfer keys it created previously. See #3561

Instead of tracking the new annex keys as they are generated per run, we read all annex keys from HEAD and recheck them against the remote.

Reading the existing keys is also necessary to avoid rechecking MD5E hashes for #3562.


thinknoack (approved):
Reviewed the removal of annexed files, push of files to same dataset. 

transferred the key and didn't make new changes

PR #3575, opened 2025-09-16
Feature/datacite mutation

thinknoack:
adds new mutation and updates resolver for reading/writing datacite.yml


packages/openneuro-server/src/utils/datacite-utils.ts
```diff
@@ -1,61 +1,206 @@
...
+  userId: string,
+  dataciteData: RawDataciteYml,
+) => {
+  const url = `http://datalad-0/datasets/${datasetId}/files/datacite.yml`
+
+  try {
+    // Directly PUT the file with cookie-based JWT auth
+    await superagent
+      .post(url)
+      .set("Cookie", generateDataladCookie(config)(userId))
```
nellh: This should use the request cookies. The reason being it will attribute the commit the user that made the change. Generating a new token here shouldn't be needed.

packages/openneuro-server/src/utils/datacite-utils.ts
```diff
@@ -1,61 +1,206 @@
...
+
+/**
+ * Save datacite.yml back to dataset
+ */
+export const saveDataciteYmlToRepo = async (
+  datasetId: string,
+  userId: string,
+  dataciteData: RawDataciteYml,
+) => {
+  const url = `http://datalad-0/datasets/${datasetId}/files/datacite.yml`
```
nellh: This URL is wrong, it needs to call getDatasetWorker to get the correct offset for a given dataset.

packages/openneuro-server/src/utils/datacite-utils.ts
```diff
@@ -1,61 +1,206 @@
...
+      c.nameIdentifiers?.[0]?.nameIdentifier,
+    )
+    return {
+      name: c.name ||
+        [c.givenName, c.familyName].filter(Boolean).join(" ") ||
+        "Unknown Contributor",
+      givenName: c.givenName,
+      familyName: c.familyName,
+      orcid: contributorOrcid,
+      contributorType: c.contributorType || "Contributor",
```
nellh: Should the default here be researcher?

packages/openneuro-server/src/utils/datacite-utils.ts
```diff
@@ -1,61 +1,206 @@
...
+  const users = await User.find({ orcid: { $in: orcids } }).exec()
+  const orcidToUserId = new Map(users.map((u) => [u.orcid, u.id]))
+
+  return raw.map((c, index) => {
+    const contributorOrcid = validateOrcid(
+      c.nameIdentifiers?.[0]?.nameIdentifier,
+    )
+    return {
+      name: c.name ||
+        [c.givenName, c.familyName].filter(Boolean).join(" ") ||
```
nellh: Might be good to follow the suggested order from the datacite spec which is "Family Name, Given Name".

nellh (changes requested):

packages/openneuro-server/src/utils/datacite-utils.ts
thinknoack: swapped.

packages/openneuro-server/src/utils/datacite-utils.ts
thinknoack: imported the lib

nellh (approved):

PR #3600, opened 2025-10-16
App/contributor user search

thinknoack:


packages/openneuro-app/src/scripts/types/event-types.ts
```diff
@@ -61,60 +74,73 @@ export const mapRawEventToMappedNotification = (
...
-        title =
-          `[${type}] An admin has requested ${targetName} be added as an Author`
-      } else {
-        const capitalizedStatus = status.charAt(0).toUpperCase() +
-          status.slice(1)
-        title = `[${type}] ${targetName} has ${capitalizedStatus} authorship`
-      }
+
+    case "contributorCitationResponse":
+      title = "had their citation responded to"
```
nellh:
```suggestion
      title = "had their citation request for"
```
I think this wording would help.

nellh (approved):
This looks pretty good so far. Added one comment about wording. We might want to default to a more general role in the role selection?

thinknoack:
> We might want to default to a more general role in the role selection?

sorry, I am not understanding. Do you mean when adding users via the UI? 
It currently defaults to Researcher, should it default to something else?

nellh:
> > We might want to default to a more general role in the role selection?
> 
> sorry, I am not understanding. Do you mean when adding users via the UI? It currently defaults to Researcher, should it default to something else?

I pushed a quick fix for it, the issue was just dataset_description authors would not have the researcher role by default.

PR #3639, opened 2025-11-11
Fixes for backup remote exports and annex drop

nellh:
* Block less often with async coroutines for S3 export tasks
* Fix an issue with credentials that would cause drop to fail depending on how the s3 remote was configured
* Fix a bug with an extra argument for the fsck / drop pipeline

### DIFF
--- a/services/datalad/datalad_service/common/asyncio.py
+++ b/services/datalad/datalad_service/common/asyncio.py
@@ -2,14 +2,23 @@ import asyncio
 import subprocess
 
 
-async def run_check(command, dataset_path):
+async def run_check(command, dataset_path, env=None):
     """Helper to run an async command and check for failure"""
-    process = await asyncio.create_subprocess_exec(
-        *command,
-        cwd=dataset_path,
-        stdout=asyncio.subprocess.PIPE,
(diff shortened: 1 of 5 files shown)


services/datalad/datalad_service/common/asyncio.py
```diff
@@ -2,14 +2,23 @@
...
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
+    else:
+        process = await asyncio.create_subprocess_exec(
+            *command,
+            cwd=dataset_path,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
```
effigies, line 21: The default value of `env` is already `None`: https://docs.python.org/3/library/subprocess.html#subprocess.Popen

```suggestion
    process = await asyncio.create_subprocess_exec(
        *command,
        cwd=dataset_path,
        env=env,
        stdout=asyncio.subprocess.PIPE,
        stderr=asyncio.subprocess.PIPE,
    )
```

Did you find this made a difference in practice?

PR #3680, opened 2025-12-05
safely return datatable to ui if missing or failed data load

thinknoack:
@nellh is there any UI Error info we we want to add for this?


thinknoack:
in trying to fix one issue. this created a new issue. deleting to fix

PR #3687, opened 2025-12-09
deps: Update to @bids/validator 2.2.5

nellh:
Update to the latest release of the bids validator.


effigies:
We probably want to fix https://github.com/bids-standard/bids-validator/issues/309 first. Sorry.

Issue #3686, opened 2025-12-09
problem updating ds006576 - GitPushError, Working directory has unstaged changes


emcdevitt:
### What went wrong?

Unable to add new and updated files to ds006576 via CLI

### Expected behavior

I need to add new and modified files to ds006576 using CLI. I have a large dataset with 74 participants (3 sessions each), and I am trying to upload a couple of participants at a time since the dataset is so large. Every time I try to add new participants' data, I run into issues. 

### How to reproduce

Prior to running, I removed the directory listed under `INFO Repo path:` (based on a previous helpdesk suggestion). Then I ran:

deno run --reload -A jsr:@openneuro/cli upload --ignoreWarnings --dataset ds006576 /path/to/dataset

this passes BIDS validation, but results in the following error after adding and transferring files:

```
Transferring annexed files 100.00%                                                    31.5s 210/210 
DEBUG Logger invoked at "push (https://jsr.io/@openneuro/cli/4.41.1/src/worker/git.ts:494:12)"
INFO Git-annex branch is up to date.
Pushing changes...
error: Uncaught (in worker "") (in promise) GitPushError: One or more branches were not updated: 
  - refs/heads/main: Working directory has unstaged changes
    at _push (file:///Users/eam7/Library/Caches/deno/npm/registry.npmjs.org/isomorphic-git/1.27.1/index.cjs:12627:11)
    at Object.runMicrotasks (ext:core/01_core.js:693:26)
    at processTicksAndRejections (ext:deno_node/_next_tick.ts:59:10)
    at runNextTicks (ext:deno_node/_next_tick.ts:76:3)
    at eventLoopTick (ext:core/01_core.js:186:21)
    at async Object.push (file:///Users/eam7/Library/Caches/deno/npm/registry.npmjs.org/isomorphic-git/1.27.1/index.cjs:12707:12)
    at async push (https://jsr.io/@openneuro/cli/4.41.1/src/worker/git.ts:544:5)
    at async PromiseQueue.processQueue (https://jsr.io/@openneuro/cli/4.41.1/src/worker/queue.ts:17:24)
error: Uncaught (in promise) null
```
When I navigate to the newly created INFO Repo path and check git status, it says that my updated participants.tsv is not staged for commit. So I think this is the file causing the issue but not sure how to fix this. 

### Desktop

- OS: Mac 15.7.1
- Browser: 
- Version: openneuro 4.41.1

### Phone

- Device:
- OS:
- Browser:
[...]

PR #3691, opened 2025-12-11
fix(cli): Apply eol git attributes during upload on local repo

nellh:
We need to apply the EOL filter before adding these files to the index on the local side. This avoids some churn when mixing edits using web app and the command line tool as well, since the transformed file will be the same locally and remotely.

Fix for #3686.

### DIFF
--- a/cli/src/worker/transformEol.test.ts
+++ b/cli/src/worker/transformEol.test.ts
@@ -1,0 +1,135 @@
+import { assertEquals } from "@std/assert"
+import { transformEol } from "./transformEol.ts"
+import type { FileHandle } from "node:fs/promises"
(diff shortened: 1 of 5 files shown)


cli/src/worker/transformEol.test.ts
effigies, line 1: Should we have a test to make sure that this doesn't modify binary data?

nellh: I thought that might be needed for text=auto but looking at how git behaves it's needed in both cases. It's added now matching the test git uses (looking for a null byte in the first 8000 bytes) and the file is passed through untransformed in that case.

PR #3737, opened 2026-01-26
FIX getRepoAccess() callers to pass instance URL as first argument

mvdoc:
I was trying to upload a new dataset with the CLI (v4.4.0), but kept getting login errors despite having logged in multiple times. The following was fixed with Claude. Now I'm able to upload the dataset (ongoing).

---
Claude's comment:

> Commit 2fae59b13 changed getRepoAccess() signature to require an instance URL as the first parameter, but these three callers were not updated. This caused getConfig(instance) to look up tokens using the dataset ID instead of the URL, resulting in "Run openneuro login" errors.

### DIFF
--- a/cli/src/commands/download.ts
+++ b/cli/src/commands/download.ts
@@ -28,7 +28,7 @@ export async function downloadAction(
 ) {
   const datasetId = accession_number
   const clientConfig = readConfig()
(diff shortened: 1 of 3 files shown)


effigies:
I think that https://github.com/OpenNeuroOrg/openneuro/pull/3739 has made this unnecessary. Can you confirm?

PR #3740, opened 2026-01-27
FIX skip .git directories in submodules

mvdoc:
I typically save derivatives as nested submodules/datalad datasets. I was uploading a dataset and noticed that the openneuro CLI was correctly skipping the parent `.git` folder but not the `derivatives/.git`. This could cause the CLI to upload duplicate annex objects as regular files when uploading datasets containing git-annex submodules.

This fix creates a new `shouldIncludeFile()` function to extract the inclusion logic, adding a check for dot files anywhere in the relative path. It also adds unit tests for the new shouldIncludeFile() function.

### DIFF
--- a/cli/src/commands/upload.test.ts
+++ b/cli/src/commands/upload.test.ts
@@ -1,0 +1,68 @@
+import { assertEquals } from "@std/assert/equals"
+import { hiddenPathPattern, shouldIncludeFile } from "./upload.ts"
+
+Deno.test("hiddenPathPattern matches hidden files/dirs in subpaths", () => {
+  assertEquals(hiddenPathPattern.test("derivatives/.git/config"), true)
+  assertEquals(hiddenPathPattern.test("derivatives/.git/annex/objects/xx/yy/key"), true)
+  assertEquals(hiddenPathPattern.test("derivatives/.datalad/config"), true)
+  assertEquals(hiddenPathPattern.test("a/.git/b"), true)
+  assertEquals(hiddenPathPattern.test("sub/.gitignore"), true)
+  // Should not match dotfiles at root (no leading slash)
+  assertEquals(hiddenPathPattern.test(".git/config"), false)
+  assertEquals(hiddenPathPattern.test(".bidsignore"), false)
+  // Should not match files that just contain a dot in name
+  assertEquals(hiddenPathPattern.test("my.gitconfig"), false)
+  assertEquals(hiddenPathPattern.test("file.nii.gz"), false)
+})
+
+Deno.test("shouldIncludeFile() includes regular dataset files", () => {
+  assertEquals(shouldIncludeFile("dataset_description.json"), true)
+  assertEquals(shouldIncludeFile("sub-01/anat/sub-01_T1w.nii.gz"), true)
+  assertEquals(shouldIncludeFile("derivatives/sub-01/func/bold.nii.gz"), true)
+})
+
+Deno.test("shouldIncludeFile() includes .bidsignore", () => {
+  assertEquals(shouldIncludeFile(".bidsignore"), true)
+})
+
+Deno.test("shouldIncludeFile() skips root hidden files and directories", () => {
+  assertEquals(shouldIncludeFile(".git/config"), false)
+  assertEquals(shouldIncludeFile(".git/annex/objects/abc/def/SHA256--xyz"), false)
+  assertEquals(shouldIncludeFile(".datalad/config"), false)
+  assertEquals(shouldIncludeFile(".gitattributes"), false)
+})
+
+Deno.test("shouldIncludeFile() skips hidden files/dirs in submodules", () => {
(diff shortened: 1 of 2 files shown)


effigies:
Curious about @yarikoptic's thoughts here, as our number one supporter of dotfiles in datasets...

This would get rid of all dotfiles except `.bidsignore`. I think that's probably reasonable, and people who want to arrange their datasets differently can use the git interface instead of the CLI uploader.

mvdoc:
I think I should also clarify what I was trying to accomplish. I was preparing a dataset for release, and set it up privately with several submodules as derivatives (e.g., fmriprep, freesurfer, qa, etc.). I then wanted to upload it as a "flat" datalad dataset, so that everything could be accessed from the same dataset on OpenNeuro.

In the end I did not manage to use the CLI because despite my fixes (including #3737), I kept getting upload errors (similar to #3741). So I manually created a new "flat" dataset and uploaded with the git interface, which worked great.

yarikoptic:
> Curious about @yarikoptic's thoughts here, as our number one supporter of dotfiles in datasets...

not all dotfiles born alike!  overall the semantic for `dotfile` is "hidden" , not necessarily "*local*" or "not needed" so we need to explicitly separate that somehow otherwise.

Overall,
- VCS directories AFAIK should not be uploaded since they are not created to be redistributed as is and carry *local* to the repository information. It is just "coincidental" that they are "dotfiles".
- VCS auxililary files shared within repositories needed for their operational procedures (like `.gitignore`, `.gitattributes`, `.gitmodules`) ALSO make little sense for distribution outside of the VCS context.
  - `CVSROOT` folders and alike should also be avoided! ;)
- some trashy `.Trash*` added by some OS, or `.*~` backup files added by editors are also *not needed*

dotfiles concerning some known semantics for datasets 

- `.bidsignore`
- `.bids-validator.json`
- `.datalad/`

so, so far I think we are doomed to be explicit and in VCS agnostic clients/mode ignore VCS files, and added known to be useful .dotfiles like `.bids*` and `.datalad` (anything else?) and WARN about other .dotfiles we do not know about but ignore them unless explicitly listed for upload.

Here -- it is not per se about .git/ subfolders at all but rather that derivatives should not be uploaded along with the `raw` dataset, (unless it is a "study" dataset and it is allowed to clump upload them all at once)?  

IMHO it should have ignored `derivatives/*` subfolders within a "raw" dataset and advised to upload them separately? (with or without .git under -- separate logic should apply as re VCS folders and regardless of the depth: I could have had `sub-*/` submodules in a big BIDS dataset)

mvdoc:
> Here -- it is not per se about .git/ subfolders at all but rather that derivatives should not be uploaded along with the raw dataset, (unless it is a "study" dataset and it is allowed to clump upload them all at once)?

I did not know this. Several datasets of openneuro have a derivatives folder, and it seems to be OK according to the current BIDS specification https://bids-specification.readthedocs.io/en/stable/common-principles.html#storage-of-derived-datasets

mvdoc:
To clarify a bit more why (I thought) the CLI needs to skip submodules with `.git/annex` here:

Without skipping `submodule/.git`, the CLI would copy both the original object (following symlinks) and the corresponding annex object in `submodule/.git/annex/objects`.
This would (probably) duplicate the data being pushed. I don't know if git-annex in this case would be smart enough to automatically resolve the duplicates?

effigies:
Yes, OpenNeuro does not restrict the use of derivatives. It's the only way we support non-compliant derivatives, as they could not be uploaded as standalone datasets.

yarikoptic:
> I did not know this. Several datasets of openneuro have a derivatives folder, and it seems to be OK according to the current BIDS specification https://bids-specification.readthedocs.io/en/stable/common-principles.html#storage-of-derived-datasets

Oh my gray hairs (where left) -- what do I hear from my dearest coauthor on the initial YODA poster!  "OK" doesn't mean "perfect" or even "good". It is just that -- "Ok". ;)  ultimately @effigies is the authority but I would recommend to modularize and upload separately whenever possible -- and thus refferring to @effigies "the only way" for non-compliant!

But hence, again, by default I think derivatives should NOT be uploaded unless explicitly requested for e.g. that reason of being non-compliant, and I hope small.  I tend to just store `derivatives/bids-validator` within 'raw' and the rest moves into the level of "BIDS study" to join them up. E.g. explore those: https://github.com/OpenNeuroStudies/OpenNeuroStudies/ 

> This would (probably) duplicate the data being pushed. I don't know if git-annex in this case would be smart enough to automatically resolve the duplicates?

that is the git and git annex feature/beauty of relying on checksum for content storage. So yes -- you transfer twice, git and git-annex would store once.

mvdoc:
Closing this. In the end I am following master YODA and uploading fMRIPrep and freesurfer derivatives as separate datasets. And it seems like anyway git-annex would be smart enough not to upload things twice.

PR #3820, opened 2026-03-18
feat(server): Send data retention policy notices for stale draft data

nellh:
This implements data retention policy notices for draft data approaching the 28 data limit. Rules implemented:

* A notice is sent when a draft is 14 days stale, then 21 days stale, and a final notice when deletion is reached (28 days).
* If a dataset is already 28 days old when this begins checking, it will send the 14 day notice immediately, starting the clock from there.
* No data is automatically removed currently, we just flag these datasets and will script dropping draft data based on the state recorded.

Automatic deletion can be enabled at some point but we should add an admin opt out mechanism first too to hold this for certain datasets if needed.

A few related fixes are included, it was possible for the redis queue to get ignored when it was created after the consumer was created. This edge case is now handled. Updates our email notices to come from The OpenNeuro Team instead of The CRN Team. Generalizes some of the scheduling here so we can reuse it for other queued tasks on a schedule at the API level. Fixes a longstanding issue where the snapshot reminder message was sent directly after upload completed, instead it is sent after 24 hours if no snapshot was created.


effigies (approved):
> Fixes a longstanding issue where the snapshot reminder message was sent directly after upload completed, instead it is sent after 24 hours if no snapshot was created.

Extremely pleasing.

PR #3849, opened 2026-03-30
fix: Resolve reviewer access regression

nellh:
Fixes reviewer users being immediately logged out and provides better compatibility for the frontend code by generating a mock user object for reviewers. Includes a fix to hide write actions from reviewers, previously a reviewer could make these requests even though they would throw errors.


effigies:
This was a regression introduced in https://github.com/OpenNeuroOrg/openneuro/pull/3829?

nellh:
> This was a regression introduced in #3829?

Yes. The user query would always fail for reviewer users and handling that error in the client meant we'd log them out.

PR #3857, opened 2026-04-01
feat(server): Change age type from Int to Float in GraphQL schema

effigies:
Allows non-integer ages from BIDS participants.tsv to be stored without truncation. Updates both the GraphQL SDL types and the Elasticsearch mapping for subjectMetadata.age.

This also sanitizes the validator output to filter non-finite-numeric ages and set them as `null`. The reason is that the use for these data is ElasticSearch, which needs continuous values. If the validator wants to treat `"89+"` as `89`, it may make that change. OpenNeuro will simply filter values it is given.

Fixes an issue where datasets with non-integral ages would fail to display metadata such as modalities.

### DIFF
--- a/packages/openneuro-server/src/graphql/schema.ts
+++ b/packages/openneuro-server/src/graphql/schema.ts
@@ -323,7 +323,7 @@ export const typeDefs = `
 
   input SubjectMetadataInput {
     participantId: String!
-    age: Int
+    age: Float
     sex: String
     group: String
   }
@@ -355,7 +355,7 @@ export const typeDefs = `
     openneuroPaperDOI: String
     seniorAuthor: String
     adminUsers: [String]
-    ages: [Int]
+    ages: [Float]
     modalities: [String]
     grantFunderName: String
     grantIdentifier: String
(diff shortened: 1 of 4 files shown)


packages/openneuro-server/src/graphql/schema.ts
```diff
@@ -323,7 +323,7 @@ export const typeDefs = `
 
   input SubjectMetadataInput {
     participantId: String!
-    age: Int
+    age: Float
```
nellh, line 326: If this is Float, the non-numeric values will never get passed to sanitizeAge. We could use the JSON scalar or sanitize to Float/null on the client side (either in the validator or the OpenNeuro worker).

nellh (changes requested):

packages/openneuro-server/src/graphql/schema.ts
effigies: Ah, I assumed that the change would need to be in JS. Adjusted in the Python service. Does that cover it?

nellh (approved):

PR #3854, opened 2026-04-01
fix: Refactor snapshot->downloadFiles resolver to use worker side cache

nellh:
Implements a cleaner version of the `downloadFiles` resolver. Instead of recursively following the tree API, we instead maintain a file cache on the workers for each tag and just stream this cache on request to API clients. If the cache isn't ready return null.

This batches the significant performance cost of returning the remote URLs for all files in the tree which can take minutes per request on large datasets.

Example usage:

```graphql
query downloadFiles {
  dataset(id: "ds001234") {
    latestSnapshot {
      downloadFiles {
        id
        filename
        urls
      }
    }
  }
}
```

In the case of a pending export or cache:
```json
{
  "data": {
    "dataset": {
      "latestSnapshot": {
        "downloadFiles": null
      }
   }
}
```

Once populated:
```json
{
  "data": {
    "dataset": {
      "latestSnapshot": {
        "downloadFiles": [
          {
            "id": "434d719164ca3cea650a86b18dd5ec2051fb665d",
            "filename": "dataset_description.json",
            "urls": [
              "https://openneuro.org/crn/datasets/ds001234/objects/a4116179a7728929e9a9579189cb50fed2b483bf"
            ]
          },
          {
            "id": "67bb157099cbc45f68fba7320e74f3f17b347c1e",
            "filename": "derivatives/mriqc/anatMRIQC.csv",
            "urls": [
              "https://s3.amazonaws.com/...."
            ]
          },
          ...
        ]
      }
    }
  }
}
```

See #3846

Issue #3846, opened 2026-03-28
`downloadFiles` returns only root-level files (cache never populated)


hoechenberger:
### Summary

On the GraphQL endpoint, the `downloadFiles` field on `Snapshot` is documented as returning a "single list of files to download this snapshot," but in practice it only returns root-level files — never files inside subdirectories. This forces clients like [openneuro-py](https://github.com/openneuro-py/openneuro-py) to recursively call `files(tree: ...)` one directory at a time, resulting in an N+1 query pattern for every dataset download.

### Expected behavior

```graphql
{
  snapshot(datasetId: "ds000117", tag: "1.0.4") {
    downloadFiles {
      filename
      urls
      size
    }
  }
}
```

Should return a flat list of **all** files in the snapshot, recursively — as described in [#2797](https://github.com/OpenNeuroOrg/openneuro/pull/2797).

### Actual behavior

The query returns only root-level files. For example:

| Dataset | Root-level files returned | Subdirectories (not traversed) |
|---------|--------------------------|-------------------------------|
| ds000001 | 5 files (all root-level) | 16 subdirectories |
| ds000117 | 26 files (all root-level) | 19 subdirectories |
| ds003104 | 5 files (all root-level) | 3 subdirectories |

No returned filename contains a path separator — subdirectory contents are never included.

### Root cause

`downloadFiles` calls `getFilesRecursive` in `datalad/snapshots.ts`, which recursively calls `getFiles` in `datalad/files.ts`. However, `getFiles` has a `doNotCache` guard:

https://github.com/OpenNeuroOrg/openneuro/blob/fb46610d70caa5d56c2e37640732cbe8de60f1be/packages/openneuro-server/src/datalad/files.ts#L109-L113

When files don't yet have S3 URLs (which appears to be common), `doNotCache(true)` is called, and the recursive results are never persisted to the Redis cache. On subsequent requests, the cache miss means only the root-level files are returned.

This was introduced in [#2797](https://github.com/OpenNeuroOrg/openneuro/pull/2797) (April 2023), which describes the feature as "a new flat API call that is indefinitely cached once a snapshot is fully exported." The caching precondition (all files having S3 URLs) appears to rarely be met in practice.

### Impact on clients

Without a working `downloadFiles`, clients must manually traverse the directory tree:

1. Query `files(tree: null)` to get root entries
2. For each directory, query `files(tree: "dirname")`
3. Recurse into subdirectories

[...]

PR #3865, opened 2026-04-06
refactor(graphql)!: Unified Redis caching for file trees

nellh:
This refactor migrates the Redis file tree cache to keyed per tree with an optimization to do a single Redis scan for commits where all trees are cached, this allows individual keys to be invalidated without dropping the entire cache and progressive addition to the cache with new commits (that often contain only a small diff from the parent tree). Also separates presigned URLs into their own cache structure on top of the file cache, no longer requiring us to drop the entire cache for unchanged trees.

* No longer caches presigned URLs for public dataset objects. These are served as anonymous S3 URLs.
* Switches to AWS signature V4 for private dataset objects.
* Implements a recursive option for the dataset files() resolver for both snapshots and drafts.
* Breaking change to the API since this cache structure requires dropping the composite location and git hash ids. The `key` field is no longer required.
* Moves all the presigning and fallback URL features to the API and out of the worker. This also lets us scale CPU demand for recomputing some of this independent of the workers.

See #3846. Replaces #3854.

There's still a few cases where the full file cache is dropped and we should flag cases where datasets are fully exported but cache miss here but this PR is already pretty big.

Requires a full Redis cache drop with deployment.


nellh:
This handles the single tree case better than the existing implementation but could use some further work on the recursive case. Going to merge and make new PRs to refine since this is already rather large. It currently works well for datasets with a small number of large trees but the many small tree cases is pretty bad on a cold cache.

hoechenberger:
Amazing work, @nellh! Once deployed, I am happy to try this out with `openneuro-py`!

PR #3868, opened 2026-04-07
refactor(worker): Use pygit2 and support batching for tree requests

nellh:
To support recursive file tree queries, we need to batch request trees during a cache miss to avoid overwhelming the backend with HTTP requests for trees. This implements a batch version of get_repo_files to return multiple trees in one call using pygit2.

### DIFF
--- a/services/datalad/datalad_service/common/annex.py
+++ b/services/datalad/datalad_service/common/annex.py
@@ -298,7 +298,7 @@ def _read_tree_pygit2(repo, tree_obj):
     return files
 
 
-async def get_repo_files_batch(dataset_path, trees):
+async def get_repo_files(dataset_path, trees):
     """Read files for multiple trees using pygit2 with shared URL resolution."""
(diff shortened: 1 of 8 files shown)


services/datalad/datalad_service/common/annex.py
```diff
@@ -283,62 +230,109 @@ async def get_repo_urls(path, files):
...
+        filename = entry.name
+        if _should_skip(filename):
+            continue
+        mode = entry.filemode
+        hex_hash = str(entry.id)
+        if mode == 0o120000:
+            # Annexed file symlink — resolve target to get annex key
+            blob = repo[entry.id]
+            target = blob.data.decode('utf-8').rstrip()
+            key = target.split('/')[-1]
```
effigies: This is probably going to cause problems for in-tree symlinks (e.g., `ln -s rh.pial.orig rh.pial` inside a FreeSurfer directory). I believe ds007116 has some examples.

PR #3885, opened 2026-04-13
refactor(server): Rework DOI minting to use DataCite JSON schema and endpoint

nellh:
* Adds support for transitioning between DOI states depending on if a dataset is published. Before publishing, DOIs will be created for snapshots in the draft state and made findable. Admin unpublish will move these back to registered.
* Consolidates the metadata extraction in one place so we can extend this to support all datacite.yml fields.
* Adds a validation step that passes errors back to the mutation.
* Uses a generated TypeScript definition based on a JSON schema version of the DataCite metadata schema. The types used here are still a little OpenNeuro specific but derived from the generated types wherever possible.


rwblair (approved):
Looks great. Using json schema from their developers python library is much cleaner than the code generation I had been looking at. Validation should prevent the sentry errors for doi minting that I'm seeing.

PR #3890, opened 2026-04-15
rf(graphql): Generate GraphQL schema with Pothos

### LINKED ISSUE #3831: Use graphql codegen?
### What would you like to see added?

At least part of #3830 might have been caught if we used [graphql-codegen](https://the-guild.dev/graphql/codegen), as the names of the passed variables should be statically checkable.

For using `id` versus `orcid`, it looks like there's a way to do [tagged types](https://medium.com/@ethanresnick/advanced-typescript-tagged-types-improved-with-type-level-metadata-5072fc125fcf) in TypeScript:

```ts
declare const tags: unique symbol;

type Tagged<BaseType, Tag extends PropertyKey> = 
  BaseType & { [tags]: { [K in Tag]: void } };

type UserID = Tagged<string, 'User'>
type ORCID = Tagged<string, 'ORCID'>
```

This is really two separate tasks, and both have potentially far-reaching tendrils, but once done I think we'll end up with something safer.

### Alternatives

Instead of codegen, we could start using [TypedDocumentNode](https://www.apollographql.com/docs/react/data/typescript#using-typeddocumentnode) manually. This would be lower-impact and could be done gradually, but I think it's only worth it if we don't plan to do code generation, as there will still be multiple sources of truth to collate.

### Do you have any interest in helping implement the feature?

Yes

### Additional information / screenshots

_No response_

effigies:
This is an initial draft at transitioning away from hand-written GraphQL schema to generated schema backed by our Mongoose models.

The import flow is:

```mermaid
flowchart
  models/ --> graphql/schema/ --> app.ts
  models/ --> graphql/resolvers/ --> graphql/schema/
```

`graphql/schema` acts as a coordination layer that will enforce type consistency, as well as allow us to emit a schema document that we can use to generate types for clients.

Closes #3831.


packages/openneuro-server/src/datalad/draft.ts
```diff
@@ -13,7 +13,7 @@ type DraftInfo = {
   hexsha: string // Duplicate of ref for backwards compatibility
   tree: string
   message: string
-  modified: Date
+  modified: string
```
effigies, line 16: This change just reflects that `.json()` (L33) will not return a `Date`, here. It seemed better not to change the runtime.

packages/openneuro-server/src/models/user.ts
```diff
@@ -12,7 +12,7 @@ export interface UserDocument extends Document {
   // User's preferred name (visible)
   name: string
   // Login provider
-  provider: StaticRangeInit
+  provider: string
```
effigies, line 15: This feels like it must have been a snippet gone wrong.

packages/openneuro-server/schema.graphql
effigies, line 1: This is here for comparison purposes, and I intend to remove it before merge. LMK if you have other thoughts.

packages/openneuro-server/src/graphql/resolvers/draft.ts
```diff
@@ -10,9 +10,16 @@ import { filterRemovedAnnexObjects } from "../utils/file.js"
...
+import type { GraphQLContext } from "../builder"
+
+export type DraftShape = { id: string; revision?: string; modified?: Date }
 
 // A draft must have a dataset parent
-export const draftFiles = async (dataset, args, { userInfo }) => {
+export const draftFiles = async (
+  dataset: DraftShape,
+  args: { tree?: string | null; recursive?: boolean | null },
+  { userInfo }: Pick<GraphQLContext, "userInfo">,
```
effigies, line 21: The use of `Pick<GraphQLContext, "userInfo">` here reflects that some calls only pass `{ userInfo }` instead of the full context. Do we want to declare this in the type directly, or would we rather do:

```suggestion
  { userInfo }: GraphQLContext,
```

And have the callers pass the full context? Passing the full context means less action at a distance if we ever start using `user` here, but more rewriting now.

packages/openneuro-server/src/graphql/resolvers/reviewer.ts
```diff
@@ -37,7 +42,10 @@ export async function deleteReviewer(
 /**
  * Resolver for dataset reviewers
  */
-/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
-export function reviewers(obj, _, { user, userInfo }) {
+export function reviewers(
+  obj: { id: string },
+  _: unknown,
+  { user: _user, userInfo: _userInfo }: GraphQLContext,
```
effigies:
```suggestion
  _ctx: GraphQLContext,
```

effigies (commented):
I think this is ready for review. This won't get us full safety around schema-related types, but this at least progress in that direction, and I think we can gradually improve.

A lot of `as never`s will be removable if we can get to the point where we can set `strict: true` for tsconfig: https://pothos-graphql.dev/docs/guide#set-up-typescript

packages/openneuro-server/src/graphql/resolvers/draft.ts
nellh: `userInfo.id` and `user` are equivalent. We can just keep this and drop the user value from the context at some point.

packages/openneuro-server/src/graphql/builder.ts
```diff
@@ -0,0 +1,42 @@
...
+  userInfo: UserInfo
+}
+
+export const builder = new SchemaBuilder<{
+  Context: GraphQLContext
+  Scalars: {
+    ID: { Input: string; Output: string }
+    Date: { Input: string; Output: Date }
+    DateTime: { Input: string; Output: Date }
+    BigInt: { Input: string; Output: string }
```
nellh:
```suggestion
    BigInt: { Input: number; Output: number }
```
We don't actually handle integers larger than MAX_SAFE_INTEGER for this field but we do need larger values than GraphQL's INT type. We can just use `number` instead of `number | bigint` for that reason.

nellh (approved):
This looks good, I think we can merge and continue to converge on more type safety. One suggestion for the BigInt scalar.

effigies:
Merged your suggestion, verified no schema change, and removed the `schema.graphql` document. This should be good to go after a final test run.

PR #3891, opened 2026-04-15
fix(app): Allow admin users to request fsck at any time

nellh:
Quick fix to allow site admin users to request fsck of a tree at any time.


effigies (approved):
Thanks!

PR #3910, opened 2026-04-23
refactor(server): Lazy-initialize Redis, Redlock, and Elasticsearch clients, fix import cycles

effigies:
The goal here is to be able to dump the graphql schema without using test infrastructure to mock import-type services.

This required deferring initialization of our JWT signing key as well as the elasticsearch, redis and redlock clients. Further, two import cycles had to be removed before we could call the script with `yarn tsx`.

The script is moved to the monorepo `/scripts` directory, and can be run with `yarn dump-schema`, which calls `tsx scripts/dump-schema.ts`.

PR #3911, opened 2026-04-25
refactor(graphql): Generate initial app types from schema

effigies:
This PR establishes a process for converting openneuro-app types from hand-written to generated. It:

1) Creates a code generation script. I had wanted to do this as a single script that imports from the server and outputs to the types, but pnp errors made me compromise by using the existing script and just gitignore-ing the intermediate file.
2) Replaces user-types.ts in 3626bd8e5af810652b6eceed628e880725540dcb. This seemed the simplest and was a decent proof of concept.
3) Replaces dataset-events.ts in d58bcd321b3ca3913f316150fc85f8179a1f86ef. This is a slightly more complex set of types.

If we're happy with this, I would suggest making further changes in a separate PR to keep the infrastructure changes separated from the majority of the mostly-mechanical type changes.

I will not have time to work on this further until after this coming week. Feel free to examine or ignore.

Builds on #3910, so those changes are duplicated here.


effigies:
Just realized I didn't finish this...

Issue #3921, opened 2026-05-06
Validate git commits/tags


effigies:
### What would you like to see added?

The validator currently validates the worktree, rather than the commit that triggered the validation step. Commits performed in rapid succession may get tagged with the wrong validation information. Further, re-validating old commits with an updated validator is not currently possible.

The upcoming bids-validator v3 release should allow us to pass `--git-ref $COMMIT` or `--git-ref $TAG`.

I think this should be a straightforward update to add `["--git-ref", ref]` to the validator call in datalad_service, but I want to open the issue in case there are other considerations here.

xref

* https://github.com/bids-standard/bids-validator/pull/370
* https://github.com/bids-standard/bids-validator/pull/384

### Alternatives

Status quo: Continue validating HEAD, as this has little chance of disruption.

### Do you have any interest in helping implement the feature?

Yes

PR #3927, opened 2026-05-06
feat: Validate datasets by git reference (uses @bids/validator 3)

effigies:
First commit bumps versions to ensure there's no breakage. Second adjusts import paths.

I can go ahead and address #3921 in this or save it for a different PR.

I would suggest waiting to merge until after the full 3.0 release, but I wanted to test this before making the full release. I'm not sure if there is additional validation beyond CI before we should go ahead.


effigies:
The Python failure is that deno won't run something <24h old. I can't set a package-specific exemption for the CLI, and a blanket `--minimum-dependency-age P0D` would apply to our dependencies as well, so it's probably easiest just to wait until tomorrow.

Otherwise this should be good to go.

nellh (approved):

PR #3938, opened 2026-05-13
Add fitlins to list of derivative suffixies indexed from s3

rwblair:
@jbwexler pushed the fitlins results to `s3://openneuro-derivatives` and the github org `OpenNeuroDerivatives` months ago.

### DIFF
--- a/packages/openneuro-server/src/graphql/resolvers/__tests__/derivatives.spec.ts
+++ b/packages/openneuro-server/src/graphql/resolvers/__tests__/derivatives.spec.ts
@@ -33,4 +33,15 @@ describe("GraphQL derivatives", () => {
       })
     })
   })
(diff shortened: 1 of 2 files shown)


packages/openneuro-server/src/graphql/resolvers/__tests__/derivatives.spec.ts
```diff
@@ -33,4 +33,15 @@ describe("GraphQL derivatives", () => {
       })
     })
   })
+  it("returns expected values for fmriprep", () => {
```
nellh:
```suggestion
  it("returns expected values for fitlins", () => {
```

nellh (approved):

jbwexler:
There are also some xcpd datasets. Would it be worth adding that as well?

PR #4012, opened 2026-07-09
Add createMosaic mutation and related functionality

jbwexler:
This sets up the basics for incorporating the bids-mosaic tool into openneuro:
1) Adds a createMosaic mutation that runs bids-mosaic on a particular dataset/snapshot, saves the output pdf to /datalad/mosaics/\<dataset\>/\<dataset\>-\<hexsha\>_mosaic.pdf, and then triggers the updateMosaic mutation to update the database.
2) Adds a boolean mosaic query within snapshot, indicating whether bids-mosaic has been run.

I haven't added any tests yet but happy to do so once everything else is ready to go.

I thought I would start with the basics here and then we can discuss whether we want to incorporate it within the UI and/or have it be auto-triggered. Another idea from @effigies was to keep the 2D images annexed in a separate branch so unchanged images don't have to be regenerated when running on a future snapshot (currently they get created in /tmp and then deleted).

### DIFF
--- a/packages/openneuro-server/src/graphql/resolvers/mosaic.ts
+++ b/packages/openneuro-server/src/graphql/resolvers/mosaic.ts
@@ -1,0 +1,101 @@
+import config from "../../config"
+import { generateDataladCookie } from "../../libs/authentication/jwt"
+import { getDatasetWorker } from "../../libs/datalad-service"
+import Mosaic from "../../models/mosaic"
+import { getRedis } from "../../libs/redis"
+import CacheItem from "../../cache/item"
+import { CacheType } from "../../cache/types"
+import type { GraphQLContext } from "../builder"
+
+
+export const mosaic = async (dataset, _, _context: GraphQLContext) => {
+  const cache = new CacheItem(
+    getRedis(),
+    CacheType.mosaic,
+    [dataset.id, dataset.revision],
+  )
+  const cacheResult = await cache.get((doNotCache) => {
+    return Mosaic.findOne({
+      id: dataset.revision,
+      datasetId: dataset.id,
+    })
+      .exec()
+      .then((data) => {
+        if (data) {
+          return true
+        } else {
+          doNotCache(true)
+          return false
+        }
+      })
+  })
+  return cacheResult
+}
+
+/**
+ * Snapshot mosaic resolver
+ */
+export const snapshotMosaic = async (snapshot, _, context: GraphQLContext) => {
+  const dataset = {
+    id: snapshot.id.split(":")[0],
+    revision: snapshot.hexsha,
+  }
+  return mosaic(dataset, _, context)
+}
+
+/**
+ * Save mosaic data returned by the datalad service
+ *
+ * Returns only a boolean if successful or not
+ */
+export const updateMosaic = (obj, args) => {
+  return Mosaic.updateOne(
+    {
+      id: args.mosaic.id,
+      datasetId: args.mosaic.datasetId,
+    },
+    args.mosaic,
+    {
+      upsert: true,
+    },
+  )
+    .exec()
+    .then(() => true)
+}
+
+export const mosaicUrl = (datasetId, ref) => {
+  return `http://${
+    getDatasetWorker(
+      datasetId,
+    )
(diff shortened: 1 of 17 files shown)


services/datalad/datalad_service/tasks/mosaic.py
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+
+@broker.task
+async def create_mosaic(dataset_id, dataset_path, ref, cookies=None, user=''):
+    async with redis_client() as client:
+        lock = client.lock(f'mosaic-lock:{dataset_id}:{ref}', timeout=60 * 60 *4)
+        if await lock.acquire(blocking=False):
+            try:
+                out_dir_path = Path(f"/datalad/mosaics/{dataset_id}")
```
nellh: This should use DATALAD_DATASET_PATH as this volume can be configured to be in different locations.

services/datalad/datalad_service/tasks/mosaic.py
```diff
@@ -0,0 +1,60 @@
...
+    return {
+        'query': 'mutation ($info: MosaicInput!) { updateMosaic(mosaic: $info) }',
+        'variables': {
+            'info': mosaicInput,
+        },
+    }
+
+
+@broker.task
+async def create_mosaic(dataset_id, dataset_path, ref, cookies=None, user=''):
```
nellh: Looks like the ref isn't actually referenced for the mosaic being produced. It would be good to read from that version of the dataset.

packages/openneuro-server/src/graphql/resolvers/mosaic.ts
```diff
@@ -0,0 +1,101 @@
...
+import { CacheType } from "../../cache/types"
+import type { GraphQLContext } from "../builder"
+
+
+export const mosaic = async (dataset, _, _context: GraphQLContext) => {
+  const cache = new CacheItem(
+    getRedis(),
+    CacheType.mosaic,
+    [dataset.id, dataset.revision],
+  )
```
nellh: The Redis cache here is pretty unnecessary for a single MongoDB query. I'd remove it to simplify this.

nellh (commented):
A few comments. This looks pretty good, if we're keeping the mosaics in the datasets volume we'd just need an endpoint to serve them (on_get for the mosaic handler would work) and a resolver for that. The advantage of moving them into a branch is we could reuse the git object endpoint to stream them but that might not be worth the added complexity.

Let's start with an admin tab button to trigger mosaic creation and automate if that seems to be working well?

services/datalad/datalad_service/tasks/mosaic.py
jbwexler: I was looking at validator.py for ideas on how to do this but it looks like ref doesn't get used there either. I can't tell if I'm missing something or if that's a bug.

nellh: It was recently implemented in https://github.com/OpenNeuroOrg/openneuro/pull/3927

services/datalad/datalad_service/tasks/mosaic.py
jbwexler: Should be resolved by [6f5a339](https://github.com/OpenNeuroOrg/openneuro/pull/4012/commits/6f5a33979f2926c6f1bdf6af6a8384e72f148304).

packages/openneuro-server/src/graphql/resolvers/mosaic.ts
jbwexler: Should be resolved by [7f99a56](https://github.com/OpenNeuroOrg/openneuro/pull/4012/commits/7f99a5699076d2d4983b62948580edde079afe5b).

services/datalad/datalad_service/tasks/mosaic.py
jbwexler: This should be fixed by https://github.com/OpenNeuroOrg/openneuro/pull/4012/changes/55ad0fdc1806094e7c505581c3fd2c54135bb7d9. Using streaming to create mosaics seems a bit slow on my machine so I’m curious how performance will be on the server.

nellh (approved):

PR #4027, opened 2026-07-22
fix(server): Increase json payload limit to 200mb

nellh:
Occasionally the 100mb limit is exceeded by validation output even with the 50k issues limit.


effigies:
FYI, ds008192 is a dataset that is currently stuck on validation pending. Piping the validator output into `wc -c` shows well under the 100mb limit, so this can't be the only source of hanging validation.

PR #4028, opened 2026-07-22
chore: Update @bids/schema to 1.2.7-datacite

effigies:
Approach: Rebased https://github.com/bids-standard/bids-specification/pull/2221 onto `schema-1.2.7`, and reran `uv run bst export > schema-1.2.7-datacite.json`.


effigies:
@nellh Actually, I think the `SingleSourceAuthors` check will break a lot of datasets. Enforcing mutual exclusion for something that's not yet valid BIDS is probably a bad idea. Will push a fix.

Issue #4040, opened 2026-07-30
Upload fails with "could not be added, check if this file is accessible and not a broken symlink" for every file


singlesp:
## Summary

`openneuro upload` reports the above message for **every file** in the dataset, a few minutes after validation passes. The files are ordinary readable files — no symlinks anywhere in the tree. Under slightly different timing the same bug instead causes a **silent hang with no output** (see [Scope](#scope-two-different-failure-modes-including-a-silent-hang)).

The message is misleading. The `stat()` call never happens. `context` is `undefined` in the git worker, so `context.fs` throws a `TypeError`, and the `catch (_err)` at [`cli/src/worker/git.ts:184`](https://github.com/OpenNeuroOrg/openneuro/blob/master/cli/src/worker/git.ts#L184) swallows it and prints the symlink message instead:

```
TypeError: Cannot read properties of undefined (reading 'fs')
    at add (cli/src/worker/git.ts:182)
    at PromiseQueue.processQueue (cli/src/worker/queue.ts:17)
    at self.onmessage (cli/src/worker/git.ts:505)
```

## Root cause

`uploadAction` posts `setup` and `clone` to the git worker immediately after `new Worker(...)`. **Deno silently discards messages posted to a `type: "module"` worker while its module graph is still evaluating.** The worker imports `isomorphic-git` and `hash-wasm` (and `annex.ts` has a top-level `await`), so it takes ~5s to evaluate.

`add` is the only message that survives, because it is posted after the directory walk — long after the worker is up. Instrumented run on a 65,558-file dataset, with a fully warm `DENO_DIR`:

```
15:28:33.859  main: creating Worker
15:28:34.107  main: posted setup       <- dropped
15:28:34.108  main: posted clone       <- dropped
15:28:39.309  worker module evaluated     (5.4s later)
15:29:22.406  main: posting add
15:29:22.580  worker RECEIVED add      <- the only message that arrives
```

`setup` never runs, so `context` is never assigned. `clone` never runs either — the log contains no `Cloning …` line, which is a second symptom of the same drop.

## Minimal reproduction

Independent of OpenNeuro. `race.ts`:

```ts
const w = new Worker(new URL("./w.ts", import.meta.url).href, { type: "module" })
w.postMessage({ command: "setup" })
w.postMessage({ command: "clone" })
const got = new Promise<string[]>((r) => { w.onmessage = (e) => r(e.data.received) })
await new Promise((r) => setTimeout(r, 5000))
[...]

PR #4042, opened 2026-07-30
fix(cli): Log the underlying error when a file cannot be stat'd

singlesp:
## Summary

`add()` discards the error from `stat()` and always reports a broken symlink. Any failure that is not a filesystem problem is therefore reported as an inaccessible file, once per file in the dataset, with no way to recover the real cause.

This is what made #4040 hard to diagnose. The worker's `setup` message was being dropped (fixed separately in #4035), so `context` was undefined and `context.fs` threw a `TypeError`. `stat()` never ran. The output was 65,558 copies of:

```
<path> could not be added, check if this file is accessible and not a broken symlink.
```

for a dataset containing no symlinks and no unreadable files. The actual error was:

```
TypeError: Cannot read properties of undefined (reading 'fs')
    at add (cli/src/worker/git.ts:182)
```

## Change

Log the underlying error at `error` level before the existing message.

The per-file `console.log` is left exactly as it is, so there is no change to normal output or to the behaviour anyone currently relies on. Because the default log level is `ERROR`, the underlying error is visible without needing `OPENNEURO_LOG`.

## Testing

- `deno test` — 36 passed, 0 failed
- `deno fmt --check` and `deno lint` clean on the changed file (the 3 pre-existing lint warnings in `git.ts` are unchanged)
- Verified against the environment from #4040: with the swallowed `TypeError` reproduced, the real cause is now printed instead of being silently converted into a symlink message

### DIFF
--- a/cli/src/worker/git.ts
+++ b/cli/src/worker/git.ts
@@ -181,7 +181,14 @@ async function add(event: GitWorkerEventAdd) {
     try {
       const stats = await context.fs.promises.stat(file.path)
       size = stats.size
-    } catch (_err) {
+    } catch (err) {
+      // Log the underlying error - it is not always a filesystem problem and
+      // discarding it makes unrelated failures look like broken symlinks.
+      logger.error(
+        `Failed to stat "${file.path}": ${
+          err instanceof Error ? (err.stack ?? err.message) : String(err)
+        }`,
+      )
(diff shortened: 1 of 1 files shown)


nellh (approved):
Thanks for the contribution!

PR #4048, opened 2026-08-03
fix(cli): Commit the git-annex branch once instead of once per key

### LINKED ISSUE #4047: Upload hangs on datasets with many annexed files: git-annex branch is committed once per key
## Summary

`openneuro upload` appears to hang indefinitely after `Detected N changes.` on datasets with a large number of annexed files. The process sits at ~100% CPU with no log output.

`commitAnnexBranch` generates a commit inside the loop over annex keys, so it creates one commit per annexed file rather than one commit for the upload. Because `CommitBuilder` accumulates its updates and never clears them, each of those commits rebuilds the tree over *every* entry added so far and rewrites a blob for each — work that grows quadratically with the number of annexed files.

On a 65,558-file / 226 GB derivative dataset the upload never completed. For scale, that dataset contains 42,311 annexed files resolving to **40,176 unique annex keys** — the loop runs over `Object.entries(annexKeys)`, so the key count is what drives the commit count.

## Reproduction

Upload any dataset with a few tens of thousands of annexed files. Measured on the dataset above (CLI 5.3.1, but the code is unchanged on `master`):

```
git-annex commits completed:  4,404 of 40,176
rate at that point:           10 commits/min, still degrading
elapsed:                      8h 35m
log output during this phase: none
CPU:                          93%, 9h02m CPU time
```

Extrapolating the quadratic curve from the measured rate puts completion at roughly **two weeks**. The job was cancelled after 1d 07:51 having produced 4.4k of the 40k commits.

The phase is silent, so it is indistinguishable from a deadlock without attaching to the process or inspecting the repo's `git-annex` ref.

## Cause

[`cli/src/worker/git.ts`](https://github.com/OpenNeuroOrg/openneuro/blob/master/cli/src/worker/git.ts) — the commit sits inside `for (const [key, _path] of Object.entries(annexKeys))`:

```ts
    builder.add(annexBranchPath, newAnnexLog)
    changes = true
  }
  // Skip commit if no changes are needed for either remote.log or new annexed objects
  if (changes) {
    ...
    await builder.commit("git-annex", "[OpenNeuro CLI] Added annexed objects")
  }
}   // <- loop actually closes here, after the commit
```

Combined with [`commitBuilder.ts`](https://github.com/OpenNeuroOrg/openneuro/blob/master/cli/src/worker/commitBuilder.ts), where `updates` is never cleared:

```ts
  add(path, content) { this.updates.set(path, data) }
  async commit(...)  { const newTree = await this.buildTree(parentTree, this.updates) }
```

`buildTree` writes a blob for every entry in `updates` on every call, so commit *N* does O(N) blob writes. Total ≈ N²/2 — roughly 800M blob writes for 40,176 keys.

Note this does **not** avoid a large final commit: since `updates` is never cleared, the last commit already contains every entry. The per-key commits produce the same final tree, plus 40,175 progressively larger intermediate rebuilds.

## Regression

Introduced by b31b943 ("refactor(cli): Build git-annex branch without a checkout", 2026-02-18), which added `CommitBuilder`. Before that refactor the loop only staged files and the commit happened once, after it:

```ts
      for (...) {
        ...
        await git.add({ ...context.config(), filepath: annexBranchPath })
      }                                        // loop ends
      if (Object.keys(annexKeys).length === 0) {
        await git.commit({ message: "[OpenNeuro CLI] Configured remote", ... })
      } else {
        await git.commit({ message: "[OpenNeuro CLI] Added annexed objects", ... })
      }
```

`git blame` shows the loop's closing braces still dating to 2024 while the `if (changes)` block dates to the refactor, so this looks like an indentation slip rather than an intentional change. The plural commit message also reads as single-commit intent.

## Fix

Move the commit after the loop, restoring the pre-refactor behaviour. Verified on the same dataset:

```
=== git-annex branch log ===
6866967 [OpenNeuro CLI] Added annexed objects    <- 1 commit from this run
5d90c2c update                                    <- pre-existing history
...
commits added by this run:      1      (was 40,176)
entries in that commit's tree:  40,179    = 40,176 key logs
                                          + activity.log, remote.log, uuid.log
```

The commit phase now completes in seconds, and the upload proceeds to the transfer phase normally — 29,717 keys transferred, 0 failures, at 8h50m elapsed.

## Impact

Scales with annexed file count, so it is invisible on small datasets and fatal on large ones.

## Environment

- OpenNeuro CLI 5.3.1 and `master` (5.4.0) — code identical in this path
- Deno 2.9.0, Linux x86_64, HPC cluster
- 65,558 files / 226 GB; 42,311 annexed files, 40,176 unique annex keys

### Expected behavior

n/a

### How to reproduce

_No response_

### Additional information

_No response_

singlesp:
Fixes #4047

## Problem

`commitAnnexBranch` builds its commit inside the loop over annex keys, so it generates one commit per annexed file rather than one commit for the upload:

```ts
  for (const [key, _path] of Object.entries(annexKeys)) {
    ...
    builder.add(annexBranchPath, newAnnexLog)
    changes = true
    if (changes) {
      await builder.commit("git-annex", "[OpenNeuro CLI] Added annexed objects")
    }
  }
```

`CommitBuilder` accumulates updates and never clears them, so each of those commits rebuilds the tree over every entry added so far and rewrites a blob for each — roughly quadratic in the number of annexed files.

On a 65,558-file / 226 GB dataset (40,176 unique annex keys) the upload never completed: 4,404 of 40,176 commits after 8 hours, rate still degrading, no log output during the phase. Extrapolating the measured curve put completion at about two weeks.

This does not avoid a large final commit either — since `updates` is never cleared, the last commit already contains every entry. The per-key commits produce the same final tree plus 40,175 progressively larger intermediate rebuilds.

## Change

Move the commit after the loop, restoring the behaviour from before the `CommitBuilder` refactor in b31b943, where the loop only staged files and a single `git.commit` followed it.

Whitespace-only for the body of the block; the diff is easier to read with `?w=1`.

## Testing

- `deno test` — 36 passed, 0 failed
- `deno fmt --check` and `deno lint` clean on the changed file (the 3 pre-existing lint warnings in `git.ts` are unchanged)
- Verified end to end on the dataset from #4047:

```
commits added by this run:      1        (was 40,176)
entries in that commit's tree:  40,179   = 40,176 key logs + activity.log, remote.log, uuid.log
```

The commit phase now completes in seconds, and the upload proceeds through the transfer phase normally — all 40,176 objects transferred, 0 failures. The dataset is published at ds008547.

### DIFF
--- a/cli/src/worker/git.ts
+++ b/cli/src/worker/git.ts
@@ -306,17 +306,17 @@ async function commitAnnexBranch(version?: string) {
       builder.add(annexBranchPath, newAnnexLog)
       changes = true
     }
-    // Skip commit if no changes are needed for either remote.log or new annexed objects
-    if (changes) {
-      // Show a better commit message for when only the remote is updated
-      if (Object.keys(annexKeys).length === 0) {
-        await builder.commit("git-annex", "[OpenNeuro CLI] Configured remote")
-      } else {
-        await builder.commit(
-          "git-annex",
-          "[OpenNeuro CLI] Added annexed objects",
-        )
-      }
+  }
+  // Skip commit if no changes are needed for either remote.log or new annexed objects
+  if (changes) {
+    // Show a better commit message for when only the remote is updated
+    if (Object.keys(annexKeys).length === 0) {
+      await builder.commit("git-annex", "[OpenNeuro CLI] Configured remote")
+    } else {
+      await builder.commit(
+        "git-annex",
+        "[OpenNeuro CLI] Added annexed objects",
+      )
     }
   }
 }


nellh (approved):
Thanks for catching this bug.

PR #4057, opened 2026-08-10
Update faq.md

rwblair:
Is there another email address we should use here?


effigies (approved):
Nah, this is probably best.

PR #4077, opened 2026-09-01
chore: Add readthedocs configuration for doc previews in PRs

effigies:
This switches from a `docs/requirements.txt` that provides version constraints to a `uv.lock` file that pins specific versions. We haven't had issues with upstream breaking our doc build, I don't think, but this heads it off.

The main purpose is so that we can get rendered PR builds from readthedocs. The GHA builds verify that the build does not error, but don't let us view the changes.

RTD will not start working until we have a `.readthedocs.yaml` in the `master` branch, so this should be judged based on not breaking the GHA build.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,6 @@
 # GitHub Linguist Override
 .yarn/* linguist-vendored
 .pnp.js linguist-vendored
+
+# Lockfile should not be treated as text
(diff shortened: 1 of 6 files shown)


effigies:
Should have no impact on code or docs. Will revert if something goes wrong.

PR #4088, opened 2026-09-14
doc: Add terms that uploaded data must be real recordings

effigies:
Small update to the terms of service to require imaging data to be actual recordings, not synthetic. I do carve out synthesis for the sake of defacing/refacing, which is arguably synthetic. By specifying "imaging data", I mean to allow synthetic stimuli.


poldrack:
LGTM!

effigies:
### Meeting summary

Discussion points:

* OpenNeuro was designed as a depository for empirical research data.
* Synthetic datasets are cheaper to produce at scale, and *prima facie* have lesser scientific value, than empirical datasets.
* Accepting synthetic datasets without limits has the potential to make finding empirical datasets more difficult.
* From a reporting standpoint, synthetic datasets skew OpenNeuro's participant numbers.
* Synthetic datasets *may* have scientific value for transparency and reproducibility, as do other derivatives, such as preprocessed or cleaned data.

Decisions:

* Technical: Synthetic datasets need to be tagged so OpenNeuro can exclude them from participant counts
* Policy: Synthetic datasets will be handled on a case-by-case basis, and depositors must contact ON support before uploading; Synthetic data deposited without discussion may be deleted

More broadly, we're a bit jealous of the thoroughness of https://docs.dandiarchive.org/terms-policies/policies/, and I'll be reviewing that policy and proposing an adaptation for OpenNeuro that can replace the https://openneuro.org/terms.

effigies:
Marking ready for review. Since this text is in the upload confirmation dialog, I don't want to complicate it further, and adding the "contact support before uploading synthetic data" can be part of the broader updated terms of service.

PR #4091, opened 2026-09-18
feat(worker): Add remote streaming to git_show_content as an option

nellh:
Preserves the fast path for failing quickly on local files but allows callers to enable remote streaming.

Used by bids_mosaic for streaming the remote images during mosaic creation.


jbwexler:
Overal looks good to me. A few questions:
- Is it safe to assume that any private objects (eg a draft or unpublished dataset) will be stored on the openneuro side? If some will be only stored on S3, can stream_from_url() access those?
- Would it be useful to add a few tests for git_show_content() or does that seem unnecessary?
- I haven't manually tested this locally yet but I can if that would be helpful.

jbwexler (approved):

nellh:
>     * Is it safe to assume that any private objects (eg a draft or unpublished dataset) will be stored on the openneuro side? If some will be only stored on S3, can stream_from_url() access those?

They are private if a snapshot exists but the dataset is still embargoed. OpenNeuro's worker can access these here.

>     * Would it be useful to add a few tests for git_show_content() or does that seem unnecessary?

Added some test coverage for both paths.

